PR #12, opened 2018-12-23
Add the 'sleeptime' cvar and behavior from the 'Old Engine'.

Slipyx:
Pretty much just duplicated the code from the old engine. Defaults to 1 millisecond. Also made the description a little more descriptive.


a1batross:
I will merge it for now, but we still need to improve that. At least disable sleeptime if FPS is lower than fps_max.

PR #20, opened 2019-01-28
Made various log messages more informative

x6herbius:
I found that certain log messages I was getting were not entirely clear, so I did a quick skim of the repo and updated some which I thought could do with more information.

### DIFF
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -2473,7 +2473,7 @@ qboolean CL_PrecacheResources( void )
 			{
 				if( FBitSet( pRes->ucFlags, RES_WASMISSING ))
 				{
-					Con_Printf( S_ERROR "Could not load sound %s%s for precache\n", DEFAULT_SOUNDPATH, pRes->szFileName );
+					Con_Printf( S_ERROR "Could not load sound %s%s\n", DEFAULT_SOUNDPATH, pRes->szFileName );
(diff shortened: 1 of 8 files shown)


a1batross:
For me everything is fine.

@mittorn, check this, please.

engine/client/cl_main.c
```diff
@@ -2473,7 +2473,7 @@ qboolean CL_PrecacheResources( void )
 			{
 				if( FBitSet( pRes->ucFlags, RES_WASMISSING ))
 				{
-					Con_Printf( S_ERROR "%s%s couldn't load\n", DEFAULT_SOUNDPATH, pRes->szFileName );
```
mittorn: Regular user does not know what is precache, it need only know that it failed to load

engine/client/s_vox.c
```diff
@@ -545,7 +545,7 @@ void VOX_ParseLineCommands( char *pSentenceData, int sentenceIndex )
 		length = pNext - pSentenceData;
 		if( tempBufferPos + length > sizeof( tempBuffer ))
 		{
-			Con_Printf( S_ERROR "sentence too long!\n" );
+			Con_Printf( S_ERROR "Sentence exceeded max length of %d\n", sizeof(tempBuffer) - 1 );
```
mittorn: "Too long" is better for not english-speaking users. Maybe better "Sentence too long (%d bytes max)\n"

PR #24, opened 2019-01-30
Resource system improvements

mittorn:
Need check for collisions with updates


engine/server/sv_client.c
```diff
@@ -1831,6 +1831,26 @@ static qboolean SV_DownloadFile_f( sv_client_t *cl )
 	{
 		if( sv_send_resources.value )
 		{
+			int i;
+
+			// security: allow download only precached resources
```
a1batross: Wouldn't this break customization?

PR #26, opened 2019-02-23
Updated instructions for Arch Linux

M0Rf30:


a1batross:
xash3d-git installs Old Engine(`xash3d` repo), which is quite incompatible with New Engine(this repo).

M0Rf30:
No, it's the xash3d-fwgs

a1batross:
@M0Rf30 oh, nice then! Merging. But it's early, New Engine isn't even in beta stage and things may change. I really recommend track changes then, for possible new dependencies, build options and so on. Thanks anyway.

a1batross:
@M0Rf30 
1) you don't need to pass `--dedicated` after you did `waf configure`
Waf is following Autotools style here. You only need to pass custom values during the configuration stage, then you can call build/clean/install without any special arguments.
2) `--vgui https://github.com/FWGS/vgui-dev`
This doesn't work that way(yet ;)), you need to clone vgui-dev first, then pass it to `waf`.
3) You don't need to `--enable-stbtt` if you have FreeType library. Just set environment variable `PKG_CONFIG_PATH` pointing to a directory, where is freetype.pc set up for 32-bit library located. On Debian it's `/usr/lib/i386-linux-gnu/pkgconfig`

There is also a to do, to set `xashds` name for dedicated build automatically and just `xash3d` for a normal build.

PR #34, opened 2019-04-01
Prevented net send errors from triggering an entire host error

x6herbius:
Swapped out a `Host_Error` for a `Con_Printf`, to prevent jumping to the console and potentially preventing progress through game menus if the network is down.


a1batross:
@x6herbius for future PRs, please follow our commit message convention. You can read more about in CONTRIBUTION.md file.

PR #46, opened 2019-06-04
SV_LoadProgs: Fixed potential memory leak if function failed

x6herbius:
I noticed this when attempting to load a corrupted server DLL. If the functions cannot be found in the DLL, the existing mempool pointer was overwritten without being freed the next time `SV_LoadProgs` was called.

### DIFF
--- a/engine/server/sv_game.c
+++ b/engine/server/sv_game.c
@@ -5025,14 +5025,14 @@ qboolean SV_LoadProgs( const char *name )
 	// fill it in
 	svgame.pmove = &gpMove;
 	svgame.globals = &gpGlobals;
-	
-	if ( !svgame.mempool )
-	{
-		svgame.mempool = Mem_AllocPool( "Server Edicts Zone" );
-	}
-	
+	svgame.mempool = Mem_AllocPool( "Server Edicts Zone" );
+
 	svgame.hInstance = COM_LoadLibrary( name, true, false );
-	if( !svgame.hInstance ) return false;
+	if( !svgame.hInstance )
+	{
+		Mem_FreePool(&svgame.mempool);
+		return false;
+	}
 
 	// make sure what new dll functions is cleared
 	memset( &svgame.dllFuncs2, 0, sizeof( svgame.dllFuncs2 ));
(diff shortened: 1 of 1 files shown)


a1batross:
Yeah, I've your noticed your PR, I will look into this a bit later.

a1batross:
Well, this memory leak is quite theoretical. 

1) I mean, if server DLL is failed to load, then the singleplayer game is completely broken. The only thing the user can do is to connect to the server, and an unused memory pool will just remain. 
We don't have any fallback server DLLs and even so it's not a good idea to implement it. If the mod has custom DLL, then it has some purpose and falling back to HL's server DLL is completely wrong.

2) SV_LoadProgs not only has been called in loading server, but it's also called when Menu DLL requests saving server config. Ok, Menu DLL is a "game developer side DLL", unlike an engine, and may potentially abuse this call every frame. :)

I'm ok to merge it, but maybe Mem_FreePool should be called when a server DLL reports a fail, so even in the first case, no unused bytes will be left in memory? :)

x6herbius:
That's fair enough, I'll modify it to free the pool on failure.

a1batross:
@x6herbius I don't know where to mail you, so...

Would you mind join our Discord server, if you use it? The badge with link is in `README.md`, near Travis badge. 

Also, we're recently merged ZIP support, which gave us zlib/miniz dependency, so there is plans to make PNG support and maintain less differences between upstream and your fork. :)
If you want, you can push your implementation, but it needs write support, for screenshots.

x6herbius:
Cool! I just joined the Discord but I can't guarantee I'll be on that much, I haven't used Discord in ages.

Zip is a good shout, actually. I've been managing the Nightfire textures in my fork by placing them directly into the textures directory on disk, but keeping everything bundled in a zip sounds like it would make things easier (and as far as I know, the original game's assets.007 file was basically a zip too). Assuming that the engine zip implementation allows looking up assets via an arbitrary path relative to the game directory, the code should just plug straight in and I can remove my texture-specific file system implementation.

The loading of the PNGs themselves was just done via the STB library, so I'd imagine it would be simple enough for me to make a write wrapper around that.

PR #48, opened 2019-06-17
engine: common: imagelib: add simple decoder/encoder for 8-bit RGB/RGBA PNG images

nekonomicon:


engine/common/imagelib/img_png.c
```diff
@@ -0,0 +1,588 @@
...
+
+	// check png signature
+	if( memcmp( buf_p, png_sign, sizeof( png_sign ) ) )
+		return false;
+
+	// move pointer
+	buf_p += sizeof( png_sign );
+
+	// get IHDR chunk length
+	chunk_len = *buf_p++ << 24;
```
a1batross: This may cause crash on ARM due to unaligned access.

engine/common/miniz.h
```diff
@@ -198,15 +198,15 @@ extern "C" {
 typedef unsigned long mz_ulong;
 
 /* mz_free() internally uses the MZ_FREE() macro (which by default calls free() unless you've modified the MZ_MALLOC macro) to release a block allocated from the heap. */
-void mz_free(void *p);
+static void mz_free(void *p);
```
a1batross: Instead of re-defining them as static, you could define `MINIZ_HEADER_FILE_ONLY` before including miniz. 
miniz implementation already defined in filesystem.c, but if you want, you can move it to dummy file.

engine/common/imagelib/img_png.c
a1batross: A note on my prev comment: reading and writing on ARM is possible only if pointer aligned by 32 bits. So moving pointer by one or two bytes will cause a crash.

I think you can read 32-bits and swap it for big endian, if needed. Engine already have these handy macros: https://github.com/FWGS/xash3d-fwgs/blob/master/common/xash3d_types.h#L88

engine/common/imagelib/img_png.c
```diff
@@ -0,0 +1,588 @@
...
+	outsize += idat_len;
+	outsize += sizeof( crc32 );
+	outsize += sizeof( uint );
+	outsize += sizeof( iend_sign );
+	outsize += sizeof( crc32 );
+
+	buffer = (byte *)Mem_Calloc( host.imagepool, outsize );
+
+	// write PNG header
+	buffer[0] = 0x89;
```
a1batross: I think, you can `memcpy` png_sign here, as well as other signatures(IDAT, IHDR) or fill structure and then copy it at once.

engine/common/miniz.h
nekonomicon: I want to refactor this library anyway.

a1batross: I have generated it as a single-header in hope that it will not change, except obvious updates. :)

x6herbius:
Given the STB libraries are already utilised in the MainUI project, would it not make sense to use stb_png for this implementation? I used it for PNG reading in my fork and I'm pretty sure it was less complex than this.

a1batross:
STB library in MainUI C++ is optional, used as a last hope when we can't use nor Win32 APIs nor FreeType, as it only roughly renders TrueType font(doesn't have hinting). And it doesn't changed at all. 

I'm quite unhappy with stb_image and stb_image_write library code, as it bloated(~7.5+1.5k loc) and implements different image formats instead of splitting it to small libraries one per image format. 

It's not the same as PNG loader. Making stb_image smaller(in lines of code) means forking it and harder updates and testing. That's why I'm also unhappy with @nekonomicon's decision to make changes in miniz library.

nekonomicon:
I have a bad experience with stb_image. This library has a bad performance=\

x6herbius:
OK, I guess it might be better not to use STB then. Once this goes in I'll replace the PNG implementation in my fork with this instead.

engine/common/imagelib/img_png.c
nekonomicon: I need swap bytes on little endian, so this macros useless for me.

nekonomicon: Seems I must use ntohl for little endian,
https://www.w3.org/TR/PNG/#7Integers-and-byte-order

a1batross: It's better to use network conversions here, but keep in mind, that we're still not linking to ws32 on Windows yet directly. So it will fail Windows builds.

a1batross:
I'm OK to merge it.

PR #57, opened 2019-07-12
Various: Fixed obscure crashes on Mac

x6herbius:
There were a couple of out-of-range/post-free accesses going on, which caused the Mac build to crash when optimisations were turned on.


a1batross:
Thanks for PR!
Everything is ok. Merging.

PR #66, opened 2019-10-05
wscript: add --enable-lto option

iZarif:


a1batross:
Please, follow commit naming guidelines. Instead of "force lto on release build" it should "wscript: force lto on release build".

Also, maybe instead of silenly forcing it, add a way to disable or enable it? LTO is known to be supported by many compilers but buggy for a long amount of time.

nekonomicon:
>Also, maybe instead of silently forcing it, add a way to disable or enable it? LTO is known to be supported by many compilers but buggy for a long amount of time.

Yep, LTO does not work correctly in some cases, and It can break linkage with some libraries.
I think would be better to add separate option to enable/disable LTO.

a1batross:
Again, please, follow the contributing guidelines:
https://github.com/FWGS/xash3d-fwgs/blob/master/CONTRIBUTING.md#code-style\

Also, any source files must have tabs instead of spaces.

PR #67, opened 2019-10-07
engine: GetGameDir should return gamedir only. Add proper stub functions for some engine APIs

CecilHarvey:
IIRC GetGameDir() used to return full path of game (in very old non-steam HL 1.1.1.0), however this has changed since steam version of Half-Life. Current versions of GoldSrc only returns the gamedir.
This was also the behavior of old versions of Xash3D FWGS.

Also using void functions for stub breaks some MODs which call those functions and rely on the return value (such as Counter-Strike 1.6 regamedll). This fixes the problem.


a1batross:
Thanks for fix!

PR #70, opened 2019-10-19
fix some format warnings

iZarif:


engine/common/identification.c
```diff
@@ -284,7 +284,7 @@ void ID_TestCPUInfo_f( void )
 	bloomfilter_t value = 0;
 
 	if( ID_ProcessCPUInfo( &value ) )
-		Msg( "Got %016llX\n", value );
+		Msg( "Got %016lX\n", value );
```
a1batross: bloomfilter_t is long long typedef.

engine/common/identification.c
```diff
@@ -650,7 +650,7 @@ void ID_Init( void )
 				cfg = fopen( va( "%s/.xash_id", home ), "r" );
 			if( cfg )
 			{
-				if( fscanf( cfg, "%016llX", &id ) > 0 )
+				if( fscanf( cfg, "%016lX", &id ) > 0 )
```
a1batross: Ditto.

engine/common/identification.c
```diff
@@ -665,7 +665,7 @@ void ID_Init( void )
 		const char *buf = (const char*) FS_LoadFile( ".xash_id", NULL, false );
 		if( buf )
 		{
-			sscanf( buf, "%016llX", &id );
+			sscanf( buf, "%016lX", &id );
```
a1batross: Ditto.

engine/common/identification.c
```diff
@@ -700,13 +700,13 @@ void ID_Init( void )
 				cfg = fopen( va( "%s/.xash_id", home ), "w" );
 			if( cfg )
 			{
-				fprintf( cfg, "%016llX", id^SYSTEM_XOR_MASK );
+				fprintf( cfg, "%016lX", id^SYSTEM_XOR_MASK );
```
a1batross: Ditto.

engine/common/identification.c
```diff
@@ -700,13 +700,13 @@ void ID_Init( void )
...
 				fclose( cfg );
 			}
 		}
 	}
 #endif
-	FS_WriteFile( ".xash_id", va("%016llX", id^GAME_XOR_MASK), 16 );
+	FS_WriteFile( ".xash_id", va("%016lX", id^GAME_XOR_MASK), 16 );
```
a1batross: Ditto.

engine/common/identification.c
iZarif: Are you sure?

GCC 9.2 prints a warning:

../engine/common/identification.c:287:19: warning: format ‘%llX’ expects argument of type ‘long long unsigned int’, but argument 2 has type ‘bloomfilter_t’ {aka ‘long unsigned int’} [-Wformat=]

Here is a screenshot:
Снимок экрана_2019-10-22_19-39-16

nekonomicon: https://github.com/FWGS/xash3d-fwgs/blob/master/engine/common/identification.c#L32
https://github.com/FWGS/xash3d-fwgs/blob/master/common/xash3d_types.h#L35

a1batross: @iZarif

Are you compiling 64-bit engine?bMaybe this is a compiler bug then. Because
long long and long int are 64-bit on LP64(amd64 on Linux) and long long is 64 and long int 32 on ILP32(i386) and LLP64(amd64 on Windows).

I think we need to use stdint ASAP...

iZarif: Yes I compile a 64 bit engine

nekonomicon: @a1batross probably, long and long long defined as different types in gcc headers for 64bit.
@iZarif https://www.viva64.com/ru/t/0012/

a1batross: @nekonomicon they can't be _defined_ in _headers_, as they are builtin types. 

That's why I'm thinking it's a compiler bug.

iZarif: @nekonomicon
Do you think this link is directly related to this PR?

nekonomicon: @iZarif Why not?
Your PR breaks valid code.

iZarif: @nekonomicon 
Could you then give a link to the English text?

a1batross:
Closed by request.

PR #71, opened 2019-10-19
fix a few more format warnings

iZarif:


engine/client/ref_common.c
```diff
@@ -443,7 +443,7 @@ static qboolean R_LoadProgs( const char *name )
 	if( !( GetRefAPI = (REFAPI)COM_GetProcAddress( ref.hInstance, GET_REF_API )) )
 	{
 		COM_FreeLibrary( ref.hInstance );
-		Con_Reportf( "R_LoadProgs: can't find GetRefAPI entry point in %s: %s\n", name );
+		Con_Reportf( "R_LoadProgs: can't find GetRefAPI entry point in %s: %s\n", name, COM_GetLibraryError() );
```
a1batross: GetProcAddress never sets an error. 

Last %s was set by an error and should be removed.

PR #85, opened 2019-11-17
Static linking

mittorn:
engine changes:
* library loading now switched by backends
* platform/swap moved to platform/misc as rarely-used
* lib_static for loading linked-in modules
xshlib features:
* linking to relocatable objects instead of shared libraries, generate export table from exports.txt
* add relocatable objects to main binary
* cprogram_static, cxxprogram_static for `gcc -static`
wscript changes:
* allow optional subprojects. If optional subroject not found, it will be ignored
* `--ignore-subproject` option for disabling unuseful subprojects
* `--enable-static-binary` option. Not very useful without static linking now, but custom linker may be also used, i've tested it with makefile before


common/defaults.h
```diff
@@ -131,6 +131,15 @@ SETUP BACKENDS DEFINITIONS
 	#endif
 #endif
 
+#ifdef XASH_STATIC_LIBS
+#define XASH_LIB LIB_STATIC
```
a1batross: Indentation.

engine/wscript
```diff
@@ -90,7 +100,11 @@ def build(bld):
 		source += bld.path.ant_glob(['platform/linux/*.c'])
 
 	if bld.get_define('XASH_CUSTOM_SWAP'):
-		source += bld.path.ant_glob(['platform/swap/*.c'])
+		source += bld.path.ant_glob(['platform/misc/kmalloc.c', 'platform/misc/sbrk.c'])
```
a1batross: Ant_glob is unneeded if there is no wildcards.

wscript
```diff
@@ -37,13 +46,15 @@ class Subproject:
 		return True
 
 SUBDIRS = [
-	Subproject('public',      dedicated=False),
+	Subproject('public',      dedicated=False, mandatory = True),
 	Subproject('game_launch', singlebin=True),
-	Subproject('ref_gl'),
-#rsw	Subproject('ref_soft'),
+	Subproject('ref_gl',),
```
a1batross: `'ref_gl',`?
add #rsw back, until we don't have merged ref_soft into upstream.

mittorn:
I cannot add subdirs with options (because options will not be called) and i annoyed to edit wscript on every ref_soft build and restore it on every commit

engine/wscript
```diff
@@ -23,7 +23,10 @@ def options(opt):
 		help = 'enable custom swap allocator. For devices with no swap support')
 
 	grp.add_option('--enable-legacy-sdl', action = 'store_true', dest = 'SDL12', default = False,
-		help = 'enable using SDL1.2 instead of SDL2(not recommended) [default: %default')
+		help = 'enable using SDL1.2 instead of SDL2(not recommended) [default: %default]')
+
+	grp.add_option('--enable-static-binary', action = 'store_true', dest = 'STATIC', default = False,
+		help = 'build static binary(not recommended, --single-binary required) [default: %default]')
```
a1batross: Why not to force activate single-binary if `--enable-static-binary` is selected?

mittorn:
Because module envs are isolated and i cannot enable SINGLE_BINARY for game_launch module from engine module

a1batross:
Well, it seems that isolated envs has it own pitfalls... 
I made them isolated on purpose.

PR #110, opened 2020-01-11
engine: fix dedicated server crash

Aimless-Wanderer:


engine/client/cl_scrn.c
```diff
@@ -342,12 +342,16 @@ SCR_BeginLoadingPlaque
...
 {
-	float	oldclear;
-
 	S_StopAllSounds( true );
 	cl.audio_prepped = false;			// don't play ambients
 	cl.video_prepped = false;
-	oldclear = gl_clear->value;
+
+#if !XASH_DEDICATED
+	float	oldclear;
```
nekonomicon: Make the code C90-compliant!

engine/client/cl_scrn.c
```diff
@@ -342,12 +342,18 @@ SCR_BeginLoadingPlaque
 */
 void SCR_BeginLoadingPlaque( qboolean is_background )
 {
+#if !XASH_DEDICATED
```
a1batross: XASH_DEDICATED useless in `client` because client sources are never added in dedicated build.

PR #118, opened 2020-01-29
ref_gl: fix remap

Velaron:


a1batross:
Looks fine for me, but didn't built and tested this.

@mittorn could you review this?

mittorn:
@a1batross are you serious? What is fine in this commit?
refState.currentmodel/refState.currententity used only in remap code. Obviously, that should be removed and left only because cleaning remap code is undone. Also setting current entity, but not model is very strange.

mittorn:
https://github.com/FWGS/xash3d-fwgs/commit/2e2a716710485df067ecb23ad7eb7b0bb37f5b14

PR #121, opened 2020-02-10
public: rename mathlib to xash3d_mathlib.

nekonomicon:


a1batross:
Now try to build the menu. :)

nekonomicon:
Need to update menu :)

a1batross:
Didn't noticed other PR. 

While yeah, we have two different mathlibs and this is awful, but menu is more SDK part, which doesn't include libpublic yet.

nekonomicon:
It just breaks compilation of my mdl decompiler, that uses libpublic sources.

a1batross:
Just put public before common includes. 

We need to replace our mathlib(which is derived from Quake's) to common instead. As part of GPL'ed HLSDK headers.

a1batross:
I meant, mv public/mathlib.h common/mathlib.h. 

I don't like this but what can we do? menu is meant to be built with both engine sources and SDK without a line of change.

nekonomicon:
Done.

a1batross:
I'm fine with that. 

@mittorn ?

mittorn:
common/mathlib.h is quake mathlib that used by sdk exports, public/mathlib.h is part of libpublic. It's better to rename public mathlib to xash3d_mathlib

nekonomicon:
>It's better to rename public mathlib to xash3d_mathlib

Done.

PR #122, opened 2020-02-11
engine: common: imagelib: use rgba_t as RGBA palette type.

nekonomicon:


engine/common/imagelib/imagelib.h
```diff
@@ -105,6 +105,12 @@ typedef struct imglib_s
 
 ========================================================================
 */
+
+#define BI_SIZE 40 // size of bitmap info header.
+#if !XASH_WIN32
```
a1batross: Maybe #if !defined(BI_RGB) then? 

Except if it wasn't maliciously defined to something else...

nekonomicon: Done.

Issue #143, opened 2020-03-28
Mousewheel doesn't work twice in a row unless you move the mouse inbetween scrolls


FOTMarut:
### How to reproduce
1. `bind mwheeldown +jump`
2. Load a map
3. Scroll the wheel down once (the player character jumps)
4. Without moving the mouse at all, scroll the wheel down again

### Expected behavior

The player character should jump again.

### Actual behavior

It doesn't, unless you move the mouse before scrolling.

### Platforms

The issue is consistently occurring in the following platforms:
1. Linux 5.5, SDL2.0.12, commit 87340ceca8a9c5beceb35cb9f45ea108a5ee6f31, both 32- and 64-bit builds
2. Windows 10, 0.20 release (32-bit)

### Notes

Will attach a proposed fix soon

FOTMarut:
Done. #144
Tested on SDL2, NOT on SDL 1.2

FOTMarut:
This fix mimicks this: https://sourceforge.net/p/quakespasm/code/HEAD/tree/trunk/quakespasm/Quake/in_sdl.c#l1080

a1batross:
Closed since MR was merged.

PR #144, opened 2020-03-28
engine: platform: sdl: Fix mousewheel in SDL2

FOTMarut:
- See issue 143
- Tested on SDL2
- Untested on SDL1.2


FOTMarut:
See issue #143

a1batross:
I don't remember exactly why, but this were added because of an issues when key down and up event happen on the same frame.

Maybe they don't needed for new engine.

a1batross:
Tested yesterday. So far, looks fine.

PR #196, opened 2020-07-22
engine: common: filesystem.c: do not load zip-archives with zip extension as bullet proof.

nekonomicon:
I think .zip extension can cause trouble with performance penalty if:
1. user put zip-archive with game data to mod folder to unpack it(and not remove it).
2. existing mods already has zip-archives with some data.

And I think It's main reason why Quake 3 uses .pk3 extension for zip-archives.


a1batross:
Indeed this must be confusing to users so LGTM.

@mittorn what to you think?

mittorn:
Ok. Maybe need ability to specify zip packages manually, but not load automaticly

PR #210, opened 2020-08-24
raw input support for windows

Velaron:


engine/client/input.c
```diff
@@ -207,8 +215,14 @@ void IN_ToggleClientMouse( int newstate, int oldstate )
 #if XASH_SDL
 			SDL_SetWindowGrab( host.hWnd, SDL_TRUE );
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
-			if( clgame.dllFuncs.pfnLookEvent )
+#if XASH_WIN32
+			if ( clgame.dllFuncs.pfnLookEvent || ( clgame.client_dll_uses_sdl && CVAR_TO_BOOL( m_rawinput ) ) )
```
a1batross, line 219: I suggest to not have platform-specific macros.

client_dll_uses_sdl will always be true on non-Windows. I gueeeesss, we can set client_dll_uses_sdl to false to not fall into this check when client.dll doesn't implement our input extension. It's not used anywhere else anyway.

Velaron: like dis?

a1batross:
Yeah, cool! LGTM.

PR #213, opened 2020-08-26
Imgbot optimization

Adnan1091:
Lol,it would be better if game files were included.


a1batross:
Thanks but we know about pngcrush.

PR #214, opened 2020-09-03
platform: win32: COM_CheckLibraryDirectDependency fixes

Velaron:


a1batross:
Please, do not reformat this in the same commit where you made some changes.

Velaron:
visual studio formats lines i edit automatically :/

engine/platform/win32/lib_win.c
```diff
@@ -992,15 +991,15 @@ qboolean COM_CheckLibraryDirectDependency( const char *name, const char *depname
...
 		goto libraryerror;
 	}
 
-	importDesc = (PIMAGE_IMPORT_DESCRIPTOR)CALCULATE_ADDRESS( data, directory->VirtualAddress );
+	importDesc = (PIMAGE_IMPORT_DESCRIPTOR)ImageRvaToVa( old_header, data, old_header->OptionalHeader.DataDirectory[IMAGE_DIRECTORY_ENTRY_IMPORT].VirtualAddress, NULL );
 
 	for( ; !IsBadReadPtr( importDesc, sizeof( IMAGE_IMPORT_DESCRIPTOR)) && importDesc->Name; importDesc++ )
 	{
-		const char *importName = ( const char* )CALCULATE_ADDRESS( data, importDesc->Name );
+		const char *importName = ImageRvaToVa( old_header, data, importDesc->Name, NULL );
```
a1batross, line 1002: Isn't dbghelp is supposed to be optional?

By looking at ReactOS code, CALCULATE_ADDRESS probably isn't correct.
https://github.com/reactos/reactos/blob/177ae91bf63810daae17f15bb21be16592c485a1/sdk/lib/rtl/image.c#L341

nekonomicon (changes requested):

engine/platform/win32/lib_win.c
```diff
@@ -957,7 +956,7 @@ qboolean COM_CheckLibraryDirectDependency( const char *name, const char *depname
 	PIMAGE_NT_HEADERS	old_header;
 	PIMAGE_DATA_DIRECTORY	directory;
 	PIMAGE_IMPORT_DESCRIPTOR importDesc;
-	string errorstring;
+	string errorstring = { 0 };
```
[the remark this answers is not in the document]
Velaron: i just saw this style being used in other places

engine/platform/win32/lib_win.c
Velaron: CALCULATE_ADDRESS would cause random access violations for me
it actually worked sometimes but then i could run the engine without any changes and it wouldn't work without any reason
this function works consistently for me

Velaron:
current code works fine and doesnt error out under any circumstances for some reason
at least this fixes garbage printed into the console

a1batross:
So casting really fixes CALCULATE_ADDRESS problem? If so, nice then, LGTM.

I actually wanted to ask you to take that code from reactos. :)

PR #220, opened 2020-09-11
engine: client: Fixed sounds emitted by viewmodel animations only playing at where they were emitted from in the world

AgentAgrimar:
This fixes a minor bug where sounds played via pfnPlaySoundByNameAtLocation (mainly from viewmodel animations e.g. MP5) would be left where it was emitted in the world, rather than following the local player as it does in Goldsrc.

I'm unsure of any long-term issues, but it fixes this one issue in particular.


a1batross:
Thanks for the patch! 

By looking at GS behaviour, it indeed should play by viewentity, so lgtm.

Issue #103, opened 2019-12-11
black screen on opposing force


revelator:
the cutscenes render ok but the start map at the scientist trying to save your life, is rendering all black making it impossible to see anything. 

Tries with vbo on or off it makes no difference.

gfx card is an R9 390 AMD

EternalCringe:
Wrong cahce, maybe?

revelator:
Not sure what cache you mean ?, i noticed though that normal rendering will resume if you wait a few minutes (way past when it should have) so something is definatly up. 

VBO's still not working on AMD also, i get world bleed through
when using flashlight and in opposing force when using the portal gun.

nekonomicon:
New engine has bug with fade.

revelator:
Ouch :S, thats a pretty nasty bug hope it is fixable.

Im wondering if porting idtech4 VBO code would work on xash as it works fine on my card.
The idtech vbo code initialises with stream draw before static draw but im not sure this is what causes the world bleed through. 

I have noticed that the vbo code in both engines are pretty prone to error if something is done incorrectly especially on AMD cards.

nekonomicon:
Create another issue for VBO then.
One issue per bug.

revelator:
will do.

sixcentgeorge:
the bug is present in latest 2.0 release but not previous 0.9 that is at moddb.
i play using hl won 1110 and gearbox . both from clean recent install .
the bug appears to me in different maps of different mods..
it is linked to the the cvar of @SolidClass = worldspawn : "World entity" : 
  startdark(choices) : "Level Fade In" : 0 =
  [ 
    0 : "No"
    1 : "Yes"
  ]

it does not appear in the first map of blu shift : ba_tram1 [ may be because player is in a moving train ] but in others maps or if you add it in a compiled map .

SaikoHaven:
It seems as of now, a workaround is to load opfor with console, and run `v_dark 0` at the proper time in of1a1

sixcentgeorge:
i think that bug is more or less related to this map..
the famous uncle mike... does not "appreciate" those who find bugs or give ideas "at their right valors" ... so he is only have few friends from "sovietscccp" team .. [ the mod only played by " devs , admins, moderators & full members [those who gave money and those who will.. ] beware in his forum : he does like solo-sniper-killer-viking at sovietscccp : he kills in the back without warning... for , mainly , stealing purposes ...lol ] "

a1batross:
[...]

PR #221, opened 2020-09-13
engine: client: update the ref state before R_NewMap is called

AgentAgrimar:
This fixes fades that occur at the beginning of a map being a tad screwy as a result of the ref state not being updated with fresh values, which only happens when a frame is being rendered (loading happens to prevent that).

For this issue:
https://github.com/FWGS/xash3d-fwgs/issues/103


a1batross:
LGTM.

PR #222, opened 2020-09-15
Add support for detecting ARMv8.

zerojay:
Needed to allow compilation on Raspberry Pi 4 under RetroPie.

### DIFF
--- a/public/build.h
+++ b/public/build.h
@@ -52,6 +52,7 @@ For more information, please refer to <http://unlicense.org/>
 #undef XASH_ARMv5
 #undef XASH_ARMv6
 #undef XASH_ARMv7
+#undef XASH_ARMv8
 //#undef XASH_BIG_ENDIAN
 #undef XASH_BSD
 #undef XASH_E2K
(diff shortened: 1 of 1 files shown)


a1batross:
It uses 32-bit OS on ARMv8? What a waste...

ARMv8 IMHO should be defined for both 32-bit and 64-bit then. So ARM64 macro can be removed. 

Also this change should be mirrored in engine/common/build.c and library naming module for Waf and CMake in HLSDK.

mittorn:
Aarch64 has ilp32 mode like x86 x32, aarch32 support may be removed from cpu later to free more chip space

a1batross:
@mittorn yeah but for us keeping yet another architecture means only two files and unchangable architecture name for game libraries.

I suggest calling 32-bit ARMv8 like "armv8_32"  but keep 64-bit ARMv8 like "aarch64". "aarch32" could be misleading since it usually means any 32-bit ARM.

Kactius (approved):

a1batross:
@zerojay 

check it with these commits:
https://github.com/FWGS/xash3d-fwgs/commit/2aaa6c6e0430256c4dd275cb282e1567a068dcae
https://github.com/FWGS/hlsdk-xash3d/commit/dbd9b1c698f00bf132d8ddbf51c06a64970eb55c

Engine should detect itself as "arm8_32" and game libraries must have _arm8_32 postfix. 

I'm closing this PR.

PR #230, opened 2020-10-07
add codeql-analysis.yml

iZarif:
Add file for integration with CodeQL


a1batross:
Yeah, let's try that tool.

PR #236, opened 2020-11-22
engine: common: simplified strings operations.

nekonomicon:


engine/common/con_utils.c
```diff
@@ -836,12 +846,13 @@ qboolean Cmd_CheckMapsList_R( qboolean fRefresh, qboolean onlyingamedir )
 	int	i, size;
 	search_t	*t;
 	file_t	*f;
+	int
```
a1batross, line 849: int what?

nekonomicon: OMG

engine/common/mod_bmodel.c
```diff
@@ -2840,7 +2840,7 @@ qboolean Mod_LoadBmodelLumps( const byte *mod_base, qboolean isworld )
 	{
 		if( !bmod->wadlist.wadusage[i] )
 			continue;
-		Q_strncat( wadvalue, va( "%s.wad; ", bmod->wadlist.wadnames[i] ), sizeof( wadvalue ));
+		Q_snprintf( wadvalue, sizeof( wadvalue ), "%s.wad; ", bmod->wadlist.wadnames[i] );
```
lewa-j, line 2843: Are you sure about that? Looks like instead of appending it wil overwrite.

public/crtlib.h
```diff
@@ -87,6 +88,7 @@ void COM_PathSlashFix( char *path );
 char COM_Hex2Char( uint8_t hex );
 void COM_Hex2String( uint8_t hex, char *str );
 #define COM_CheckString( string ) ( ( !string || !*string ) ? 0 : 1 )
+#define COM_CheckString2( string ) ( ( !*string ) ? 0 : 1 )
```
a1batross, line 91: I don't really like that we have COM_CheckString2 that just don't check for NULL pointer.

Can we just use COM_CheckString for preallocated strings? Are compilers are smart enough to always assume that they are never equal to NULL? I think this macro already used somewhere on preallocated string.

public/crtlib.c
```diff
@@ -603,6 +603,27 @@ int Q_sprintf( char *buffer, const char *format, ... )
 	return result;
 }
 
+char *Q_strpbrk(const char *s, const char *accept)
```
a1batross, line 606: This function can be simplified to:
```
        for( ; *s; s++ )
	{
		const char *k;

		for( k = accept; *k; k++ )
		{
			if( *s == *k )
				return (char*)s;
		}
	}
```

public/crtlib.h
nekonomicon: Unfortunately, compiler gives warning if we check pointer for preallocated string.

public/crtlib.c
nekonomicon: Hmm... Ok.

public/crtlib.h
a1batross: @nekonomicon then maybe give it more useful name? CheckStringEmpty? 

Just for reference, other languages standard libraries call this operation something like "isNullOrEmpty". We could adapt it, thus having COM_IsNullOrEmpty and COM_IsEmpty.

engine/common/infostring.c
```diff
@@ -437,7 +437,7 @@ qboolean Info_SetValueForStarKey( char *s, const char *key, const char *value, i
 
 	Info_RemoveKey( s, key );
 
-	if( !value || !Q_strlen( value ))
+	if( !value || !COM_CheckString( value ))
```
a1batross, line 440: Just COM_CheckString.

public/crtlib.h
nekonomicon: Ok

PR #239, opened 2020-11-29
engine: server: simplified strings operations.

nekonomicon:


engine/server/sv_client.c
```diff
@@ -1289,7 +1289,7 @@ void SV_PutClientInServer( sv_client_t *cl )
 			svgame.dllFuncs.pfnParmsChangeLevel();
 
 			MSG_BeginServerCmd( &msg, svc_restore );
-			Q_snprintf( name, sizeof( name ), "%s%s.HL2", DEFAULT_SAVE_DIRECTORY, sv.name );
+			Q_snprintf( name, sizeof( name ), DEFAULT_SAVE_DIRECTORY "%s.HL2", sv.name );
```
a1batross, line 1292: Is it really worth it? What if this macro gonna be replaced by a variable.

It's not something that executes every frame so why bother?

engine/server/sv_cmds.c
```diff
@@ -379,7 +379,7 @@ void SV_Load_f( void )
 		return;
 	}
 
-	Q_snprintf( path, sizeof( path ), "%s%s.sav", DEFAULT_SAVE_DIRECTORY, Cmd_Argv( 1 ));
+	Q_snprintf( path, sizeof( path ), DEFAULT_SAVE_DIRECTORY "%s.sav", Cmd_Argv( 1 ));
```
a1batross, line 382: Same here.

engine/server/sv_client.c
a1batross: Ah wait. We probably never gonna change it. 

Nevermind.

a1batross (changes requested):

a1batross (approved):

Issue #247, opened 2020-12-27
bug: gles-only platforms uses soft renderer by default


glebm:
Currently, on a platform with only gles (and no gl) the game always use `soft` renderer by default due to a startup logic issue.

Though the code makes [some effort](https://github.com/FWGS/xash3d-fwgs/blob/7b127c23f5d078c0f538aba732caaf978435d8f3/engine/client/ref_common.c#L586) to check which renderer `so` are available at startup, it always tries  `DEFAULT_ACCELERATED_RENDERER` first and falls back to software rendering if that fails:

https://github.com/FWGS/xash3d-fwgs/blob/7b127c23f5d078c0f538aba732caaf978435d8f3/engine/client/ref_common.c#L672-L684

`DEFAULT_ACCELERATED_RENDERER` is hard-coded to `gl`:

https://github.com/FWGS/xash3d-fwgs/blob/7b127c23f5d078c0f538aba732caaf978435d8f3/common/defaults.h#L195-L201

The default accelerated renderer should probably be set to the first available one if `DEFAULT_ACCELERATED_RENDERER` is not available.

mittorn:
Maybe need add selection of DEFAULT_ACCELERATED_RENDERER from build system when GL is disabled?

mittorn:
https://github.com/batocera-linux/batocera.linux/pull/2703
Fallback is broken here

glebm:
While I don't know what the best solution is, it may be best to both set `DEFAULT_ACCELERATED_RENDERER` from waf and also check if the `DEFAULT_ACCELERATED_RENDERER` is available and fallback to the first available one if not

glebm:
Fix verified

PR #253, opened 2020-12-28
engine/wscript: Add --default-accelerated-renderer

glebm:
Refs #247

### DIFF
--- a/engine/wscript
+++ b/engine/wscript
@@ -28,6 +28,9 @@ def options(opt):
 	grp.add_option('--enable-static-binary', action = 'store_true', dest = 'STATIC', default = False,
 		help = 'build static binary(not recommended, --single-binary required) [default: %default]')
 
+	grp.add_option('--default-accelerated-renderer', action='store', dest = 'DEFAULT_ACCELERATED_RENDERER', default = None,
+		help = 'The default accelerated renderer. One of: gl, gles1, gles2, gl4es')
+
 	opt.load('sdl2')
 
 def configure(conf):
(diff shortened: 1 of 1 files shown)


mittorn:
Do we really need this in build system? When GL render is disabled, engine will fallback to GLES.
The reason of your issue is that GL was enabled, failed to initialize, ahd fallback failed because of broken unintitalization.
On special targets you may define it in defaults.h

glebm:
> When GL render is disabled, engine will fallback to GLES.

Are you sure? It does not do this

glebm:
Was replying from phone, a more detailed explanation about why it does not fall back to GLES is here: https://github.com/FWGS/xash3d-fwgs/issues/247#issue-775118991

a1batross:
Sorry it took me whole day to reply. :)

I think we need to rework fallback code, that will check availability of gl, then gles, then gles2, then gl4es renderer and end at software renderer.

We don't really want to give any new build-time options anymore.

mittorn:
Ok, it does not really fallback to gles, only to soft
But separate option is not good idea here.
macro only shpuld be set when user disables gl renderer in build-time, but gles renderer enabled. Or rework fallback code like a1batross proposed

glebm:
I do agree with both of these suggestions. Will someone more familiar with the code be willing to implement them?

glebm:
I'm trying to figure out at least the waf build (never worked with waf before)

glebm:
OK I think I've figured out the `waf` part: https://github.com/FWGS/xash3d-fwgs/pull/257
I'll leave the runtime part to someone more familiar with the way things are done around here

PR #254, opened 2020-12-28
gl4es: Fix `so` name and add to renderer collector

glebm:
1. `libref_gles2_gl4es.so` -> `libref_gl4es.so` because this is what's passed to `R_LoadRenderer`.

2. Add `gl4es` to `DEFAULT_RENDERERS` so that `R_CollectRendererNames` finds it.

### DIFF
--- a/engine/common/com_strings.h
+++ b/engine/common/com_strings.h
@@ -65,7 +65,7 @@ GNU General Public License for more details.
 
 #define XASH_ENGINE_NAME "Xash3D FWGS"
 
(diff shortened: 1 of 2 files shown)


glebm:
This surfaced another issue, it does not appear that `initialize_gl4es()` is called when the library is loaded, causing it to crash on close in `R_CollectRendererNames`

We should either ensure `initialize_gl4es` is called or perhaps set `GL4ES_COMPILE_FOR_USE_IN_SHARED_LIB`

PR #249, opened 2020-12-28
Add gl4es submodule

glebm:
Fixes #248

PR #255, opened 2020-12-28
gl4es: Add GL4ES_COMPILE_FOR_USE_IN_SHARED_LIB

glebm:
This fixes gl4es crashing on unload if it wasn't initialized

It does not get initialized on load because we set `NO_INIT_CONSTRUCTOR`:

https://github.com/FWGS/xash3d-fwgs/blob/5a36a26dd116bdbddbf8b011f59597eb77e6d4cd/ref_gl/wscript#L114

See https://github.com/ptitSeb/gl4es/blob/215b5356ea300ec35c9db2f649964a1bc4d30182/src/gl/init.c#L669-L672

### DIFF
--- a/ref_gl/wscript
+++ b/ref_gl/wscript
@@ -111,7 +111,7 @@ def build(bld):
 			target   = 'gl4es',
 			features = 'c',
 			includes = ['gl4es/src', 'gl4es/src/gl', 'gl4es/src/glx', 'gl4es/include'],
-			defines = ['NOX11', 'NO_GBM', 'NO_INIT_CONSTRUCTOR', 'DEFAULT_ES=2', 'NOEGL', 'EXTERNAL_GETPROCADDRESS=GL4ES_GetProcAddress', 'NO_LOADER', 'STATICLIB'],
+			defines = ['NOX11', 'GL4ES_COMPILE_FOR_USE_IN_SHARED_LIB', 'NO_GBM', 'NO_INIT_CONSTRUCTOR', 'DEFAULT_ES=2', 'NOEGL', 'EXTERNAL_GETPROCADDRESS=GL4ES_GetProcAddress', 'NO_LOADER', 'STATICLIB'],
 			cflags = ['-w', '-fvisibility=hidden', '-std=gnu99'],
 			use      = libs,
 			subsystem = bld.env.MSVC_SUBSYSTEM)


glebm:
This requires a fix in gl4es
Sent https://github.com/ptitSeb/gl4es/pull/260

glebm:
Upstream gl4es PR merged (I've bumped the version in #249)

mittorn:
It is initialized here:
https://github.com/FWGS/xash3d-fwgs/blob/master/ref_gl/gl_opengl.c#L1218

glebm:
That does not get called when collecting available renderers, resulting in a crash

glebm:
Was replying from my phone, here are some more details of the scenario in which it isn't initialized:

`R_CollectRendererNames` dlopens the library:

https://github.com/FWGS/xash3d-fwgs/blob/5a36a26dd116bdbddbf8b011f59597eb77e6d4cd/engine/client/ref_common.c#L600

Initialization does not happen on `dlopen` because we set `NO_INIT_CONSTRUCTOR`:

https://github.com/ptitSeb/gl4es/blob/d14be632a8dc6e15611c154d8ccc9e9da3967650/src/gl/init.c#L63-L68

`R_CollectRendererNames` closes the library soon after:

https://github.com/FWGS/xash3d-fwgs/blob/5a36a26dd116bdbddbf8b011f59597eb77e6d4cd/engine/client/ref_common.c#L633

Closing the library calls `close_gl4es`, which does not check if the library was ever open without `GL4ES_COMPILE_FOR_USE_IN_SHARED_LIB`:

https://github.com/ptitSeb/gl4es/blob/d14be632a8dc6e15611c154d8ccc9e9da3967650/src/gl/init.c#L671-L676

a1batross:
That's because `CollectRendererNames` calls to `GetRefHumanReadableName` export in renderer DLL and then unloads it.

This option is _only_ used to show up renderer description in the game options. 

I'm not sure how to fix this properly.

lewa-j:
Try with latest gl4es now

glebm:
@lewa-j That's a nice fix. Does this fix mean we never close gl4es now?

lewa-j:
You're right. Now the close_gl4es should be called somewhere. I think in the GL_ClearExtensions or the R_Shutdown. @mittorn

glebm:
@lewa-j But we should only call it if we ever called init (which doesn't happen if we didn't create a gl context)

lewa-j:
R_Shutdown shouldn't be called if context creation failed.

glebm:
Filed https://github.com/FWGS/xash3d-fwgs/issues/258

PR #257, opened 2020-12-29
waf: Set DEFAULT_ACCELERATED_RENDERER based on ref_gl options

glebm:
Refs #247


a1batross:
That's not really correct because we build several renderers using various wrappers from same source. 

As it really requires nothing special from enviornment, it's easy to just enable them in common builds.

glebm:
Can you please do what you propose? I don't understand waf that well

PR #262, opened 2020-12-30
makepak.py: Skip all filenames starting with .

glebm:
This makes it much simpler to define a buildroot package for xash-extras
buildroot places various buildroot-internal dot files in the package build directory


a1batross (approved):

a1batross:
For some reaaon I thought it already skips dotfiles.

PR #265, opened 2021-01-02
system.c: Fix `Sys_Warn/Error` text buffer sizes

glebm:
This was using the wrong define
E.g. `-dev 2 -help` output did not fit into it


a1batross:
Oh. Weird, why it used max_syspath at all?

PR #267, opened 2021-01-02
.gitignore: Fix waf ignores

glebm:
Dot (`.`) is not a special character in `.gitignore` files
Reference: https://git-scm.com/docs/gitignore#_pattern_format

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -314,8 +314,8 @@ build-*
 
 # Waf
 build_current
-\.?waf-*/
-\.?waf3-*/
+.waf-*/
+.waf3-*/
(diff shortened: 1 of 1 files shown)


a1batross:
`gitignore` should also take in account non-dotfiles because on Windows, waf unpacks itself into directory without dot.

glebm:
OK but that's a separate issue

a1batross:
I guess it's fine to add them now.

Just `waf-*` and `waf3-*`

a1batross:
Closed in favor of b104855

PR #268, opened 2021-01-02
engine: ref: try to bruteforce all available renderers if we're failed to load requested one

a1batross:
@glebm can you try this patch if #247 got fixed?

### DIFF
--- a/engine/client/ref_common.c
+++ b/engine/client/ref_common.c
@@ -669,17 +669,20 @@ qboolean R_Init( void )
 
 	if( !(success = R_LoadRenderer( refopt )))
 	{
-		// check if we are tried to load default accelearated renderer already
-		// and if not, load it first
-		if( Q_strcmp( refopt, DEFAULT_ACCELERATED_RENDERER ) )
-		{
-			success = R_LoadRenderer( refopt );
-		}
+		int i;
 
-		// software renderer is the last chance...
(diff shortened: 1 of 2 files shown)


glebm:
Thanks!

Looks like this should work (will test later).
It does seem a bit wasteful to try to open a renderer which may not exist.

Perhaps we should merge both this and #257?

a1batross:
> It does seem a bit wasteful to try to open a renderer which may not exist.

We reuse the list that was acquired during CollectRendererNames. It will not try to load something that doesn't exist twice.

glebm:
On line 670 `refopt` would usually be the `DEFAULT_ACCELERATED_RENDERER` (unless `-ref` is passed on the command line)

a1batross:
Ah I see. 

I don't see any problem in this to be honest. So, will dlopen() return NULL, is this a big issue?

glebm:
Not a big deal just makes the log a bit noisy

If we don't want the default to be confiurable, perhaps we should get rid of `DEFAULT_ACCELERATED_RENDERER` entirely?

Then we can just call the first detected renderer (or `ref` if given).

glebm (approved):

a1batross:
Yep, it must be removed. We can use first available then, as we already listed them by the precendence.

PR #269, opened 2021-01-02
SDL GameController: Hotplug and multiple gamepads

glebm:
1. Process events from all game controllers.
2. Support controller hotplug.

### DIFF
--- a/engine/platform/sdl/events.c
+++ b/engine/platform/sdl/events.c
@@ -356,6 +356,48 @@ static void SDLash_ActiveEvent( int gain )
 	}
 }
 
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+static size_t num_open_game_controllers = 0;
+
+static void SDLash_GameController_Add( int index )
+{
+	extern convar_t *joy_enable; // private to input system
+	SDL_GameController *controller;
+	if( !joy_enable->value )
+		return;
+	controller = SDL_GameControllerOpen( index );
+	if( !controller )
+	{
+		Con_Reportf( "Failed to open SDL GameController %d: %s\n", index, SDL_GetError( ) );
+		SDL_ClearError( );
+		return;
+	}
+#if SDL_VERSION_ATLEAST( 2, 0, 6 )
+	Con_Reportf( "Added controller: %s (%i:%i:%i)\n",
+		SDL_GameControllerName( controller ),
+		SDL_GameControllerGetVendor( controller ),
+		SDL_GameControllerGetProduct( controller ),
+		SDL_GameControllerGetProductVersion( controller ));
+#endif // SDL_VERSION_ATLEAST( 2, 0, 6 )
+
+	++num_open_game_controllers;
+	if( num_open_game_controllers == 1 )
+		Joy_AddEvent( );
+}
+
+
+static void SDLash_GameController_Remove( SDL_JoystickID joystick_id )
+{
+	Con_Reportf( "Removed controller %i\n", joystick_id );
+
+	// `Joy_RemoveEvent` sets `joy_found` to `0`.
+	// We only want to do this when all the game controllers have been removed.
+	--num_open_game_controllers;
+	if( num_open_game_controllers == 0 )
+		Joy_RemoveEvent( );
+}
+#endif
+
 /*
 =============
 SDLash_EventFilter
@@ -523,11 +565,11 @@ static void SDLash_EventFilter( SDL_Event *event )
 	}
 
 	case SDL_CONTROLLERDEVICEADDED:
-		Joy_AddEvent( );
+		SDLash_GameController_Add( event->cdevice.which );
 		break;
 
 	case SDL_CONTROLLERDEVICEREMOVED:
-		Joy_RemoveEvent( );
+		SDLash_GameController_Remove( event->cdevice.which );
 		break;
 
 	case SDL_WINDOWEVENT:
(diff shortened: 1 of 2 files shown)


engine/platform/sdl/events.c
```diff
@@ -356,6 +356,38 @@ static void SDLash_ActiveEvent( int gain )
...
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+static size_t num_open_game_controllers = 0;
+
+static void SDLash_GameController_Add( int index )
+{
+	extern convar_t *joy_enable; // private to input system
+	SDL_GameController *controller;
+	if ( !joy_enable->value ) return;
+	controller = SDL_GameControllerOpen( index );
+	if (controller== NULL) {
```
a1batross, line 368:
```suggestion
	if( !controller )
	{
```

engine/platform/sdl/events.c
```diff
@@ -356,6 +356,38 @@ static void SDLash_ActiveEvent( int gain )
...
 }
 
+#if SDL_VERSION_ATLEAST( 2, 0, 0 )
+static size_t num_open_game_controllers = 0;
+
+static void SDLash_GameController_Add( int index )
+{
+	extern convar_t *joy_enable; // private to input system
+	SDL_GameController *controller;
+	if ( !joy_enable->value ) return;
```
a1batross, line 366:
```suggestion
	if( !joy_enable->value )
		return;
```

engine/platform/sdl/events.c
```diff
@@ -356,6 +356,38 @@ static void SDLash_ActiveEvent( int gain )
...
+		return;
+	}
+#if SDL_VERSION_ATLEAST( 2, 0, 6 )
+	Con_Reportf( "Added controller: %s (%i:%i:%i)\n",
+		SDL_GameControllerName( controller ),
+		SDL_GameControllerGetVendor( controller ),
+		SDL_GameControllerGetProduct( controller ),
+		SDL_GameControllerGetProductVersion( controller ));
+#endif // SDL_VERSION_ATLEAST( 2, 0, 6 )
+	if ( ++num_open_game_controllers == 1 ) Joy_AddEvent( );
```
a1batross, line 380:
```suggestion
	if( ++num_open_game_controllers == 1 )
		Joy_AddEvent( );
```

Wait, why 1 though? What if we connect more than 1 controller? I know, we don't use it at this moment and there is no mod that uses multiple gamepads(Decay could be one of them but there is no source code of Decay :)), but even if it's unused, we still need to notify engine about them.

Also, I don't really like the `++x == y` construction. Maybe put increment somewhere else, so check will be much more obvious?

engine/platform/sdl/events.c
```diff
@@ -356,6 +356,38 @@ static void SDLash_ActiveEvent( int gain )
...
+}
+
+
+static void SDLash_GameController_Remove( SDL_JoystickID joystick_id )
+{
+	Con_Reportf( "Removed controller %i\n", joystick_id );
+	if ( --num_open_game_controllers == 0 ) Joy_RemoveEvent( );
```
a1batross, line 387:
```suggestion
	if( --num_open_game_controllers == 0 )
		Joy_RemoveEvent( );
```

Same question here as on similar comment above.

engine/platform/sdl/in_sdl.c
```diff
@@ -202,10 +198,7 @@ SDLash_JoyInit_New
 */
 static int SDLash_JoyInit_New( int numjoy )
 {
-	int temp, num;
-	int i;
-
-	Con_Reportf( "Joystick: SDL GameController API\n" );
```
a1batross, line 208: That was pretty useful line because we support both GameController API and legacy Joystick API.

engine/platform/sdl/in_sdl.c
```diff
@@ -214,54 +207,14 @@ static int SDLash_JoyInit_New( int numjoy )
...
-
-	for( i = 0; i < temp; i++ )
-	{
-		if( SDL_IsGameController( i ))
-			num++;
-	}
+	count = 0;
+	numJoysticks = SDL_NumJoysticks();
+	for ( i = 0; i < numJoysticks; i++ )
+		if ( SDL_IsGameController( i ) ) ++count;
```
a1batross, line 215:
```suggestion
		if( SDL_IsGameController( i ))
			++count;
```

a1batross (changes requested):

engine/platform/sdl/events.c
glebm: We only want to call `Joy_Add/RemoveEvent` once because these methods set the `joy_found` variable.

If we called it every time, the following sequence would break `joy_found`

* Add gamepad 1 -- joy_found is 1
* Add gamepad 2 -- joy_found is 1
* Unplug gamepad 2 -- joy_found is 0 -- wrong

Instead, we only call `Add` when activating the first joystick and `Remove` when removing the last joystick.

TLDR: The `Joy_Add/RemoveEvent` code assumes only a single joysticks

engine/platform/sdl/in_sdl.c
glebm: Restored

engine/platform/sdl/in_sdl.c
glebm: Done

engine/platform/sdl/events.c
glebm: I've also added a comment in `SDLash_GameController_Remove` explaining the check.
The check in `Add` is not strictly necessary with the current `Joy_AddEvent` implementation.

a1batross: >The Joy_Add/RemoveEvent contract seems to assume only a single joystick

I wanted to get all events from the operating system, maybe to do something with them in the future. But you're right, we don't do anything with this and I doubt we will.

Marking as resolved then.

a1batross (approved):

PR #271, opened 2021-01-02
Add .clang-format

glebm:
Based on https://github.com/FWGS/xash3d/blob/725fd84b5ba8d405979d63df7a906feb1bc87d51/.clang-format with the following changes:

1. Uncomments `BasedOnStyle: WebKit` to be explicit.
2. Adds:

       AlignConsecutiveDeclarations: true
       AlignConsecutiveMacros: true

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,35 @@
+BasedOnStyle: WebKit
+IndentWidth: 4
+TabWidth: 4
+UseTab: ForIndentation
+
+BreakBeforeBraces: Allman
+
+AlignAfterOpenBracket: true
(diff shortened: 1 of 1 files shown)


a1batross:
.clang-format is fine but I wonder if it still correctly formats Xash code. 
Mostly it was formatted by hand and I kinda like it more than automated formatting.

glebm:
@a1batross Well we can improve upon the config over time

It already formats closer to xash3d style than any default IDE settings

a1batross:
Yep, agree.

Forgot to merge. :)

PR #272, opened 2021-01-02
Remove DEFAULT_(ACCELERATED|SOFTWARE)_RENDERER

glebm:
This is no longer needed as we now use the `DEFAULT_RENDERERS` order priority and only try to load the detected renderers.


engine/client/ref_common.c
```diff
@@ -659,23 +659,27 @@ qboolean R_Init( void )
...
+	if( !success )
 	{
 		int i;
 
 		// cycle through renderers that we collected in CollectRendererNames
 		for( i = 0; i < ref.numRenderers; i++ )
 		{
 			// skip renderer that was requested but failed to load
-			if( Q_strcmp( refopt, ref.shortNames[i] ))
+			if( !Q_strcmp( requested, ref.shortNames[i] ) )
```
glebm, line 682: @a1batross This is also a bugfix: Q_strcmp returns 0 when strings *are* equal

a1batross: Damn, right.

engine/client/ref_common.c
```diff
@@ -600,7 +600,7 @@ void R_CollectRendererNames( void )
 		dll = COM_LoadLibrary( temp, false, true );
 		if( !dll )
 		{
-			Con_Reportf( "R_CollectRendererNames: can't load library %s: %s\n", temp, COM_GetLibraryError() );
+			Con_Reportf( "R_CollectRendererNames: can't load library %s: %s", temp, COM_GetLibraryError() );
```
a1batross, line 603: It's not a good idea to remove newline here.

Keep it, if COM_GetLibraryError puts newline, then it should be fixed.

a1batross (changes requested):

engine/client/ref_common.c
glebm: Done

PR #274, opened 2021-01-03
Trim all trailing whitespace

glebm:
The `.editorconfig` file in this repo is configured to trim all trailing whitespace regardless of whether the line is modified.

Trims all trailing whitespace in the repository to make the codebase easier to work with in editors that respect `.editorconfig`.

`git blame` becomes less useful on these lines but it already isn't very useful.

Commands:

```bash
find . -type f -name '*.h' -exec sed --in-place 's/[[:space:]]\+$//' {} \+
find . -type f -name '*.c' -exec sed --in-place 's/[[:space:]]\+$//' {} \+
```


glebm:
Proof-link that these are all whitespace-only changes:
https://github.com/FWGS/xash3d-fwgs/pull/274/files?diff=unified&w=1

PR #289, opened 2021-01-09
ref_gl: gl_rsurf.c: don't use asserts to prevent engine from stopping when vbo is disabled.

nekonomicon:


a1batross:
@mittorn pls review this

mittorn:
This asserts may be removed/commented out. This only causes incorrect rendering when triggered

PR #317, opened 2021-02-18
partially fix amd64 build on windows

w23:
Quality of this PR is questionable. I have no idea whether what I did here makes sense or not.
But it compiles and runs (provided with hlsdk-xash3d amd64 build, which is another story).

This hacks `wscript` `DEST_CPU` value really badly (tldr: `msvc` will detect 64-bit cpu as `amd64`, but many parts of xash build scripts, including some in waifu, expect it to be `x86_64` and break badly if it's not set properly). I've yet to figure out how to fix it properly.
Hence DRAFT status of this PR.

This change is a precursor for RTX Vulkan effort -- VK_KHR_ray_tracing_pipeline and friends are only available on 64-bit nvidia drivers (no idea about AMD, pls send GPUs onegai).

I'd appreciate reviews and any other help in stabilizing Windows amd64 target.

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -3945,7 +3945,7 @@ qboolean CL_LoadProgs( const char *name )
 
 	// a1ba: we need to check if client.dll has direct dependency on SDL2
 	// and if so, disable relative mouse mode
-#if XASH_WIN32 && !defined(XASH_64BIT)
+#if XASH_WIN32 && !XASH_64BIT
 	if( ( clgame.client_dll_uses_sdl = COM_CheckLibraryDirectDependency( name, OS_LIB_PREFIX "SDL2." OS_LIB_EXT, false ) ) )
 	{
 		Con_Printf( S_NOTE "%s uses SDL2 for mouse input\n", name );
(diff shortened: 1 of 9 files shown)


magisk6666:
fuck




---Original---

magisk6666:
fuck


---Original---

engine/client/cl_game.c
```diff
@@ -3945,7 +3945,7 @@ qboolean CL_LoadProgs( const char *name )
 
 	// a1ba: we need to check if client.dll has direct dependency on SDL2
 	// and if so, disable relative mouse mode
-#if XASH_WIN32
+#if XASH_WIN32 && !defined(XASH_64BIT)
```
a1batross, line 3948: If I'm not wrong, we eliminated empty definitions in favor of 1 or 0.

So you can rewrite this to #if XASH_WIN32 && !XASH_64BIT

Or replace by XASH_WIN64. See public/build.h

w23:
I'm looking into finalizing this and so far I have trouble understanding how does `-8` waf argument even work.
It seems that it sets `conf.options.ALLOW64`, which then is used to set `conf.env.BIT32_MANDATORY` to negative of its value. But after that nothing happens! Neither `ALLOW64` nor `BIT32_MANDATORY` are referenced nowhere else in the repo no matter how hard I grep.

a1batross:
@w23 can you check if it's working on latest build for you?

w23:
@a1batross smoke-tested it, seems to work. :+1:

a1batross:
Merged to master.

PR #363, opened 2021-03-15
readme: improved installation instructions

SNMetamorph:


README.md
```diff
@@ -26,6 +26,18 @@ Read more about Xash3D on ModDB: https://www.moddb.com/engines/xash3d-engine
 * Voice support
 * Vulkan renderer
 
+## Installation & Running
+0) Download Xash3D binaries: you can use [release](https://github.com/FWGS/xash3d-fwgs/releases) build (recommended) or [testing](https://github.com/FWGS/xash3d-deploy/tree/anewengine-master) build, also you can compile engine from sources.
```
a1batross, line 30: Those links are available at the top of README. 

I don't like repeating information but maybe it can be useful for people who don't know how to read.

I also think, we should not recommended using stable release at this point, it's too old.

README.md
```diff
@@ -26,6 +26,18 @@ Read more about Xash3D on ModDB: https://www.moddb.com/engines/xash3d-engine
...
+Choose proper build package depending on which platform you're using.
+1) Copy engine binaries to some directory.
+2) Copy `valve` directory from [Half-Life](https://store.steampowered.com/app/70/HalfLife/) to mentioned above directory with engine binaries.
+Also if you're using Windows: you should copy `SDL2.dll` and `vgui.dll` libraries from Half-Life directory to Xash3D directory.
```
a1batross, line 34: We distribute our own SDL2.dll.

a1batross (changes requested):

README.md
SNMetamorph: >I also think, we should not recommended using stable release at this point, it's too old.
Also thinking about it, but decided to keep both

README.md
nekonomicon: It may be not a good solution if Valve games provides SDL2 and linux distributives too.

a1batross: @nekonomicon naah.

SDL2 keeps binary compatibility. At least, it should.

The only thing, we can't go backwards and there isn't too much sense doing that.

a1batross (approved):

PR #458, opened 2021-05-29
readme: update discord invite

Velaron:


a1batross:
We should use discord.fwgs.ru redirect for that...

Velaron:
it literally didn't work like a minute ago, so i thought it's dead

PR #461, opened 2021-05-30
engine: backport some fixes from 0.19 

SNMetamorph:
Too early to merge now.

### DIFF
--- a/engine/client/client.h
+++ b/engine/client/client.h
@@ -844,6 +844,12 @@ movevars_t *pfnGetMoveVars( void );
 
 _inline cl_entity_t *CL_EDICT_NUM( int n )
 {
+	if( !clgame.entities )
+	{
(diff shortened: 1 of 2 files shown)


mittorn:
Did you test demo recording/playback?

engine/client/client.h
```diff
@@ -844,6 +844,9 @@ movevars_t *pfnGetMoveVars( void );
 
 _inline cl_entity_t *CL_EDICT_NUM( int n )
```
mittorn, line 845: Maybe it should not be called when clgame.entities is NULL?

SNMetamorph: Yeah, that's better way

SNMetamorph: Maybe write assert here instead return NULL? Or not assert, but something for to break debugger here

SNMetamorph:
@mittorn yes I tested demo playing/recoring, it's properly working

PR #463, opened 2021-06-01
engine: server: fix missing baselines

Velaron:
Fixes https://github.com/FWGS/xash3d-fwgs/issues/422


a1batross:
Check itself seems do nothing useful.

PR #470, opened 2021-06-02
Draft: Rework mouse

a1batross:
- [x] Use IN_MouseEvent(int) for client DLL
- [x] Probably fix https://github.com/FWGS/xash3d-fwgs/issues/449
- [x] Fix another VGUI issue when mouse clicks are also triggering +attack (or whatever bound to MOUSE1)
- [ ] Touch emulation
- [ ] SDL1.2, evdev, Android fixes

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -3967,7 +3967,7 @@ qboolean CL_LoadProgs( const char *name )
 #else
 	// this doesn't mean other platforms uses SDL2 in any case
 	// it just helps input code to stay platform-independent
-	clgame.client_dll_uses_sdl = false;
+	clgame.client_dll_uses_sdl = true;
 #endif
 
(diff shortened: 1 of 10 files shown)


Velaron:
Tested on Windows with TFC, #449 is fixed, but I noticed two problems:
The mouse cursor moves even while the VGUI window is not active, the mouse cursor should appear in the same place as it was when the VGUI menu was closed.
After closing a VGUI menu, the player's camera moves a few degrees.

Vladislav4KZ:
I confirm, but I also noticed this bug in the console when you move the mouse

a1batross:
Finally fixed that mouse movement after VGUI menu closes. It works perfectly on my (Linux) side.

I'll merge it as soon as CI passes, other bug reports must be opened as separate issues.

PR #478, opened 2021-06-04
Haiku OS support

exstrim401:


engine/wscript
```diff
@@ -62,6 +62,10 @@ def configure(conf):
 			conf.fatal('SDL2 not availiable! If you want to build dedicated server, specify --dedicated')
 		conf.define('XASH_SDL', 2)
 
+	if conf.env.DEST_OS == 'haiku':
+		conf.env.LIB_HAIKU = ['network']
```
a1batross, line 66: Haiku has nothing similar to pkg-config?

a1batross: Also is this library guaranteed to always exist at this path?

a1batross: Otherwise lgtm.

Does port really work?

exstrim401: network is system library needed to sockets work. it is guaranteed to always exists at this path.
port is completely working

exstrim401: haiku has pkg-config but i'm not famillar with waf

a1batross: Nah, pkg-config is good (there is check_pkg method for it) but it's mostly used for something that's not guaranteed to exist at certain location.

For Haiku it means we probably can't crosscompile engine but I doubt there is real necessity in it.

a1batross: Gonna merge it now.

For future patches: please use our commit message style, it's explained in CONTRIBUTING.md.

PR #483, opened 2021-06-05
engine: Haiku BeGameLauncher support

exstrim401:
https://github.com/EXL/BeGameLauncher

### DIFF
--- a/game_launch/game.cpp
+++ b/game_launch/game.cpp
@@ -20,14 +20,8 @@ GNU General Public License for more details.
 #include <stdlib.h>
 #include <stdarg.h>
 
-#if defined(__APPLE__) || defined(__unix__)
+#if defined(__APPLE__) || defined(__unix__) || defined(__HAIKU__)
 	#define XASHLIB    "libxash." OS_LIB_EXT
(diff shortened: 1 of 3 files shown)


a1batross:
What MIRRORDIR does?

To be honest, I don't remember specifics of Haiku port by EXL but to me this doesn't look right.
If I'm not wrong, MIRRORDIR is another layer for filesystem? Today we have RoDir for that.

exstrim401:
I also made XASH3D_GAME env available for all platforms to choose game at start

game_launch/game.cpp
```diff
@@ -138,7 +139,12 @@ _inline int Sys_Start( void )
 		changeGame = Sys_ChangeGame;
 #endif
 
-	ret = Xash_Main( szArgc, szArgv, GAME_PATH, 0, changeGame );
+	char *game = getenv( E_GAME  );
```
a1batross, line 142: make it `const char` please, so we won't lose const modifier two lines below. Otherwise lgtm

a1batross (changes requested):

PR #487, opened 2021-06-06
engine: client: hack clientdata being overwritten by delta parse, so predicted data won't get lost

a1batross:


Velaron:
Extremely broken, gauss keeps releasing secondary fire by itself, servers with infinite ammo are broken, it seems like predicted data doesn't get overriden at all.

a1batross:
YAAAAAAAY

This is not needed anymore, with ea5937bc.

PR #493, opened 2021-06-12
Reject Connection on Undefined Protocol

bariscodefxy:


a1batross:
What's the problem? If version is 0, then it will not pass version != PROTOCOL_VERSION anyway.

bariscodefxy:
> What's the problem? If version is 0, then it will not pass version != PROTOCOL_VERSION anyway.

Safety is paramount.

lewa-j:
Current PROTOCOL_VERSION is 49. So it will reject any version expept 49. It includes 0.

a1batross:
> > What's the problem? If version is 0, then it will not pass version != PROTOCOL_VERSION anyway.
> 
> Safety is paramount.

Can you show an example where 0 != 49 passes the check?

bariscodefxy:
> > > What's the problem? If version is 0, then it will not pass version != PROTOCOL_VERSION anyway.
> > 
> > 
> > Safety is paramount.
> 
> Can you show an example where 0 != 49 passes the check?

No :D

PR #497, opened 2021-06-15
server: mark TSource responses as connectionless 

ridvan1995:
Also changing the OS type based on which platform engine was built on.


a1batross:
Yep, seems correct. Thanks!

PR #522, opened 2021-07-09
ref_gl: gl_triapi: fixed culling enum in TriCullFace

SNMetamorph:


a1batross:
I think that's fixed #516 #515 and #512

Vladislav4KZ:
> I think that's fixed #516 #515 and #512

Ok, I'll recheck it later.

PR #550, opened 2021-07-30
engine: client: added API for GoldSrc Monitor

SNMetamorph:


a1batross:
Well, I'm glad you're extending engine on your own needs, we're all doing that.

But how about using existing APIs? Like, you can extend our MobilityAPI or RenderAPI, what fits better in your opinion. For example, SPR_Load already present in RenderAPI. SPR_Frames present in normal client interface.

I'm not exactly sure how GSMon works, in my opinion it should wrap client.dll. Doing so carefully will never cause problems with compatibility.

SNMetamorph:
The problem is GoldSrc Monitor does not proxying client.dll and therefore can't use any on this APIs. 
Instead of proxying client.dll, tool injects in game and search for client engine functions. This solution used to avoid modifying game files and it's just more convenient for user.

a1batross:
Then both SPR_Frames and SPR_LoadExt must be set as exports instead. Anyway, they are already part of another APIs and shouldn't drastically change.

SNMetamorph:
Ok I did it

PR #615, opened 2021-09-15
RISC-V Support

archanox:


a1batross:
Thank you for your contribution!

Patch seems to be ok but I would add some things, just for the future.

0) RISC-V defines three float ABIs: soft, single and double. Unfortunately, Half-Life SDK interfaces are defined that way that we need to care about all three. 

1) It's kinda possible that 32-bit RISC-V exists? I see you added a placeholder so I think it's ok for now.
But there is also a chance that somebody will use 64-bit CPU with ILP32 which is questionable but we've already ran into it with AArch64 https://github.com/FWGS/xash3d-fwgs/commit/2aaa6c6e0430256c4dd275cb282e1567a068dcae

Remember, the library naming isn't a support measure but rather "let's define this OS/arch/ABI with unique name".

2) And of course, please follow our contribution guidelines, especially for the commit messages. No worry, I can just squash-merge it to rename it, since patch looks very trivial.

archanox:
> Thank you for your contribution!
> 
> Patch seems to be ok but I would add some things, just for the future.
> 
> 0) RISC-V defines three float ABIs: soft, single and double. Unfortunately, Half-Life SDK interfaces are defined that way that we need to care about all three. 

No problems, I can make these changes
> 
> 1) It's kinda possible that 32-bit RISC-V exists? I see you added a placeholder so I think it's ok for now.
> But there is also a chance that somebody will use 64-bit CPU with ILP32 which is questionable but we've already ran into it with AArch64 https://github.com/FWGS/xash3d-fwgs/commit/2aaa6c6e0430256c4dd275cb282e1567a068dcae

No this shouldn't be possible. 
More info here https://forums.sifive.com/t/compatibility-issue-a-risc-v-32-bit-app-can-work-well-on-risc-v-64-bit-core/1980
tl;dr 32bit compiled binaries won't work on 64bit cores

32bit RISC-V Linux isn't really a thing with major distros, but it is possible through buildroot and the like. So best catch it and say "no"!
> 
> Remember, the library naming isn't a support measure but rather "let's define this OS/arch/ABI with unique name".

Would these changes make more sense when there is actual software and hardware support for these configurations?
> 
> 2) And of course, please follow our contribution guidelines, especially for the commit messages. No worry, I can just squash-merge it to rename it, since patch looks very trivial.

Sorry I can clean this up, I made the assumption you'd squash the commits on pull requests.

a1batross:
Yeah, you're right. :) 

I guess there is no sense to think about rv32 yet if it's not even virtually supported anywhere.

PR #622, opened 2021-09-22
engine: filesystem: fixed current directory changing for Windows (issue #571)

SNMetamorph:


engine/common/filesystem.c
```diff
@@ -2394,6 +2394,24 @@ qboolean FS_SysFileExists( const char *path, qboolean caseinsensitive )
...
+*/
+int FS_SetCurrentDirectory( const char *path )
+{
+#if XASH_WIN32
+	static wchar_t pathBuffer[MAX_PATH];
+	MultiByteToWideChar(CP_UTF8, 0, path, -1, pathBuffer, MAX_PATH);
+	return SetCurrentDirectoryW(pathBuffer);
+#elif XASH_POSIX
+	return !chdir(path);
+#endif
```
a1batross, line 2412:
```
#else                 
#error               
#endif          
```

a1batross (changes requested):

a1batross:
Does it work? I may add #else branch myself and merge it.

@SNMetamorph

nekonomicon:
Nope. It's not solution.
I think will be better to use getcwd instead of SDL functions on Windows.

a1batross:
We can just apply code from original Xash that's confirmed to be working. :)

a1batross:
Also getcwd is wrong. We need to get path where exe is located, not working directory.

SNMetamorph:
> 
> 
> Does it work? I may add #else branch myself and merge it.
> 
> @SNMetamorph

Yes, it works, but there is another issue somewhere in WAD loading code I guess. Current directory successfully changes, but engine doesn't start with this error:
изображение
But it happens only if path to game directory has cyrillic characters, I guess it happens because some functions which works with string can't handle UTF-8 properly.

engine/common/filesystem.c
```diff
@@ -463,6 +463,22 @@ static const char *FS_FixFileCase( const char *path )
...
+/*
+====================
+FS_PathToWideChar
+
+Converts input UTF-8 string to wide char string.
+====================
+*/
+const wchar_t *FS_PathToWideChar(const char *path)
+{
+	static wchar_t pathBuffer[MAX_PATH];
```
a1batross, line 476: Isn't MAX_PATH is cringe 260 bytes while Windows API can potentially use more? Or it can't?
MS docs says it still can't and for long path support engine must be patched and it must be enabled in registry. Oh well...

Also please follow our code style ^^

SNMetamorph: Only Windows 10 supports long path. Ok I fixed code style.

mittorn:
I think it's better to change SetCurrentDirectory to narrow API, because msvc runyime still use native codepages instead of utf-8

nekonomicon:
> I think it's better to change SetCurrentDirectory to narrow API, because msvc runyime still use native codepages instead of utf-8

Yep, conversion looks useless.

PR #658, opened 2021-10-25
engine: sv_save: added entity_state_t->startpos to save fields

SNMetamorph:
Should fix XashXT bug described here https://github.com/SNMetamorph/PrimeXT/issues/10
But I'm not sure, will it break compatibility with old saves?


SNMetamorph:
ДМ это комментировал вот как:
https://hlfx.ru/forum/showthread.php?s=&threadid=5279&postid=196528#post196528

PR #666, opened 2021-11-03
Tighten engine and client security: Implement cl_filterstuffcmd

a1batross:
[ ] Select CVars and commands to protect
[ ] Decide what to do with touch cvars/commands, probably should be filtered by default. Or allowed but with unprivileged mode flag
[ ] Testing, testing, testing!


a1batross:
Approving myself, I guess, as nobody reviewed this.

PR #674, opened 2021-11-15
engine: Add 0.19.2 enttools

xWhitey:


a1batross:
I'm not against enttools in new engine, I think it was one of the reason what made old engine at least fun, but the reason we didn't added this is that @mittorn who wrote enttools had plans on remaking this feature.

SNMetamorph:
> 
> 
> I'm not against enttools in new engine, I think it was one of the reason what made old engine at least fun, but we didn't added this is that @mittorn who wrote enttools had plans on remaking this feature.

So I think even in current state it will be useful, anyway.

engine/server/server.h
```diff
@@ -445,6 +445,10 @@ extern	convar_t		*sv_hostmap;
 extern	convar_t		*sv_validate_changelevel;
 extern	convar_t		*public_server;
 
+// 0.19.2 stuff
```
a1batross, line 448: You don't need to comment "it's from 0.19.2".

Just modernize code a bit and it's ok for new engine.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
 	return true;
 }
 
+edict_t* pfnFindEntityInSphere(edict_t* pStartEdict, const float* org, float flRadius);
```
a1batross, line 2094: Maybe move function definition in server.h, under all sv_game.c definitions?

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+
+static edict_t* SV_GetCrossEnt(edict_t* player)
```
a1batross, line 2096: Reformat this patch following our code style: https://github.com/FWGS/xash3d-fwgs/blob/master/CONTRIBUTING.md#code-style

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+	SV_ClientPrintf(cl, "flags: 0x%x\n", ent->v.flags);
+	SV_ClientPrintf(cl, "spawnflags: 0x%x\n", ent->v.spawnflags);
+}
+
+void SV_EntSendVars(sv_client_t* cl, edict_t* ent)
+{
+	if (!ent)
+		return;
+	MSG_WriteByte(&cl->netchan.message, svc_stufftext);
+	MSG_WriteString(&cl->netchan.message, va("set ent_last_name \"%s\"\n", STRING(ent->v.targetname)));
```
a1batross, line 2371: Something tells me we don't have "set" command anymore. It's have to be restored.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+	MSG_WriteString(&cl->netchan.message, "ent_getvars_cb\n");
+}
+
+void SV_EntGetVars_f(sv_client_t* cl)
+{
+	edict_t* ent = NULL;
+
+	if (Cmd_Argc() != 2)
+	{
+		SV_ClientPrintf(cl, "Use ent_getvars <index|name|inst>\n");
```
a1batross, line 2390: `S_USAGE "ent_getvars <index|name|inst>\n"`

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+*/
+void SV_EntFire_f(sv_client_t* cl)
+{
+	edict_t* ent = NULL;
+	int	i = 1, count = 0;
+	qboolean single; // true if user specified something that match single entity
+
+	if (Cmd_Argc() < 3)
+	{
+		SV_ClientPrintf(cl, "Use ent_fire <index||pattern> <command> [<values>]\n"
```
a1batross, line 2414: Same as above.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+		}
+		else if (!Q_stricmp(Cmd_Argv(2), "angles"))
+		{
+			ent->v.angles[0] = Q_atof(Cmd_Argv(3));
+			ent->v.angles[1] = Q_atof(Cmd_Argv(4));
+			ent->v.angles[2] = Q_atof(Cmd_Argv(5));
+		}
+		else if (!Q_stricmp(Cmd_Argv(2), "setflag"))
+		{
+			ent->v.flags |= 1U << Q_atoi(Cmd_Argv(3));
```
a1batross, line 2624: Use SetBits and BIT macros.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+			SV_ClientPrintf(cl, "flags set to 0x%x\n", ent->v.flags);
+		}
+		else if (!Q_stricmp(Cmd_Argv(2), "clearflag"))
+		{
+			ent->v.flags &= ~(1U << Q_atoi(Cmd_Argv(3)));
```
a1batross, line 2629: Use ClearBits and BIT macros.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+			SV_ClientPrintf(cl, "flags set to 0x%x\n", ent->v.flags);
+		}
+		else if (!Q_stricmp(Cmd_Argv(2), "setspawnflag"))
+		{
+			ent->v.spawnflags |= 1U << Q_atoi(Cmd_Argv(3));
```
a1batross, line 2634: Use SetBits and BIT macros.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+			SV_ClientPrintf(cl, "spawnflags set to 0x%x\n", ent->v.spawnflags);
+		}
+		else if (!Q_stricmp(Cmd_Argv(2), "clearspawnflag"))
+		{
+			ent->v.spawnflags &= ~(1U << Q_atoi(Cmd_Argv(3)));
```
a1batross, line 2639: Use ClearBits and BIT macros.

engine/server/sv_client.c
```diff
@@ -2091,6 +2091,744 @@ static qboolean SV_SendBuildInfo_f( sv_client_t *cl )
...
+*/
+void SV_EntCreate_f(sv_client_t* cl)
+{
+	edict_t* ent = NULL;
+	int	i = 0;
+	string_t classname;
+
+	if (Cmd_Argc() < 2)
+	{
+		SV_ClientPrintf(cl, "Use ent_create <classname> <key1> <value1> <key2> <value2> ...\n");
```
a1batross, line 2702: `S_USAGE "ent_create <classname> <key1> <value1> <key2> <value2> ...\n"`

a1batross (changes requested):

engine/server/sv_client.c
xWhitey: It seems that this method is useless so I just deleted it.

a1batross: It isn't?

The idea is to send new entity parameters like name, origin, class, etc so it can be used in callback alias.

Like this:

```
alias ent_last_cb "echo New entity origin is $ent_last_origin"
ent_create monster_scientist
```

It will print entity origin. AFAIK it's already used in some simple mods.

mittorn:
I think, admin features should be removed from engine as it does not have any admin utils to control it.
User enttools with multiplayer features are already implemented in GGM:
https://github.com/FWGS/hlsdk-xash3d/blob/gravgun/dlls/enttools.cpp
BTW, look at updated function:
https://github.com/FWGS/hlsdk-xash3d/blob/gravgun/dlls/enttools.cpp#L153
engine implementation is very buggy now.
engine enttools are useful for modders and engine developers to use with unmodified server code. It still not merged because it is very depended on scripting (cmd_scripting) which is removed from new engine and i doubt it really needed in such limited way
Modern engines integrating complex languages like lua for scripting, but it will break compatibility with configs

a1batross:
cmd_scripting is available in new engine

SNMetamorph:
Can this be merged in soon time?

PR #743, opened 2022-01-28
engine: make pfnDrawString & pfnDrawStringReverse return width

kungfulon:
Looking at https://github.com/ValveSoftware/halflife/blob/master/cl_dll/hud_redraw.cpp:

```cpp
int CHud :: DrawHudString(int xpos, int ypos, int iMaxX, char *szIt, int r, int g, int b )
{
	return xpos + gEngfuncs.pfnDrawString( xpos, ypos, szIt, r, g, b);
}
```

```cpp
int CHud :: DrawHudStringReverse( int xpos, int ypos, int iMinX, char *szString, int r, int g, int b )
{
	return xpos - gEngfuncs.pfnDrawStringReverse( xpos, ypos, szString, r, g, b);
}
```

We can see that `pfnDrawString` and `pfnDrawStringReverse` are supposed to return width instead of new X position.

I checked the behavior by using current Half-Life `client.dll`. I sent `ShowMenu` message to draw this titles entry:

```
Team_Menu_Join_IG
{
\wa\db\yc\rd\wa\db\yc\rd\wa\db\yc\rd\wa\db\yc\rd\wa\db\yc\rd\wa\db\yc\rd\wa\db\yc\rd
}
```

Result using current Xash3D master:

bad_menu

Result using this patch:

good_menu

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -2929,15 +2929,16 @@ pfnDrawString
 */
 static int GAME_EXPORT pfnDrawString( int x, int y, const char *str, int r, int g, int b )
 {
+	int iWidth = 0;
 	Con_UtfProcessChar(0);
 
 	// draw the string until we hit the null character or a newline character
 	for ( ; *str != 0 && *str != '\n'; str++ )
 	{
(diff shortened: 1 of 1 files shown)


a1batross:
Thanks! Merging~

PR #750, opened 2022-01-31
engine: implement support for secured client dlls

kungfulon:
Mainly for loading DoD (latest Steam version). Besides VGUI2 and filesystem_stdio, DoD calls some `dst` functions which are given by the engine through `F` function. I used a new define to not break compatibility with current `GetClientAPI` export.

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -3932,6 +3932,8 @@ void CL_UnloadProgs( void )
 	Cmd_Unlink( CMD_CLIENTDLL );
 }
 
+void GetSecuredClientAPI( CL_EXPORT_FUNCS F );
+
 qboolean CL_LoadProgs( const char *name )
 {
 	static playermove_t		gpMove;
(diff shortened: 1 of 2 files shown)


a1batross:
I don't think this should be hidden under macro. Only F export actually tells us if we need APIProxy stubs, so it must be safe for compatibility.

kungfulon:
Yeah for the stubs themselves, macro guard is unneeded. But my concern is the `cldll_func_s` struct where there're some new fwgs extensions (need to insert 2 more functions before these). Also, do you want to check for both `F` and `GetClientAPI`? Or we should just go with `F`?

a1batross:
`F` is a Valve's proprietary export, seems to be some kind of leftover of encrypted DLLs or VAC. I'm still not sure is it even supposed to give client interface.

`GetClientAPI` is Xash specific and only used in XashXT and Paranoia 2 for now. Though we can use it and extend in the future.

It's probably better to have two versions of `cl_dll_funcs_t` structure. First for compatibility with `F` and second for `GetClientAPI` and to use internally like with `clgame.dllFuncs`.

kungfulon:
Yeah I agree. I modified the patch, which now looks for `F` if it cannot find `GetClientAPI`, and map the exports accordingly.

a1batross (approved):

a1batross:
Yep, looks fine to me.

This security module is a completely insane thing. I think it's fine if we leave this on this level.

PR #776, opened 2022-02-22
engine: workaround -Wstringop-overflow

hgkjshegfskef:
```
GCC 11.1 produces -Wstringop-overflow warnings on Linux, e.g.:

./engine/client/cl_parse.c:1173:17: warning: ‘memset’ writing 340 bytes
into a region of size 0 overflows the destination [-Wstringop-overflow=]
 1173 |        memset( &ent->prevstate, 0, sizeof( ent->prevstate ));
      |        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Explicitly checking whether the pointer that is to be written into is
not null works it around.
```

### DIFF
--- a/engine/client/cl_parse.c
+++ b/engine/client/cl_parse.c
@@ -1170,7 +1170,8 @@ void CL_ParseBaseline( sizebuf_t *msg )
 			Host_Error( "CL_AllocEdict: no free edicts\n" );
 
 		ent = CL_EDICT_NUM( newnum );
-		memset( &ent->prevstate, 0, sizeof( ent->prevstate ));
+		if ( ent ) // workaround -Wstringop-overflow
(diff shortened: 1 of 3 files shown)


lewa-j:
Probably better to put under the condition all the code that uses `ent`, not just memset.

a1batross:
Thanks for the patch!

However, `CL_EDICT_NUM` shouldn't return NULL. Albeit there is a line `return NULL`, it goes into `Host_Error` which will either shut down the engine or abort the current frame. SV_InitEdict also has an assertion for NULL.

Thus, NULL checks don't make any sense. Compiler is probably giving you a false-positive, GCC diagnostics are known to be buggy sometimes. Better disable them in your build (`-Wno-stringop-overflow`) or upgrade the compiler.

PR #795, opened 2022-03-12
engine: common: imagelib: fixed loading BMP files with v4/v5 header formats

SNMetamorph:


engine/common/imagelib/img_bmp.c
```diff
@@ -53,9 +53,9 @@ qboolean Image_LoadBMP( const char *name, const byte *buffer, fs_offset_t filesi
 		return false;
 	}
 
-	if( bhdr.bitmapHeaderSize != 0x28 )
+	if(!( bhdr.bitmapHeaderSize == 40 || bhdr.bitmapHeaderSize == 108 || bhdr.bitmapHeaderSize == 124 ))
 	{
-		Con_DPrintf( S_ERROR "Image_LoadBMP: invalid header size %i\n", bhdr.bitmapHeaderSize );
+		Con_DPrintf( S_WARN "Image_LoadBMP: %s have non-standard header size %i\n", name, bhdr.bitmapHeaderSize );
```
a1batross, line 58: Let's keep it as error. Imagelib refuses to load such files anyway

a1batross (changes requested):

engine/common/imagelib/img_bmp.c
SNMetamorph: Hm, maybe instead just try to load file? It looks like a more versatile solution.

a1batross: No, it's not a good idea to pass potentially broken images and will lead to random bugs.

a1batross (changes requested):

engine/common/imagelib/img_bmp.c
SNMetamorph: Ok

PR #805, opened 2022-03-25
Vulkan

benadhit:


a1batross:
If you want to work on Vulkan renderer, there is https://github.com/w23/xash3d-fwgs.

It does RTX in the first place, but has unfinished rasterizing mode.

PR #815, opened 2022-04-06
Major refactoring of Win32 console code

SNMetamorph:


engine/server/sv_main.c
```diff
@@ -153,6 +153,69 @@ qboolean SV_HasActivePlayers( void )
...
+}
+
+/*
+================
+SV_UpdateStatusLine
+
+updates dedicated server status line in console
+================
+*/
+void SV_UpdateStatusLine( void )
```
a1batross, line 198: If this function doesn't do anything on other platforms, it makes sense to move it to `platform/win32` and rename to Platform_UpdateStatusLine. And leave it empty for other, so no cycles wasted on Linux or Android.

engine/platform/win32/con_win.c
```diff
@@ -25,197 +25,428 @@ WIN32 CONSOLE
...
 {
-	if( host.type != HOST_DEDICATED ) return;
-	SetWindowText( s_wcd.hwndInputLine, inputText );
-	SendMessage( s_wcd.hwndInputLine, EM_SETSEL, Q_strlen( inputText ), -1 );
+	if( host.type != HOST_DEDICATED )
+		return;
+
+	while( s_wcd.consoleTextLen-- )
+	{
+		Wcon_PrintInternal( "\b \b", 0 );
```
a1batross, line 157: Oh, that annoying bell when you press backspace on empty input... xD

Awesome :D

a1batross (changes requested):

engine/server/sv_main.c
a1batross:
```
#if XASH_WIN32
void Platform_UpdateStatusLine( void );
#else 
static inline void Platform_UpdateStatusLine( void ) { }
#endif      
```

Just drop it in platform.h, and it's ok

SNMetamorph: I think it just should be implemented for other platforms too in future. Something like it works in ReHLDS. I would implement it also for Linux but I don't know how to work with it :)

a1batross: It's an out of scope of this PR. Anyway, it's gonna be very platform-specific. On Android I would like to put that in notification bar, on Linux... we'll see, I'm dreaming about making an ncurses based console. :)

SNMetamorph: Then ok I'll do it platform-specific

SNMetamorph: But there are some issues:
1. Required serverside data unavailable from platform code
2. Can I mix platform-specific and serverside code together? I think it's bad practice.

SNMetamorph: Maybe just don't call this function for other platforms instead?

a1batross: We had this in Old Engine and it was a mess of macros in random places over the whole engine.

There is still work to be done, of course.

a1batross: >Can I mix platform-specific and serverside code together? I think it's bad practice.

Why not, server side code is always available.
It can be that client code isn't available and there is XASH_DEDICATED macro for that.

SNMetamorph: Но даже если это вынести в платформо-специфичную часть, мне все равно придётся входные данные считать снаружи из серверного кода, опять таки это будут лишние циклы, если не юзать #ifdef

SNMetamorph: Вроде понял, как сделать нормально

a1batross (approved):

PR #827, opened 2022-04-10
engine: keydefs: fixed mouse buttons code to match GoldSrc

SNMetamorph:


a1batross:
How they were that different?

SNMetamorph:
MOUSE2 and MOUSE3 were mixed up

a1batross:
Well, I'm wondering how this even happened. Anyway, I've checked and indeed they were mixed up. Thanks for the fix!

PR #830, opened 2022-04-12
Fixed TriAPI WorldToScreen behavior to match GoldSrc

SNMetamorph:
Tested
@a1batross

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -3178,16 +3178,7 @@ convert world coordinates (x,y,z) into screen (x, y)
 */
 int TriWorldToScreen( const float *world, float *screen )
 {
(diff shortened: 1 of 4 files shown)


a1batross:
Seems fine to me. 

But since we don't need TriWorldToScreen wrapper, can you get rid of it? See CL_FillTriAPIFromRef function in ref_common.c file.

SNMetamorph:
Ok why not

SNMetamorph:
I have looked and not sure should I do that or not. Because TriWorldToScreen is widely used across all code, therefore I should remove all this references and it causes many file changes. Is it worth it?

a1batross:
It can be separate patch then, thanks!

PR #841, opened 2022-05-04
engine: common: made check for Blue Shift maps format more reliable

### LINKED ISSUE #835: BSPGuy and "Bad Normal for Plane" Error Messages
The story with this error is that whilst I am editing some of the Life's End maps for example, in order to crossover content from Half-Life: Absolute Zero, with XASH 3D FWGS occasionally causes plane errors on the map I've edited, which caused the game to automatically disconnect, and when I try to type in the same map with this issue, it crashes the whole game.

Untitled
[New WinRAR ZIP archive.zip](https://github.com/FWGS/xash3d-fwgs/files/8519619/New.WinRAR.ZIP.archive.zip)

SNMetamorph:
Should fix #835, but also fixes issues with other maps

### DIFF
--- a/engine/common/mod_bmodel.c
+++ b/engine/common/mod_bmodel.c
@@ -2785,8 +2785,7 @@ qboolean Mod_LoadBmodelLumps( const byte *mod_base, qboolean isworld )
 	if( header->version == HLBSP_VERSION )
 	{
 		// only relevant for half-life maps
-		if( header->lumps[LUMP_ENTITIES].fileofs <= 1024 &&
-			(header->lumps[LUMP_ENTITIES].filelen % sizeof( dplane_t )) == 0 )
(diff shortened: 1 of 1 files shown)


a1batross:
No, no, we shouldn't do this. Even if it fixes this case in #835, such check also breaks the usage of bshift maps outside bshift and normal Half-Life maps in bshift. Thus also opening a Pandora Box of horrible mods like Blue Shift Unlocked.

nekonomicon:
Also, some people can try to play multiplayer with barney.

a1batross:
We probably can validate the lumps before continuing loading the level. And if data mismatch, then it's probably swapped. And even if this doesn't help, then it's not Blue Shift case and probably broken BSP.

SNMetamorph:
Superceded by #842

PR #842, opened 2022-05-05
engine: common: make blue-shift map detect more robust

a1batross:
Better version of #841


a1batross:
cc @SNMetamorph @nekonomicon

a1batross:
TestBmodelLumps only used for auto-completion code, where only 1kb of map is loaded, so we only validating it's length, without validating lump data. Which in current version of loader (without any file length checks!!! >_<) can crash the engine.

PR #848, opened 2022-05-08
Fix console attaching to existing

zgdump:
## Описание

Исправляет создание новой консоли, если Xash запускается из командной строки.

Сначала происходит вызов `AttachConsole` и только если он завершается ошибкой, вызывается `AllocConsole`

## Скриншот

Было:

image

Стало:

image

## Баги

- [x] После запуска движок убивает консоль, даже если мы запускались из командной строки
- [x] Восстановление заголовка после закрытия движка
- [x] Отображение prompt и нормализация положения курсора

### DIFF
--- a/engine/platform/win32/con_win.c
+++ b/engine/platform/win32/con_win.c
@@ -581,6 +581,8 @@ destroy win32 console
 */
 void Wcon_DestroyConsole( void )
 {
+	int return_key = VK_RETURN;
+
 	// last text message into console or log
 	Con_Reportf( "Sys_FreeLibrary: Unloading xash.dll\n" );
 
@@ -595,6 +597,7 @@ void Wcon_DestroyConsole( void )
 	if( s_wcd.is_attached )
 	{
 		SetConsoleTitle( &s_wcd.old_title );
+		SendMessage( s_wcd.hWnd, WM_CHAR, &return_key, 0 );
 	}
 
 	FreeConsole();
--- a/engine/platform/win32/con_win.c
+++ b/engine/platform/win32/con_win.c
@@ -494,6 +494,8 @@ create win32 console
 */
 void Wcon_CreateConsole( void )
 {
+	qboolean is_attached;
+
 	if( Sys_CheckParm( "-log" ))
 		s_wcd.log_active = true;
 
@@ -509,7 +511,11 @@ void Wcon_CreateConsole( void )
 		s_wcd.log_active = true; // always make log
 	}
 
-	AllocConsole();
+	is_attached = ( AttachConsole( ATTACH_PARENT_PROCESS ) != 0 );
+	
+	if( !is_attached )
+		AllocConsole();
+
 	SetConsoleTitle( s_wcd.title );
 	SetConsoleCP( CP_UTF8 );
 	SetConsoleOutputCP( CP_UTF8 );
@@ -525,7 +531,8 @@ void Wcon_CreateConsole( void )
 		return;
 	}
 
-	SetWindowPos( s_wcd.hWnd, HWND_TOP, 0, 0, 0, 0, SWP_NOSIZE | SWP_NOREPOSITION | SWP_SHOWWINDOW );
+	if( !is_attached )
+		SetWindowPos( s_wcd.hWnd, HWND_TOP, 0, 0, 0, 0, SWP_NOSIZE | SWP_NOREPOSITION | SWP_SHOWWINDOW );
 
 	// show console if needed
 	if( host.con_showalways )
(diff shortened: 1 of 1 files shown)


engine/platform/win32/con_win.c
```diff
@@ -29,6 +29,9 @@ WIN32 CONSOLE
 
 typedef struct
 {
+	qboolean    is_attached;
+
+	char        old_title[512];
```
a1batross, line 34: You can straight away use types like string, which is usually long enough to store generic strings.

engine/platform/win32/con_win.c
```diff
@@ -568,17 +581,25 @@ destroy win32 console
...
-	if( s_wcd.hWnd )
+	if( !s_wcd.is_attached && s_wcd.hWnd )
 	{
 		ShowWindow( s_wcd.hWnd, SW_HIDE );
 		s_wcd.hWnd = 0;
 	}
 
+	if( s_wcd.is_attached )
+	{
+		SetConsoleTitle( &s_wcd.old_title );
```
a1batross, line 599: You can minimize the scope of return_key variable by defining it's inside a block here, just at the beginning.

Also, I highly doubt that the third parameter of SendMessage is actually just int*. Is it even supposed to be a pointer? MSDN says wParam is just a character code for WM_CHAR.

a1batross (changes requested):

a1batross:
Also, please, follow commit message conventions.

engine/platform/win32/con_win.c
nekonomicon: WPARAM defined as UINT_PTR or uintptr_t.

a1batross: I'm not a Windows programmer you know but everywhere I see, WM_CHAR uses just a key value rather than a pointer. Even in the Wine source!

nekonomicon: Even in the ReactOS %)
`_PTR` suffix means different size on different architectures, not a pointer.

a1batross: I'm talking about the code above, which sends a pointer to a variable located in the stack.

Something really tells me it's just wrong.

zgdump: I thought for a long time and it also seemed wrong to me. But if you remove the pointer, then not a return is sent, but some ^M. The last parameter should be NULL, not 0, but this does not affect anything

zgdump:
I wrote a PR in Russian, but you respond in english. How do you communicate here?)

Let's squash the commits when accepting the PR and name them correctly. Small fixes, I think it's ok

engine/platform/win32/con_win.c
a1batross: ^M is carriage return. As I read in Discord chat, it's only to solve the "missing" command prompt?

How does the pointer get decoded then?

nekonomicon: Try to send just '\n'.
VK_RETURN has the same code as CR in ascii, but code of LF is reserved.
Pointer cause undefined behavior anyway.

zgdump: @nekonomicon 

We have two different behaviors in cmd and powershell.

In cmd, it is enough to send VK_RETURN, and the prompt will be displayed immediately after the log. Nice and neat

Some kind of mess in the powershell. The prompt is drawn over the text when any key is pressed. If I send '\n' as you suggested, then the caret advances to the next line, displaying '>>'. Sending VK_RETURN will display '^M'...

engine/platform/win32/con_win.c
zgdump: Fix in [a70fa30](https://github.com/FWGS/xash3d-fwgs/pull/848/commits/a70fa304ee1e0a1bae1de4a2fb26346578246f56)

engine/platform/win32/con_win.c
a1batross: As a last resort, you can print "Press Enter to continue". xD

SNMetamorph:
Ну как оно? Готовое или нет?

engine/platform/win32/con_win.c
a1batross: @SNMetamorph well, maybe you can help solve the mystery of should wParam be a pointer or a key code, in this discussion.

Or get rid of this nonsense code, which again looks like a workaround of buggy Windows interactive shells and send the other parts.

That's why I always tell PRs should be minimal and solve exactly one problem, and leave debatable stuff to the next patches. Or at least next commit lol.

SNMetamorph: @a1batross I think it's better to get rid of SendMessage and just try to print `\n` with conventional methods.

a1batross:
send a new pr then pls

сб, 12 нояб. 2022 г., 17:56 Vladimir ***@***.***>:

SNMetamorph:
Superceded by #1313

PR #876, opened 2022-06-10
engine: strip color codes when writing to log

Velaron:


engine/common/sys_con.c
```diff
@@ -230,6 +230,9 @@ void Sys_PrintLog( const char *pMsg )
 	time( &crt_time );
 	crt_tm = localtime( &crt_time );
 
+	// strip color codes
+	COM_StripColors( pMsg, (char *)pMsg );
```
a1batross, line 234: You're modifying the string that you're not supposed to modify. Just a few lines below this, you have Sys_PrintColorized that just converts color codes to escape sequences that terminals interpret as color change.

engine/common/sys_con.c
```diff
@@ -248,21 +251,21 @@ void Sys_PrintLog( const char *pMsg )
 #ifdef XASH_COLORIZE_CONSOLE
 	Sys_PrintColorized( logtime, pMsg );
 #else
-	printf( "%s %s", logtime, pMsg );
+	printf( "%s%s", logtime, pMsg );
```
a1batross, line 254: Here you can replace printf by function similar to Sys_PrintColorized that will print but ignoring color codes. Except try to avoid copying, it's useless, and I don't know why PrintColorized was made that way.

engine/common/sys_con.c
a1batross, line 234: Here you can replace printf by function similar to Sys_PrintColorized that will print but ignoring color codes. Except try to avoid copying, it's useless, and I don't know why PrintColorized was made that way.

a1batross (changes requested):

a1batross (approved):

a1batross:
Thanks! <3

PR #877, opened 2022-06-10
engine: common: filesystem.c: do not load archives with zip extention… again.

nekonomicon:
Do not load fucking zip anymore!

1. Some mods already has zip and rar archives with additional content.
2. Users can place many archives with game data to mod folder.

It will be better to load compressed zip archives with `.pk3` extention only as bullet proof.


a1batross (approved):

a1batross:
Never liked that pk3-convention to be honest but what can we do

PR #878, opened 2022-06-12
engine: try to detect LoadLibrary/FreeLibrary leaks

a1batross:
Should be moved out of lib_posix.c to generic lib_common utils, to enable it for most platforms.

IMHO, it is a purely an engine development debugging tool, so must be disabled by default.


a1batross:
It's just an example of what we can do by wrapping platform-specific LoadLibrary/FreeLibrary. I don't think we need this in upstream. This PR will be closed, but if anyone wants to debug our pairs of LoadLibrary/FreeLibrary, this can be rebased and reapplied.

PR #886, opened 2022-06-21
utils: mdldec: try to create user defined destination directory.

nekonomicon:


utils/mdldec/mdldec.c
```diff
@@ -164,9 +167,9 @@ static qboolean LoadMDL( const char *modelname )
 
 	if( destdir[0] != '\0' )
 	{
-		if( !IsFileExists( destdir ) )
+		if( -1 == _mkdir( destdir ) && EEXIST != errno )
```
a1batross, line 170: EEXIST doesn't say it's a directory. Should it be checked with stat also?

a1batross (approved):

PR #911, opened 2022-06-29
wscript: added parallel builds compiler flag for MSVC

SNMetamorph:


a1batross:
Thanks, even on Actions it's faster :)

PR #914, opened 2022-07-01
filesystem: introduce new module, based on engine filesystem. 

a1batross:
The goal is to share filesystem code between engine and utilities and provide C++ VFileSystem interface in the future.

~~Does not link now, requires all exports to be set.~~

Next stage is to have dynamic C API (or not?)


nekonomicon:
It seems you forget to remove `miniz.h` from `engine/common`:
https://github.com/FWGS/xash3d-fwgs/blob/fs/engine/common/miniz.h

SNMetamorph:
> As suggested by @mittorn, should it be actually named "filesystem_stdio"?

I think it's better to name it "filesystem" to avoid confusing. It looks like a more flexible solution - we don't stick to Valve's interface between FS and engine, therefore it can be expanded if it'll be needed.

a1batross:
@SNMetamorph but then what we do for compatibility? My goal is to make zero changes in existing mods.

Make a simple wrapper between extended interface and VFileSystem009?

a1batross:
@SNMetamorph for extended interface, we can make a new interface, since CreateInterface just returns a pointer by interface name.

a1batross:
The interface between engine and filesystem module is simple.

It's just a bunch of exported functions that you can link with (see filesystem.h). It's never meant to be stable or used by anything else than engine and utils.

The interface for games is planned to be only VFileSystem009.

nekonomicon:
Any yet another wrapper can be replaced by user too.
Ideally will be better to have compatibility with filesystem_stdio.
And I think we can just link statically filesystem module to engine and to utils plus provide wrapper for mods. Or there something wrong?

a1batross:
Yeah. Split filesystem into a separate library

And link it statically

пн, 4 июл. 2022 г., 21:28 nekonomicon ***@***.***>:

filesystem/VFileSystem009.cpp
```diff
@@ -0,0 +1,473 @@
...
+	}
+
+	void Flush( FileHandle_t handle ) override
+	{
+		FS_Flush( (file_t *)handle );
+	}
+
+	bool EndOfFile( FileHandle_t handle ) override
+	{
+		return !FS_Eof( (file_t *)handle );
```
lewa-j, line 222: Why invert?

filesystem/VFileSystem009.cpp
```diff
@@ -0,0 +1,473 @@
...
+}
+
+class CXashFS : public IVFileSystem009
+{
+private:
+	class CSearchState
+	{
+	public:
+		CSearchState( CSearchState **head, search_t *search ) :
+			next( *head ), search( search ), index( 0 ), handle( (*head)->handle + 1 )
```
lewa-j, line 64: `(*head)->handle` may segfault when `*head` in null (first search)

a1batross:
With VFileSystem009 implementation, halflife-updated finally starts to work.

a1batross:
Let's merge it and see what it breaks later.

As anybody who I asked to review this code, haven't responded or found anything suspicious.

PR #951, opened 2022-08-07
Documentation: not-supported-mod-list-and-reasons-why.md: update.

nekonomicon:


a1batross:
SC 5.0+ wouldn't work anyway, it uses modified GoldSrc engine that I have no intention to support.

nekonomicon:
Ah... ok.

SNMetamorph:
> SC 5.0+ wouldn't work anyway, it uses modified GoldSrc engine that I have no intention to support.

I agree, it's better to focus on SC 4.8 instead

a1batross (approved):

PR #965, opened 2022-08-15
Voice chat support

SNMetamorph:
I think it's time to merge this, so players can test and report about problems/their suggestions


SNMetamorph:
@a1batross @Velaron @mittorn

engine/client/cl_parse.c
```diff
@@ -1696,7 +1699,30 @@ CL_ParseVoiceData
...
-	// TODO: ???
+	int size, idx, frames;
+	unsigned char received[8192];
+
+	idx = MSG_ReadByte( msg ) + 1;
+
+	frames = MSG_ReadByte( msg );
+
+	size = MSG_ReadShort( msg );
+	size = Q_min( size, 8192 );
```
a1batross, line 1710: Use `sizeof( received )` here

engine/client/cl_parse.c
```diff
@@ -3127,6 +3154,7 @@ void CL_ParseLegacyServerMessage( sizebuf_t *msg, qboolean normal_message )
 			break;
 		case svc_voicedata:
 			CL_ParseVoiceData( msg );
```
a1batross, line 3156: Will legacy servers send `svc_voicedata`?

Highly doubt about it.

engine/client/cl_view.c
```diff
@@ -286,9 +286,6 @@ qboolean V_PreRender( void )
 	if( !ref.initialized )
 		return false;
 
-	if( host.status == HOST_NOFOCUS )
```
a1batross, line 289: What's the reason of this change?

engine/client/s_main.c
```diff
@@ -1860,6 +1863,33 @@ void S_SoundInfo_f( void )
...
 }
 
+/*
+=================
+S_VoiceRecordStart_f
+=================
+*/
+void S_VoiceRecordStart_f( void )
+{
+	if( cls.state != ca_active || cls.legacymode )
```
a1batross, line 1873: Since there is a check for legacy mode servers, we absolutely don't need any `svc_voicedata/init` parsing for legacy protocol above.

engine/client/voice.c
```diff
@@ -0,0 +1,368 @@
+#include "voice.h"
```
a1batross, line 1: Add GPL notice for yourself and @Velaron

engine/client/voice.c
```diff
@@ -0,0 +1,368 @@
...
+
+wavdata_t *input_file;
+fs_offset_t input_pos;
+
+voice_state_t voice;
+
+CVAR_DEFINE_AUTO( voice_enable, "1", FCVAR_ARCHIVE, "enable voice chat" );
```
a1batross, line 8: This and other client-specific cvars probably should FCVAR_PRIVILEGED, so server will not change them arbitrarily.

engine/client/voice.c
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	if ( !Voice_IsRecording() )
+		voice.is_recording = VoiceCapture_RecordStart();
+
+	if ( Voice_IsRecording() )
+		Voice_Status( -1, true );
+}
+
+void Voice_AddIncomingData( int ent, byte *data, uint size, uint frames )
```
a1batross, line 309: Isn't second parameter better be `const`, so you avoid cast later?

engine/client/voice.h
```diff
@@ -0,0 +1,70 @@
+#ifndef VOICE_H
```
a1batross, line 1: Add GPL notice for you and @Velaron

engine/platform/stub/s_stub.c
```diff
@@ -94,5 +94,20 @@ void SNDDMA_Shutdown( void )
 	}
 }
 
+qboolean VoiceCapture_Init( void )
```
a1batross, line 97: Add appropriate stubs to `engine/platform/linux/s_alsa.c` as well

engine/server/sv_client.c
```diff
@@ -2561,6 +2561,58 @@ void SV_ParseCvarValue2( sv_client_t *cl, sizebuf_t *msg )
...
+	MSG_ReadBytes( msg, received, size );
+
+	for( i = 0, cur = svs.clients; i < svs.maxclients; i++, cur++ )
+	{
+		if ( cur->state < cs_connected && cl != cur )
+			continue;
+		
+		length = size;
+
+		if ( MSG_GetNumBytesLeft( &cur->datagram ) < length + 6 )
```
a1batross, line 2602: Wtf `length + 6` means here?

engine/server/sv_client.c
```diff
@@ -2561,6 +2561,58 @@ void SV_ParseCvarValue2( sv_client_t *cl, sizebuf_t *msg )
...
+	client = cl - svs.clients;
+
+	if ( size > sizeof( received ) )
+	{
+		Con_DPrintf( "SV_ParseVoiceData: invalid incoming packet.\n" );
+		SV_DropClient( cl, false );
+		return;
+	}
+
+	if ( !Cvar_VariableInteger( "sv_voiceenable" ) )
```
a1batross, line 2590: You can avoid cvar lookup here with just `sv_voiceenable.value`

engine/server/sv_client.c
a1batross: At least, leave a comment that 6 is a number of bytes for other parts of message. :)

engine/server/sv_init.c
```diff
@@ -386,6 +386,18 @@ void SV_CreateResourceList( void )
...
+/*
+================
+SV_WriteVoiceCodec
+================
+*/
+void SV_WriteVoiceCodec( sizebuf_t *msg )
+{
+	MSG_BeginServerCmd( msg, svc_voiceinit );
+	MSG_WriteString( msg, "opus" );
+	MSG_WriteByte( msg, Cvar_VariableInteger( "sv_voicequality" ));
```
a1batross, line 398: Same, you can just do `sv_voicequality.value` instead.

engine/wscript
```diff
@@ -165,8 +165,10 @@ def build(bld):
 			'client/*.c',
 			'client/vgui/*.c',
 			'client/avi/*.c'])
+	
+		libs.append('opus')
 
-	includes = ['common', 'server', 'client', 'client/vgui', 'tests', '.', '../public', '../common', '../filesystem', '../pm_shared' ]
+	includes = ['common', 'server', 'client', 'client/vgui', 'tests', '.', '../public', '../common', '../filesystem', '../pm_shared', '../3rdparty/opus/opus/include' ]
```
a1batross, line 171: Make opus include path is part of it's uselib, so it will be added automatically by waf. See `sdl2.py` for manual usage of uselib.

wscript
```diff
@@ -54,6 +54,7 @@ class Subproject:
 		return True
 
 SUBDIRS = [
+	Subproject('3rdparty/opus'),
```
a1batross, line 57: Probably better be `mandatory = True`? For now, this code doesn't seem to have alternative code path for no voice support anyway.

a1batross (changes requested):

engine/client/cl_view.c
SNMetamorph: Guess it came from Velaron's fork

Velaron: IIRC, it was causing sound issues when the game window is inactive (not minimized) as I was testing the voice chat basically with myself on the same machine

engine/client/cl_parse.c
SNMetamorph: Maybe allocating such a big buffer on stack is a bad idea? I can make it static

engine/client/cl_parse.c
SNMetamorph: Yes, I think too, will remove this

engine/wscript
SNMetamorph: I tried:
```
def configure(conf):
	if conf.check_pkg('opus', 'OPUS', OPUS_CHECK, fatal = False):
		conf.env.HAVE_OPUS = True
		return

	if not conf.path.find_dir('opus') or not conf.path.find_dir('opus/src'):
		conf.fatal('Can\'t find opus submodule. Run `git submodule update --init --recursive`.')
		return

	conf.env.INCLUDES_OPUS = [
		os.path.abspath('opus/include'),
	]
```
But this doesn't work, what I do wrong?

a1batross: Ah, I forgot you're compiling it yourself. Sure, you can do this way.

But you also have export_includes set up already. It should work just right away, I even tested building, and it works without changing include paths for engine.

engine/client/cl_parse.c
a1batross: Not at all?

engine/client/cl_view.c
a1batross: @Velaron I don't remember the actual desired behavior in this case.

I think for the sake of debugging, it's better to limit fps and mute only in minimized state. But then NOFOCUS state make no sense.

SNMetamorph: What do with this? Keep or remove?

engine/client/cl_parse.c
SNMetamorph: I made it as static, keep as it is?

a1batross: Use sizeof. It's fine to have 8k buffer on stack, I think we have somewhere even more than that.

engine/client/cl_view.c
a1batross: It's ok.

engine/client/cl_parse.c
SNMetamorph: Ok fixed

a1batross (approved):

PR #985, opened 2022-08-20
engine: added audio backend print to s_info command

SNMetamorph:


engine/client/s_main.c
```diff
@@ -1927,6 +1927,7 @@ qboolean S_Init( void )
 	Cmd_AddCommand( "spk", S_SayReliable_f, "reliable play a specified sententce" );
 	Cmd_AddCommand( "speak", S_Say_f, "playing a specified sententce" );
 
+	dma.backendName = NULL;
```
a1batross, line 1930: Do you really need to set it to NULL beforehand?

engine/platform/sdl/s_sdl.c
```diff
@@ -141,6 +141,7 @@ qboolean SNDDMA_Init( void )
 	Con_Printf( "Using SDL audio driver: %s @ %d Hz\n", SDL_GetCurrentAudioDriver( ), obtained.freq );
 
 	dma.initialized = true;
+	dma.backendName = "SDL";
```
a1batross, line 144: For SDL, we probably can write something like "SDL (%s)", where %s is SDL_GetCurrentAudioDriver output.

engine/client/s_main.c
SNMetamorph: I think better to do it, that's will protect if dma.backendName won't set in some of audio backends, for example

engine/platform/sdl/s_sdl.c
SNMetamorph: Yes, but then we need create char array for that in s_sdl.c

SNMetamorph: I did it

engine/client/s_main.c
a1batross: @SNMetamorph we better not have such backends at all. :)

Also, globals are zero initialized anyway by standard.

SNMetamorph: Ok

nekonomicon: May be set "None" as backend by default?

SNMetamorph: Yes

a1batross (approved):

PR #990, opened 2022-08-21
engine: client: enable interpolation of local player angles

SNMetamorph:
To look player model in third person look smooth

### DIFF
--- a/engine/client/cl_frame.c
+++ b/engine/client/cl_frame.c
@@ -534,7 +534,7 @@ void CL_ComputePlayerOrigin( cl_entity_t *ent )
 	vec3_t	origin;
 	vec3_t	angles;
 
-	if( !ent->player || ent->index == ( cl.playernum + 1 ))
+	if( !ent->player )
 		return;
 
 	if( cl_nointerp->value > 0.f )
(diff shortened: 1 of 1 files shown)


a1batross:
I'm sure it will break something else, since it modifies both origin and angles.

SNMetamorph:
Otherwise, in that case origin will be overwritten with subsequent code. The only thing I can suggest is make functions that doesn't modify entity state to pass in const cl_entity_t pointer

SNMetamorph:
I tested this and it's ok. Angles are interpolated, but origin didn't affected at all

a1batross:
It's not about const or not. And you added const modifier where I didn't asked to and half of the patch is ugly casts now.

Updating angles from interpolated data is just unnecessary by design. This is local data, angles can be easily taken from input for example. That's IF we're actually turning to same direction we're looking into. I think it can be up to client.dll to interpolate angles or calculate them from input.

Nonetheless, I approve this patch anyway. But please, update your patch to be minimal as possible.

SNMetamorph:
> It's not about const or not. And you added const modifier where I didn't asked to and half of the patch is ugly casts now.
> 
> Updating angles from interpolated data is just unnecessary by design. This is local data, angles can be easily taken from input for example. That's IF we're actually turning to same direction we're looking into. I think it can be up to client.dll to interpolate angles or calculate them from input.
> 
> Nonetheless, I approve this patch anyway. But please, update your patch to be minimal as possible.

Ok

a1batross (approved):

PR #1029, opened 2022-09-09
Sleep timers implementation

SNMetamorph:
Need a review, especially on POSIX timer implementation

### DIFF
--- a/engine/common/host.c
+++ b/engine/common/host.c
@@ -325,18 +325,6 @@ double Host_CalcFPS( void )
 		if( fps == 0.0 ) fps = MAX_FPS;
 		fps = bound(MIN_FPS, fps, MAX_FPS);
 	}
-
-	// probably left part of this condition is redundant :-)
-	if( host.type != HOST_DEDICATED && Host_IsLocalGame() && !CL_IsTimeDemo() )
-	{
-		// ajdust fps for vertical synchronization
-		if( CVAR_TO_BOOL( gl_vsync ))
-		{
-			if( vid_displayfrequency->value != 0.0f )
-				fps = vid_displayfrequency->value;
-			else fps = 60.0; // default
-		}
-	}
 #endif
 
 	return fps;
--- a/engine/common/host.c
+++ b/engine/common/host.c
@@ -53,6 +53,7 @@ struct tests_stats_s tests_stats;
 CVAR_DEFINE( host_developer, "developer", "0", FCVAR_FILTERABLE, "engine is in development-mode" );
 CVAR_DEFINE_AUTO( sys_timescale, "1.0", FCVAR_CHEAT|FCVAR_FILTERABLE, "scale frame time" );
 CVAR_DEFINE_AUTO( sys_ticrate, "100", 0, "framerate in dedicated mode" );
+CVAR_DEFINE_AUTO( sys_usespinlock, "0", FCVAR_FILTERABLE, "using spinlock instead of high-precision timer to count time between frames" );
 
 convar_t	*host_serverstate;
 convar_t	*host_gameloaded;
@@ -268,35 +269,137 @@ void Host_AbortCurrentFrame( void )
 }
 
 /*
+=================
+Host_IsLocalGame
+
+singleplayer game detect
+=================
+*/
+qboolean Host_IsLocalGame( void )
+{
+	if( SV_Active() )
+	{
+		return ( SV_GetMaxClients() == 1 ) ? true : false;
+	}
+	else
+	{
+		return ( CL_GetMaxClients() == 1 ) ? true : false;
+	}
+}
+
+qboolean Host_IsLocalClient( void )
+{
+	// only the local client have the active server
+	if( CL_Initialized() && SV_Initialized() )
+		return true;
+	return false;
+}
+
+/*
+===================
+Host_CalcFPS
+
+compute actual FPS for various modes
+===================
+*/
+double Host_CalcFPS( void )
+{
+	double	fps = 0.0;
+
+	if( Host_IsDedicated() )
+	{
+		fps = sys_ticrate.value;
+	}
+#if !XASH_DEDICATED
+	else if( CL_IsPlaybackDemo() || CL_IsRecordDemo() ) // NOTE: we should play demos with same fps as it was recorded
+	{
+		fps = CL_GetDemoFramerate();
+	}
+	else if( Host_IsLocalGame() )
+	{
+		fps = host_maxfps->value;
+	}
+	else
+	{
+		fps = host_maxfps->value;
+		if( fps == 0.0 ) fps = MAX_FPS;
+		fps = bound(MIN_FPS, fps, MAX_FPS);
+	}
+
+	// probably left part of this condition is redundant :-)
+	if( host.type != HOST_DEDICATED && Host_IsLocalGame() && !CL_IsTimeDemo() )
+	{
+		// ajdust fps for vertical synchronization
+		if( CVAR_TO_BOOL( gl_vsync ))
+		{
+			if( vid_displayfrequency->value != 0.0f )
+				fps = vid_displayfrequency->value;
+			else fps = 60.0; // default
+		}
+	}
+#endif
+
+	return fps;
+}
+
+/*
 ==================
 Host_CheckSleep
 ==================
 */
 void Host_CheckSleep( void )
 {
-	int sleeptime = host_sleeptime->value;
+	double fps;
+	int sleeptime;
+	qboolean use_spinlock;
 
+	fps = Host_CalcFPS();
+	sleeptime = host_sleeptime->value;
+	use_spinlock = sys_usespinlock.value > 0.0f;
+	
 	if( Host_IsDedicated() )
 	{
 		// let the dedicated server some sleep
-		Sys_Sleep( sleeptime );
+		if( use_spinlock )
+			Sys_Sleep( sleeptime );
+		else
+			Platform_Delay( 1.0 / fps );
 	}
 	else
 	{
 		if( host.status == HOST_NOFOCUS )
 		{
-			if( SV_Active() && CL_IsInGame( ))
-				Sys_Sleep( sleeptime ); // listenserver
-			else Sys_Sleep( 20 ); // sleep 20 ms otherwise
+			if( SV_Active() && CL_IsInGame() )
+			{
+				// listenserver
+				if( use_spinlock )
+					Sys_Sleep( sleeptime ); 
+				else
+					Platform_Delay( 1.0 / fps );
+			}
+			else
+			{
+				// sleep 20 ms otherwise
+				if( use_spinlock )
+					Sys_Sleep( 20 ); 
+				else
+					Platform_Delay( 1.0 / 50.0 );
+			}
 		}
 		else if( host.status == HOST_SLEEP )
 		{
 			// completely sleep in minimized state
-			Sys_Sleep( 20 );
+			if( use_spinlock )
+				Sys_Sleep( 20 );
+			else
+				Platform_Delay( 1.0 / 50.0 );
 		}
 		else
 		{
-			Sys_Sleep( sleeptime );
+			if( use_spinlock )
+				Sys_Sleep( sleeptime );
+			else
+				Platform_Delay( 1.0 / fps );
 		}
 	}
 }
@@ -469,33 +572,6 @@ void Host_Minimize_f( void )
 
 /*
 =================
-Host_IsLocalGame
-
-singleplayer game detect
-=================
-*/
-qboolean Host_IsLocalGame( void )
-{
-	if( SV_Active( ))
-	{
-		return ( SV_GetMaxClients() == 1 ) ? true : false;
-	}
-	else
-	{
-		return ( CL_GetMaxClients() == 1 ) ? true : false;
-	}
-}
-
-qboolean Host_IsLocalClient( void )
-{
-	// only the local client have the active server
-	if( CL_Initialized( ) && SV_Initialized( ))
-		return true;
-	return false;
-}
-
-/*
-=================
 Host_RegisterDecal
 =================
 */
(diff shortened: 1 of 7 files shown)


engine/platform/win32/sys_win.c
```diff
@@ -46,6 +62,92 @@ qboolean Sys_DebuggerPresent( void )
...
+		0,                      // use default stack size
+		Sys_TimerThread,		// thread function name
+		0,						// argument to thread function
+		0,                      // use default creation flags
+		&g_timer.iThreadID);
+}
+
+void Platform_TimerShutdown( void )
+{
+	CloseHandle( g_timer.hThread );
```
nekonomicon, line 145: You can close thread handle when you create thread.
One correct way to stop thread - return error code from thread.
CloseHandle just clears thread handle and nothing more.

nekonomicon (changes requested):

engine/platform/win32/sys_win.c
SNMetamorph: Should I stop thread in this case at all?

nekonomicon: You can stop thread via [ExitThread](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-exitthread) and [TerminateThread](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminatethread) callbacks, but it's not recommended:

> TerminateThread is used to cause a thread to exit. When this occurs, the target thread has no chance to execute any user-mode code. DLLs attached to the thread are not notified that the thread is terminating. The system frees the thread's initial stack.

> Windows Server 2003 and Windows XP:  The target thread's initial stack is not freed, causing a resource leak.

> TerminateThread is a dangerous function that should only be used in the most extreme cases. You should call TerminateThread only if you know exactly what the target thread is doing, and you control all of the code that the target thread could possibly be running at the time of the termination. For example, TerminateThread can result in the following problems:

>    If the target thread owns a critical section, the critical section will not be released.
>    If the target thread is allocating memory from the heap, the heap lock will not be released.
>    If the target thread is executing certain kernel32 calls when it is terminated, the kernel32 state for the thread's process could be inconsistent.
 >   If the target thread is manipulating the global state of a shared DLL, the state of the DLL could be destroyed, affecting other users of the DLL.

Or you can save thread handle and reuse it later.

SNMetamorph: So I think it's better to stop thread by setting flag in main thread and checking it in timer thread, instead of all this nightmare xD

engine/platform/posix/sys_posix.c
```diff
@@ -169,3 +182,63 @@ void Platform_Sleep( int msec )
...
 }
 #endif // XASH_TIMER == TIMER_POSIX
+
+static void *Sys_TimerThread( void *v )
+{
+	struct timespec time, time1, time2, time3 = { 0,0 };
+	time.tv_sec = 0;
+	time.tv_nsec = g_timer.frametime * 1e9;
+	
+	while (1)
```
a1batross, line 192: Please reformat this file :)

engine/platform/posix/sys_posix.c
```diff
@@ -169,3 +182,63 @@ void Platform_Sleep( int msec )
...
+	while (!g_timer.flag)
+		pthread_cond_wait(&g_timer.tsignal, &g_timer.tmutex);
+	g_timer.flag = 0;
+	pthread_mutex_unlock(&g_timer.tmutex);
+}
+
+void Platform_TimerInit( void )
+{
+	g_timer.flag = 1;
+	g_timer.frametime = 1.0 / host_maxfps->value;
```
a1batross, line 237: What if maxfps value is zero? Will it just run with +infinity?

Otherwise, what if this is a dedicated server? Dedicated servers don't depend on fps_max, but on sys_ticrate instead. I'm not sure if dedicated servers need this timer code at all, they're servers after all.

engine/platform/posix/sys_posix.c
```diff
@@ -169,3 +182,63 @@ void Platform_Sleep( int msec )
...
+	pthread_create(&g_timer.thread, NULL, Sys_TimerThread, NULL);
+}
+
+void Platform_TimerShutdown( void )
+{
+	// ... what we should do here? i need help for this
```
a1batross, line 243: You don't generally need to clean pthreads. It's enough for routine to stop. In our case, the main thread will close child threads anyway.

engine/platform/win32/sys_win.c
```diff
@@ -46,6 +62,92 @@ qboolean Sys_DebuggerPresent( void )
...
+	ReleaseMutex( g_timer.hMutex ); // unlock mutex
+	WaitForSingleObject( g_timer.hEvent, INFINITE ); // wait when event will be triggered
+}
+
+void Platform_TimerInit( void )
+{
+	Sys_SetTimerHighResolution();
+	g_timer.hTimer = CreateWaitableTimer(NULL, TRUE, NULL);
+	g_timer.hEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
+	g_timer.flInterval = 1.0 / 60.0;
```
a1batross, line 127: Same as for `Platform_TimerInit` for posix threads about dedicated servers.

a1batross: Also, maybe we need to target maxfps value here? Instead of hardcoded 60 Hz?

a1batross (changes requested):
While I'm in favor of these changes, this needs a lot of work done considering all cases.

Maybe we don't need to run timer thread for dedicated servers? In case of VSync we can assume that engine runs unlimited frames, as waiting done in drivers in SwapBuffers anyway, and no frame limit adjusting should be done, especially with dead cvars.

Don't forget that thanks to GoldSrc/Quake legacy, we have multiple cvars controlling desired frame limit, `sys_ticrate` in dedicated and `fps_max` in full engine. `fps_max` cvar can be zero, which just means unlimited frames. Please avoid divisions by zero.

Also, I personally don't like the wording of `sys_spinlock` cvar. Spin lock is a synchronization primitive, it's term that more applicable to threading. In our single threaded game engine, we just run busy-wait loop until system timers hits desired target. We only added small sleeps here.

engine/platform/posix/sys_posix.c
```diff
@@ -14,13 +14,26 @@ GNU General Public License for more details.
...
+typedef struct {
+	double frametime;
+	pthread_mutex_t tmutex;
+	qboolean flag;
+	pthread_t thread;
+	pthread_cond_t tsignal;
+} timer_posix_t;
+
+static timer_posix_t g_timer;
+g_timer.tmutex = PTHREAD_MUTEX_INITIALIZER;
```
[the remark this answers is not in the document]
SNMetamorph: Especially this place throw error:
```
2022-09-09T13:36:13.7909928Z ../engine/platform/posix/sys_posix.c:34:8: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘.’ token
2022-09-09T13:36:13.7911792Z  g_timer.tmutex = PTHREAD_MUTEX_INITIALIZER;
2022-09-09T13:36:13.7912429Z         ^
2022-09-09T13:36:13.7913104Z ../engine/platform/posix/sys_posix.c:35:8: error: expected ‘=’, ‘,’, ‘;’, ‘asm’ or ‘__attribute__’ before ‘.’ token
2022-09-09T13:36:13.7913469Z  g_timer.tsignal = PTHREAD_COND_INITIALIZER;
2022-09-09T13:36:13.8132501Z         ^
2022-09-09T13:36:13.8132676Z 
```
How to fix it?

engine/common/host.c
```diff
@@ -267,36 +268,138 @@ void Host_AbortCurrentFrame( void )
...
+		fps = bound(MIN_FPS, fps, MAX_FPS);
+	}
+
+	// probably left part of this condition is redundant :-)
+	if( host.type != HOST_DEDICATED && Host_IsLocalGame() && !CL_IsTimeDemo() )
+	{
+		// ajdust fps for vertical synchronization
+		if( CVAR_TO_BOOL( gl_vsync ))
+		{
+			if( vid_displayfrequency->value != 0.0f )
```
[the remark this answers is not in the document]
SNMetamorph: You about `vid_displayfrequency`? Ok I can remove this check

engine/common/host.c
```diff
@@ -267,36 +268,138 @@ void Host_AbortCurrentFrame( void )
...
+					Sys_Sleep( 20 ); 
+				else
+					Platform_Delay( 1.0 / 50.0 );
+			}
 		}
 		else if( host.status == HOST_SLEEP )
 		{
 			// completely sleep in minimized state
-			Sys_Sleep( 20 );
+			if( use_spinlock )
```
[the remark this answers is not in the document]
SNMetamorph: Yes, it'll be better

engine/platform/posix/sys_posix.c
SNMetamorph: What do you mean exactly?

engine/platform/posix/sys_posix.c
SNMetamorph: I think it's better to replace this with some hardcoded value, it's just for first frame, anyway

engine/platform/win32/sys_win.c
SNMetamorph: > Also, maybe we need to target maxfps value here? Instead of hardcoded 60 Hz?

Don't think it's better than hardcode, in this case

SNMetamorph:
> While I'm in favor of these changes, this needs a lot of work done considering all cases.
> 
> Maybe we don't need to run timer thread for dedicated servers? In case of VSync we can assume that engine runs unlimited frames, as waiting done in drivers in SwapBuffers anyway, and no frame limit adjusting should be done, especially with dead cvars.
> 
> IMHO, timer thread should run only if frame limiter is active.
> 
> Don't forget that thanks to GoldSrc/Quake legacy, we have multiple cvars controlling desired frame limit, `sys_ticrate` in dedicated and `fps_max` in full engine. `fps_max` cvar can be zero, which just means unlimited frames. Please avoid divisions by zero.
> 
> Also, I personally don't like the wording of `sys_spinlock` cvar. Spin lock is a synchronization primitive, it's term that more applicable to threading. In our single threaded game engine, we just run busy-wait loop until system timers hits desired target. We only added small sleeps here.

You mean that for dedicated server we should use busy-wait loop instead timers?

SNMetamorph:
@mittorn why timers use separate thread? For what purpose?

engine/common/host.c
```diff
@@ -268,35 +269,132 @@ void Host_AbortCurrentFrame( void )
...
+		fps = host_maxfps->value;
+		if( fps == 0.0 ) fps = MAX_FPS;
+		fps = bound(MIN_FPS, fps, MAX_FPS);
+	}
+#endif
+
+	return fps;
+}
+
+static qboolean Host_IgnoreTimers()
```
nekonomicon, line 333: static qboolean Host_IgnoreTimers( void )

nekonomicon (changes requested):

engine/platform/posix/sys_posix.c
nekonomicon: Use Quake C style format.

nekonomicon (changes requested):

engine/platform/posix/sys_posix.c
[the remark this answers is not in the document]
nekonomicon: Use lazy initialization or use struct/array initializer:
```
array[] = {val1,val2,val3,val4};
```

nekonomicon (changes requested):

engine/platform/win32/sys_win.c
```diff
@@ -46,6 +62,92 @@ qboolean Sys_DebuggerPresent( void )
...
+	pfnNtQueryTimerResolution_t pNtQueryTimerResolution;
+	
+	hndl = GetModuleHandle( "ntdll.dll" );
+	pNtSetTimerResolution = (pfnNtSetTimerResolution_t)GetProcAddress( hndl, "NtSetTimerResolution" );
+	pNtQueryTimerResolution = (pfnNtQueryTimerResolution_t)GetProcAddress( hndl, "NtQueryTimerResolution" );
+
+	if( pNtSetTimerResolution && pNtQueryTimerResolution )
+	{
+		// undocumented NT API functions, for some mystery reason works better than timeBeginPeriod
+		ULONG min, max, cur;
```
nekonomicon, line 78: Wrong format.

nekonomicon (changes requested):

engine/platform/win32/sys_win.c
```diff
@@ -46,6 +62,92 @@ qboolean Sys_DebuggerPresent( void )
...
+	WaitForSingleObject( g_timer.hMutex, INFINITE ); // lock mutex
+	g_timer.flInterval = time;
+	ReleaseMutex( g_timer.hMutex ); // unlock mutex
+	WaitForSingleObject( g_timer.hEvent, INFINITE ); // wait when event will be triggered
+}
+
+void Platform_TimerInit( void )
+{
+	Sys_SetTimerHighResolution();
+	g_timer.hTimer = CreateWaitableTimer(NULL, TRUE, NULL);
```
nekonomicon, line 125: Wrong format

nekonomicon (changes requested):

engine/platform/win32/sys_win.c
nekonomicon: Anyway, you don't need to save thread handle and probably no reason to clear all other open handles if process will be stopped.

nekonomicon (changes requested):

engine/platform/win32/sys_win.c
SNMetamorph: What do you mean?

nekonomicon: Add spaces between brackets and other characters

nekonomicon (changes requested):

SNMetamorph:
Finally, it compiles on *nix. Can somebody test this all?

nekonomicon:
POSIX timers has a bad precision under FreeBSD.
Or may be pthreads callbacks is too slow.
I see such fps: 
```
58-59 for fps_max 60
70 for fps_max 72
97 for fps_max 100
```

a1batross:
Superseded by f469b56b93d712db502477604b629a30f6c18a3d.

This way we let engine sleep once between frames but also not hurting fps limit cap precision.

SNMetamorph:
Sad that this code was not used :(

PR #1031, opened 2022-09-10
Reorganize directories

a1batross:
Root repository folder feels cluttered a bit, so there is small reorganization.

Let's merge this quicker, don't want to turn it into yet another bike shedding discussion.


SNMetamorph:
But what about commits history and git blame?

a1batross:
@SNMetamorph it's not an issue since Git preserves history no matter the file path

PR #1046, opened 2022-09-19
NaN detect

a1batross:
This is absolutely horrible.

But it still helped to detect two bugs earlier, before the engine stuck in endless loop or crashed.


a1batross:
I don't think we need this in upstream. This PR will be closed, but if anyone wants to debug NaNs, this can be rebased and reapplied.

PR #1088, opened 2022-11-02
filesystem: set root directory as static

kungfulon:
Counter-Strike uses COM_ExpandFileName to get full path to filesystem_stdio. When reading liblist.gam, the engine will clear the search path and the root search path will be cleared. Set it as static so it will be excluded from clearing.

First patch to prepare for VGUI2 support.

### DIFF
--- a/filesystem/filesystem.c
+++ b/filesystem/filesystem.c
@@ -1452,7 +1452,7 @@ qboolean FS_InitStdio( qboolean caseinsensitive, const char *rootdir, const char
 	}
 
 	// build list of game directories here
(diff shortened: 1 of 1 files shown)


a1batross:
Won't this break because root directory will be considered for search after game was initialized?

a1batross:
Well if GoldSrc does it, there isn't much sense to not follow it.

kungfulon:
Yeah, this commit will make xash follow that behavior, so I think it's ok to merge?

a1batross:
Sure, indeed. LGTM then.

PR #1092, opened 2022-11-04
common: com_model: changed common structs reserved fields type to intptr_t

SNMetamorph:
To make porting mod to 64-bit less painful


a1batross:
I still don't think we should modify the structs that's public and used in SDK in both GoldSrc and Xash. My idea is to put only one pointer in reserved area in this case.

We should move towards more compatibility with GoldSrc, than have custom extensions in random places and have potentially broken mods. For non Half-Life stuff there will be XashNT... some day?

Migration from GoldSrc to Xash must be seamless as possible, down to minifying all changes to support 64-bit address space to one line.

a1batross:
In the future, we could even probably generate GoldSrc structs from debug info and get rid of HLSDK code in main engine repository.

SNMetamorph:
Ok, I reverted `texture_t` expansion

SNMetamorph:
So what about this?

a1batross (approved):

Issue #947, opened 2022-08-02
Eliminate FreeLibrary/dlclose


a1batross:
dlclose and FreeLibrary functions do not guarantee that the library was really unloaded.

It may lead to crashes, especially with global variables, when these don't get reinitialized to default value. Simplest example is pm_shared_initialized global in HLSDK code. If it somehow was marked as NODELETE, it will instantly crash the engine.

glibc and windows tries to unload the library but for example musl do not implement that at all. Also in Android NDK 25 release notes there is an bug mentioned where using dlclose can lead to crashes on some Android versions. NODELETE flag or eliminating dlclose is the only fix.

Some mods also are not intended to be reloaded all the time, like Metamod. I also get easy crashes with halflife-unified-sdk.

We already switched to execv() call for changing game because of dlclose/FreeLibrary being unreliable.

Here is the list of some troublesome parts that needs to be fixed as soon as possible:
- [ ] RefAPIs GetHumanReadable call. I guess we cannot collect renderers descriptions this way, loading the library, calling the export and unloading.
- [ ] pfnCheckGameDll menu call. It's used for menu to detect if there is a usable server library, or we can only connect to multiplayer games.
- [ ] settings.scr cvars collect. Engine uses that to generate listenserver.cfg
- [ ] Engine still can unload server library. I believe we don't need this in general, goldsrc never unloads it.

a1batross:
cc @nekonomicon @SNMetamorph @mittorn

SNMetamorph:
Yea I agree that this should be done.

nekonomicon:
IDK why need to unload libraries, It's just waste of time.

a1batross:
@nekonomicon glad to see you again! :)

I like loading libraries dynamicly, it makes sense for plugin architecture.

But unloading is unnecessary. We can only restart the engine.

Though guys from RTX branch @zgdump @0x4E69676874466F78 would not agree, since they implemented hot renderer swap.

zgdump:
Eh, I really wanted to switch renders...
But I still haven't figured out why after a couple of ref switches everything explodes in random places of the engine

a1batross:
@zgdump well, I doubt we can do anything with runtime that doesn't cleanup or even do anything at all after dlclose() and FreeLibrary().

Crashes and bugs just unavoidable with them.

zgdump:
[...]

PR #1101, opened 2022-11-10
Internal vgui support rework

a1batross:
VGUI part of #947 

This PR introduces new `internal_vgui_support` key in `gameinfo.txt` that signals the engine to skip loading VGUI until the client.dll is loaded.

Linking vgui_support code into client.dll makes sense if game developer want to replace VGUI by something else, or include a specific version of it. Like PrimeXT uses ImGUI now by re-using VGUI support APIs, and tf15-client statically links with OpenVGUI to have same interface across all platforms.

cc @SNMetamorph @Velaron 

also cc @kungfulon, we need to decide on VGUI startup sequence. VGUI1 cannot be loaded after client.dll. In mods, VGUI1 can be called in C++ constructors, before API was initialized. I placed a check in SCR_VidInit so VGUI1 won't get loaded here.

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -3937,10 +3937,6 @@ qboolean CL_LoadProgs( const char *name )
 	clgame.mempool = Mem_AllocPool( "Client Edicts Zone" );
 	clgame.entities = NULL;
 
-	// NOTE: important stuff!
-	// vgui must startup BEFORE loading client.dll to avoid get error ERROR_NOACESS
-	// during LoadLibrary
-	VGui_Startup( name, gameui.globals->scrWidth, gameui.globals->scrHeight );
 
 	// a1ba: we need to check if client.dll has direct dependency on SDL2
 	// and if so, disable relative mouse mode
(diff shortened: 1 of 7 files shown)


SNMetamorph:
> Linking vgui_support code into client.dll makes sense if game developer want to replace VGUI by something else, or include a specific version of it. Like PrimeXT uses ImGUI now by re-using VGUI support APIs, and tf15-client statically links with OpenVGUI to have same interface across all platforms.

I think this is good idea, but how is was working before?

a1batross:
@SNMetamorph you mean probing client.dll for InitVGUISupportAPI? 

It's almost a magic. There are many bug reports where mods just straight up crashes

kungfulon:
@a1batross,

About VGUI1 and VGUI2 startup sequence:

For VGUI1 we know that it has to be loaded before `client.dll`, that's the current logic yeah. But for VGUI2, there are some objects that are used in VGUI2 viewport of `client.dll` will be constructed when calling `client.dll!Initialize`. So, to properly start VGUI2, we must only start *after* calling `client.dll!Initialize`.

Before this PR, the VGUI dll is actually loaded in `SCR_VidInit`, and the call in `CL_LoadProgs` would be the second one. Now that we made sure to only load VGUI dll in `CL_LoadProgs`, it is necessary to call `VGui_Startup` once again after calling `client.dll!Initialize`.

a1batross:
What if we make two separate functions then?
PreInit and PostInit?

These will may be included in new more future-proof vgui_support API.

пт, 11 нояб. 2022 г., 10:39 Bien Pham ***@***.***>:

kungfulon:
Yeah, 2 separate functions would save us from using hackish way to detect initialization of vgui systems. PreInit should be used only to initialize vgui_support stuff, while PostInit would be for VGUI2 to startup client viewport.

Maybe PreInit can still be named Startup?

a1batross:
@kungfulon sure, yes.

If this PR is ok, I'm merging it and will add needed exports for the new API. 

Existing should be dropped ASAP, unless it didn't shoot somebody's legs off.

PR #1119, opened 2022-11-22
BMP file loading fixes

SNMetamorph:


engine/common/imagelib/img_bmp.h
```diff
@@ -25,7 +25,12 @@ GNU General Public License for more details.
 #define BI_FILE_HEADER_SIZE 14
 #define BI_SIZE 40 // size of bitmap info header.
 #if !defined(BI_RGB)
-#define BI_RGB 0 // uncompressed RGB bitmap(defined in wingdi.h)
+#define BI_RGB		  0	// uncompressed RGB bitmap (defined in wingdi.h)
```
nekonomicon, line 28: May be use enumeration?

engine/common/imagelib/img_bmp.c
```diff
@@ -174,7 +180,7 @@ qboolean Image_LoadBMP( const char *name, const byte *buffer, fs_offset_t filesi
 		break;
 	}
 
-	estimatedSize = ( buf_p - buffer ) + ( image.width + padSize ) * image.height * ( bhdr.bitsPerPixel >> 3 );
+	estimatedSize = ( buf_p - buffer ) + image.width * image.height * ( bhdr.bitsPerPixel >> 3 );
```
a1batross, line 183: Why did you remove padding here?

SNMetamorph: It causes invalid estimated size calculation on images which dimensions is not power of 2. And therefore throws error.

engine/common/imagelib/img_bmp.h
SNMetamorph: But why?

PR #1124, opened 2022-11-25
engine: common: minor code fixes in hpak.c

SNMetamorph:
Including fix of bug that causes writing crap in custom.hpk
изображение

### DIFF
--- a/engine/common/hpak.c
+++ b/engine/common/hpak.c
@@ -17,8 +17,8 @@ GNU General Public License for more details.
 #include "hpak.h"
 
 #define HPAK_MAX_ENTRIES	0x8000
-#define HPAK_MIN_SIZE	(1 * 1024)
-#define HPAK_MAX_SIZE	(128 * 1024)
+#define HPAK_ENTRY_MIN_SIZE	(512)
+#define HPAK_ENTRY_MAX_SIZE	(128 * 1024)
 
 typedef struct hash_pack_queue_s
 {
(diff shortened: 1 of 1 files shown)


engine/common/hpak.c
```diff
@@ -512,7 +512,7 @@ void HPAK_CheckSize( const char *filename )
 	Q_strncpy( pakname, filename, sizeof( pakname ));
 	COM_ReplaceExtension( pakname, ".hpk" );
 
-	if( FS_FileSize( pakname, false ) > ( maxsize * 1000000 ))
+	if( FS_FileSize( pakname, false ) > ( maxsize * 1048576 ))
```
a1batross, line 515: What's this magical number? 1024 * 1024?

engine/common/hpak.c
```diff
@@ -17,8 +17,8 @@ GNU General Public License for more details.
 #include "hpak.h"
 
 #define HPAK_MAX_ENTRIES	0x8000
-#define HPAK_MIN_SIZE	(1 * 1024)
-#define HPAK_MAX_SIZE	(128 * 1024)
+#define HPAK_ENTRY_MIN_SIZE	(512)
```
a1batross, line 20: Do we want to have minimal entry size here?

engine/common/hpak.c
```diff
@@ -325,7 +325,7 @@ void HPAK_AddLump( qboolean bUseQueue, const char *name, resource_t *pResource,
 	// make a new container
 	dstpak.count = srcpak.count + 1;
 	dstpak.entries = Z_Malloc( sizeof( hpak_lump_t ) * dstpak.count );
-	memcpy( dstpak.entries, srcpak.entries, srcpak.count );
+	memcpy( dstpak.entries, srcpak.entries, sizeof( hpak_lump_t ) * srcpak.count );
```
a1batross, line 328: Nice catch!

engine/common/hpak.c
SNMetamorph: You mean we just can set minimal entry size 0 and just remove this macro?

engine/common/hpak.c
SNMetamorph: Yes, it's 1024 * 1024, 1 megabyte

a1batross:
Nah, actually, it's fine

a1batross (approved):

PR #1129, opened 2022-11-26
engine: common: imagelib: img_png.c: initial APNG support.

nekonomicon:
Testing required.


a1batross:
I doubt we want animated logos, it would be a complete anarchy to enable them for everyone.

As for animated PNGs, it's better to implement WebM support instead and solve two problems (proper video playback and animated WON logo in main menu)

PR #1132, opened 2022-12-01
Support for PNG player decals

a1batross:
- [x] Tested in local play
- [ ] Tested in network play

Resolves https://github.com/FWGS/xash3d-fwgs/issues/1127 

Needs server update, as customization is validated on server side as well.

Can be easily abused, needs texture resolution limit (before decoding, as it can be heavy).


engine/client/cl_main.c
```diff
@@ -1206,38 +1207,44 @@ resource_t *CL_AddResource( resourcetype_t type, const char *name, int size, qbo
 
 void CL_CreateResourceList( void )
 {
-	char		szFileName[MAX_OSPATH];
+	char szFileName[MAX_OSPATH];
```
SNMetamorph, line 1210: Wrong identation

SNMetamorph:
Add fix to mainui from #1130 also.

a1batross:
@SNMetamorph this is out of scope but I fixed it. 

And of course rebased on top of png-decals branch, so PNG support is already in master.

Well, I don't see anything wrong with it anyway.

a1batross:
Was merged to master already but GitHub didn't catch it :(

PR #1139, opened 2022-12-04
Various engine fixes

SNMetamorph:


engine/client/cl_main.c
```diff
@@ -2893,7 +2898,7 @@ void CL_InitLocal( void )
 	name = Cvar_Get( "name", Sys_GetCurrentUser(), FCVAR_USERINFO|FCVAR_ARCHIVE|FCVAR_PRINTABLEONLY, "player name" );
 	model = Cvar_Get( "model", "", FCVAR_USERINFO|FCVAR_ARCHIVE, "player model ('player' is a singleplayer model)" );
 	cl_updaterate = Cvar_Get( "cl_updaterate", "20", FCVAR_USERINFO|FCVAR_ARCHIVE, "refresh rate of server messages" );
-	cl_dlmax = Cvar_Get( "cl_dlmax", "0", FCVAR_USERINFO|FCVAR_ARCHIVE, "max allowed outcoming fragment size" );
+	cl_dlmax = Cvar_Get( "cl_dlmax", "1200", FCVAR_USERINFO, "max allowed outcoming fragment size" );
```
a1batross, line 2901: Why you removed ARCHIVE flag from here?

It's expected to exist because it controls the fragment size, and as far as I know, it's used by some players to set the exact MTU.

a1batross: You better return it, as either bandwidth test can be skipped (by archived cvar as well), either it's value directly used for legacy mode.

a1batross:
by the way, you don't have to write file name as a commit tag, but rather a part of the code you changed. For example, all `s_` files can be named as just `sound`, or `cl_` as `client`.

engine/client/cl_main.c
SNMetamorph: Ok, I'll return it. I thought that storing `cl_dlmax` is useless, because packet size can vary among different servers, different times, etc. Maybe I understand it's purpose incorrectly, but why did it overwriting in engine code if it's just a user preference?

SNMetamorph: Anyway, I reverting these changes, maybe somebody another may resolve misunderstandings with `cl_dlmax`

SNMetamorph:
vnes pravki

a1batross:
spasiba

a1batross (approved):

PR #1174, opened 2022-12-18
Get rid of va.

nekonomicon:


a1batross:
That's. So. Awesome.

engine/client/cl_game.c
```diff
@@ -1727,7 +1728,7 @@ static int GAME_EXPORT pfnClientCmd( const char *szCmdString )
 	else
 	{
 		// will exec later
-		Q_strncat( host.deferred_cmd, va( "%s\n", szCmdString ), sizeof( host.deferred_cmd ));
+		Q_snprintf( host.deferred_cmd, sizeof( host.deferred_cmd ), "%s\n", szCmdString );
```
lewa-j, line 1731: It was append before

engine/client/s_main.c
```diff
@@ -1787,16 +1787,18 @@ void S_Music_f( void )
 
 		for( i = 0; i < 2; i++ )
 		{
-			const char *intro_path = va( "media/%s.%s", intro, ext[i] );
-			const char *main_path = va( "media/%s.%s", main, ext[i] );
+			string buf;
+
+			const char *intro_path = snva( buf, sizeof(string), "media/%s.%s", intro, ext[i] );
+			const char *main_path = snva( buf, sizeof(string), "media/%s.%s", main, ext[i] );
```
lewa-j, line 1793: Overwrites buffer used for intro_path

engine/common/cvar.c
```diff
@@ -1166,8 +1168,8 @@ void Cvar_List_f( void )
 			continue;
 
 		if( Q_colorstr( var->string ))
-			value = va( "\"%s\"", var->string );
-		else value = va( "\"^2%s^7\"", var->string );
+			value = snva( buf, sizeof(string), "\"%s\"", var->string );
+		else value = snva( buf, sizeof(string), "\"^2%s^7\"", var->string );
```
lewa-j, line 1172: Now variable "value" can be removed and "buffer" used instead

filesystem/VFileSystem009.cpp
```diff
@@ -145,7 +145,9 @@ class CXashFS : public IVFileSystem009
 
 	void CreateDirHierarchy( const char *path, const char *id ) override
 	{
-		FS_CreatePath( va( "%s/%s", IdToDir( id ), path )); // FS_CreatePath is aware of slashes
+		string dir, tempbuf;
+
+		FS_CreatePath( snva( tempbuf, sizeof(string), "%s/%s", IdToDir( dir, id ), path )); // FS_CreatePath is aware of slashes
```
a1batross, line 150: This is became an unreadable and absolutely ugly abomination with `snva`.

filesystem/VFileSystem009.cpp
```diff
@@ -449,10 +445,12 @@ class CXashFS : public IVFileSystem009
 
 	bool AddPackFile( const char *path, const char *id ) override
 	{
-		char *p = va( "%s/%s", IdToDir( id ), path );
-		CopyAndFixSlashes( p, path );
+		string dir, tempbuf;
+
+		Q_snprintf( tempbuf, sizeof(string), "%s/%s", IdToDir( dir, id ), path );
```
a1batross, line 450: Let's `sizeof` of variable than it's type.

Variable type may change, but sizeof may be forgotten.

filesystem/filesystem.c
```diff
@@ -1433,8 +1439,8 @@ qboolean FS_InitStdio( qboolean caseinsensitive, const char *rootdir, const char
 
 		for( i = 0; i < dirs.numstrings; i++ )
 		{
-			char *roPath = va( "%s" PATH_SPLITTER "%s" PATH_SPLITTER, fs_rodir, dirs.strings[i] );
-			char *rwPath = va( "%s" PATH_SPLITTER "%s" PATH_SPLITTER, fs_rootdir, dirs.strings[i] );
+			Q_snprintf( roPath, sizeof(string), "%s" PATH_SPLITTER "%s" PATH_SPLITTER, fs_rodir, dirs.strings[i] );
```
a1batross, line 1442: Limit the scope of `roPath`, `rwPath`.

filesystem/filesystem.c
```diff
@@ -2463,7 +2469,7 @@ Build direct path for file in the filesystem
 return NULL for file in pack
 ==================
 */
-const char *FS_GetDiskPath( const char *name, qboolean gamedironly )
+const char *FS_GetDiskPath( const char *name, char *buf, qboolean gamedironly )
```
a1batross, line 2472: It's too optimistic to think `buf` will never be overflowed. :)

ref/gl/gl_image.c
```diff
@@ -1586,11 +1587,11 @@ int GL_LoadTextureArray( const char **names, int flags )
 	for( i = 0; i < numLayers; i++ )
 	{
 		COM_FileBase( names[i], basename );
-		Q_strncat( name, basename, sizeof( name ) );
-		if( i != ( numLayers - 1 )) Q_strncat( name, "|", sizeof( name ));
+		len += Q_strncpy( &name[len], basename, sizeof( name ) - len );
```
a1batross, line 1590: Why not just keep `Q_strncat` for a while? There is too critical to be part of this PR.

filesystem/VFileSystem009.cpp
a1batross: This actually applies to every usage of `sizeof` here. It's just too optimistic.

a1batross (commented):
For now this looks like more of an automated convert which works as a proof of concept but sadly just not good enough.

I don't think there is a real need to keep the `va`-like functions like `snva` and `sva`. It just makes everything more cluttered and unreadable.

Sure, having `Q_snprintf` where `va` was used wouldn't be very much better than just `va`. But it's a price we pay to get rid of thread unsafe function.

Mindless convert like this also introduced a few possible bugs that I noted above.

engine/client/cl_parse.c
```diff
@@ -946,13 +949,13 @@ void CL_ParseServerData( sizebuf_t *msg )
 
 	// get splash name
 	if( cls.demoplayback && ( cls.demonum != -1 ))
-		Cvar_Set( "cl_levelshot_name", va( "levelshots/%s_%s", cls.demoname, refState.wideScreen ? "16x9" : "4x3" ));
-	else Cvar_Set( "cl_levelshot_name", va( "levelshots/%s_%s", clgame.mapname, refState.wideScreen ? "16x9" : "4x3" ));
+		Cvar_Set( "cl_levelshot_name", snva( buf, sizeof(string), "levelshots/%s_%s", cls.demoname, refState.wideScreen ? "16x9" : "4x3" ));
```
a1batross, line 952: Stuff like this much be more beatiful without `snva` for example.

engine/client/cl_game.c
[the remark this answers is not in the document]
lewa-j: \n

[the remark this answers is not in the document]
a1batross: Yep. Still, such commands better be concatenated.

a1batross:
If you really want to continue on this PR, there is one thing that still confuses me: sometimes you check for overflow, sometimes you don't.

We can check skipping for overflow for now because va() didn't as well. But `string` size is merely 256 bytes, whereas `va()` has an internal buffer of 1024 bytes.

I've added `MAX_VA_STRING` macro that tells current `va()` buffer size and took your `Cmd_Apropos_f` simplification as a possible example.

Also let's turn this PR into multiple smaller separate PRs, one that will change everything for client side, other for server side, third that will add format specifying helpers like `Info_SetValueForKeyf`, so it will be easier to review and test against regressions.

nekonomicon:
> there is one thing that still confuses me: sometimes you check for overflow, sometimes you don't.
I just think bounds check is useless in some cases.

PR #1175, opened 2022-12-19
engine: server: simplify strings operations.

nekonomicon:


engine/server/sv_save.c
```diff
@@ -2433,10 +2434,10 @@ int GAME_EXPORT SV_GetSaveComment( const char *savename, char *comment )
 
 		// split comment to sections
 		if( Q_strstr( savename, "quick" ))
-			Q_strncat( comment, "[quick]", CS_SIZE );
+			len = Q_strncpy( comment, "[quick]", CS_SIZE );
```
a1batross, line 2437: Again, is there any evidence that string length tracking is better than just calling strncat a few times in a function that's not performance critical or has string concatenation bottleneck?

Also, once you're here, `sizeof( comment )`

engine/server/sv_save.c
```diff
@@ -2433,10 +2433,9 @@ int GAME_EXPORT SV_GetSaveComment( const char *savename, char *comment )
 
 		// split comment to sections
 		if( Q_strstr( savename, "quick" ))
-			Q_strncat( comment, "[quick]", CS_SIZE );
+			Q_snprintf( comment, sizeof( comment ), "[quick]%s", description );
 		else if( Q_strstr( savename, "autosave" ))
-			Q_strncat( comment, "[autosave]", CS_SIZE );
-		Q_strncat( comment, description, CS_SIZE );
+			Q_snprintf( comment, sizeof( comment ), "[autosave]%s", description );
```
a1batross, line 2438: It seems incorrect. `description` should be in saved no matter if `savename` is not quick or autosave.

The correct strcat-less replacement would be: 
```
		if( Q_strstr( savename, "quick" ))
			Q_snprintf( comment, sizeof( comment ), "[quick]%s", description );
		else if( Q_strstr( savename, "autosave" ))
			Q_snprintf( comment, sizeof( comment ), "[autosave]%s", description );
		else Q_strncpy( comment, description, sizeof( comment ));
```

engine/server/sv_client.c
```diff
@@ -938,7 +941,15 @@ void SV_BuildNetAnswer( netadr_t from )
 			{
 				edict_t *ed = svs.clients[i].edict;
 				float time = host.realtime - svs.clients[i].connection_started;
-				Q_strncat( string, va( "%c\\%s\\%i\\%f\\", count, svs.clients[i].name, (int)ed->v.frags, time ), sizeof( string ));
+				ret = Q_snprintf( &string[len], sizeof( string ) - len, "%c\\%s\\%i\\%f\\", count, svs.clients[i].name, (int)ed->v.frags, time );
```
a1batross, line 944: `ret` scope could be limited in this block

a1batross (approved):

PR #1176, opened 2022-12-19
ref: gl: simplify strings operations.

nekonomicon:


a1batross:
Do we... win too much from this tracking string length spaghetti?

a1batross:
In other words, is it really simplification for _us_? Especially that it's not done each frame and only during world/texture loading and there is just no evidence in any profiler that `strncat` is our performance killer here?

a1batross (approved):

PR #1185, opened 2022-12-24
filesystem: fix lseek for mpg123

MoeMod:
Mpg123 requires a lseek-like api, which can be used as `lseek(fd, 0, SEEK_END)` to get filesize. However, `FS_Seek` just returns 0 on success. 
It seems like media/gamestartup.mp3 cannot be loaded due to this inconsistency.


a1batross:
Nice catch!

(haha seekex)

PR #1244, opened 2023-02-19
Add support for MP3 looping through custom ID3v2 tagging

a1batross:
Highly experimental

### DIFF
--- a/Documentation/extensions/mp3-loops.md
+++ b/Documentation/extensions/mp3-loops.md
@@ -1,0 +1,22 @@
+## Looping MP3 extension
+
+It is now possible to loop MP3 file in Xash3D FWGS by adding a custom text tag with `LOOP_START` or `LOOPSTART` in description and time point (in raw samples) in value.
+
+### Example with foobar2000
+1. Open Foobar2000
+2. Add your .mp3 file to playlist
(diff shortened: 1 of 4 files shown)


a1batross:
Forgot to support de-unsynching tag frames.

SNMetamorph:
@CASPERX69XSOFT tested it with ambient_generic, and it doesn't works. Link to looped file: https://disk.yandex.ru/d/5VjnGQ2VDUiVxA

a1batross:
So it turns out ID3v2.3 doesn't actually support UTF-8.

In engine, we explicitly drop UTF-16 tags. Gonna go back and enforce 2.4 only.

SNMetamorph:
So, what's next? We got it to work

a1batross:
Well, the comments in this PR is basically a TODO.

At this point I want to add:
* Very basic support for UTF-16 tags. Because tags we are using have only basic latin characters, I can do just memcmp(), and simple atoi for UTF-16.
* Unsynchornisation support, this is very important, as it can be automatically added by ID3 encoders, and may be not possible to disable.
* Add an example on how to work with mp3tag.

a1batross:
I will merge this as is, if somebody has time to add example for mp3tag, send a patch.

PR #1250, opened 2023-02-24
Fix multiline CL_CenterPrint

w23:
After recent refactoring it sets `clgame.centerPrint.lines` to 1 and doesn't update it anymore.

I'm not sure `cl_font.c` is supposed to be aware of `clgame`, so a better fix is probably to add an `int *lines` argument to `CL_DrawStringLen()`. Let me know if that's the case and I'll update the PR.


a1batross:
Fixed in another commit.

Issue #1205, opened 2023-01-08
Generating minidumps support


SNMetamorph:
Something like https://github.com/dreamstalker/rehlds/pull/703. For Windows it will be pretty straightforward.

a1batross:
How debuggable they are for Windows though?

a1batross:
On *nix we can just enable coredumps and "run" it under GDB. If it's compatible, which is not always possible for us here.

a1batross:
As we're talking about debugging already crashed program here, our crashhandler should print registers and stack as well. It would make it a little bit more usable.

SNMetamorph:
> How debuggable they are for Windows though?

It's good. We just can open minidump file in Visual Studio, and it will sumilate process debugging just like when you debugging running program itself.

SNMetamorph:
> As we're talking about debugging already crashed program here, our crashhandler should print registers and stack as well. It would make it a little bit more usable.

It's cool but it's not enough

a1batross:
I'm not sure if I will use them myself or even ask users to send me a core/minidump. Can be useful for mods though.

As for platform support, supporting just Windows can be enough for proof of concept.

SNMetamorph:
> I'm not sure if I will use them myself or even ask users to send me a core/minidump. Can be useful for mods though.
> 
> As for platform support, supporting just Windows can be enough for proof of concept.

Yes, it's must have for mods

PR #1257, opened 2023-03-08
Added support for minidumps generating

SNMetamorph:
Solves #1205

### DIFF
--- a/engine/common/crashhandler.c
+++ b/engine/common/crashhandler.c
@@ -34,6 +34,7 @@ Crash handler, called from system
 #include <winnt.h>
 #include <dbghelp.h>
 #include <psapi.h>
+#include <time.h>
 
 #ifndef XASH_SDL
 typedef ULONG_PTR DWORD_PTR, *PDWORD_PTR;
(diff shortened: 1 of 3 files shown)


engine/common/crashhandler.c
```diff
@@ -189,6 +190,62 @@ static void Sys_StackTrace( PEXCEPTION_POINTERS pInfo )
...
+
+static void Sys_GetMinidumpFileName( const char *processName, char *mdmpFileName, size_t bufferSize )
+{
+	time_t currentUtcTime = time( NULL );
+	struct tm *currentLocalTime = localtime( &currentUtcTime );
+
+	Q_snprintf( mdmpFileName, bufferSize, "%s_%s_crash_%d%.2d%.2d_%.2d%.2d%.2d.mdmp",
+		processName,
+		Q_buildcommit(),
+		currentLocalTime->tm_year + 1900,
```
a1batross, line 208: You can use strftime for formatting the date

a1batross: Nevermind, it's fine too.

a1batross:
Good job!

a1batross:
I will test it on my machine and merge. :)

a1batross:
It worked for me under Wine. Though MiniDumps here aren't so mini, got 400 MB file, that got compressed to 100 MB with Zstd. While it was generating, engine got freezed for few seconds, so that could be a problem. Would be nice to print them in Sys_Crash but that could be solved later.

Merging.

SNMetamorph:
> While it was generating, engine got freezed for few seconds, so that could be a problem. Would be nice to print them in Sys_Crash but that could be solved later.
> 
> Merging.

May be because of Wine? On my setup it's writes almost immediately. And about compression: LZMA does very good with minidimps.

PR #1298, opened 2023-03-18
Sound client-side precaching fix

SNMetamorph:


a1batross:
What does this fix exactly?

SNMetamorph:
> What does this fix exactly?

This is for timely precaching on client side. Otherwise, files are being downloaded to client, but not precached immediatly after it, and therefore causing a late precaching of sound (obvious, this is bad)

engine/server/sv_init.c
```diff
@@ -276,11 +276,20 @@ model_t *SV_ModelHandle( int modelindex )
 	return sv.models[modelindex];
 }
 
+static resourcetype_t SV_DetermineResourceType( const char *filename )
+{
+	if( !Q_strncmp( filename, "sound/", 6 ) && Sound_SupportedFileFormat( COM_FileExtension( filename )))
```
nekonomicon, line 281: You can use `DEFAULT_SOUNDPATH` macro here:
```
if( !Q_strncmp( filename, DEFAULT_SOUNPATH, sizeof( DEFAULT_SOUNPATH ) - 1 ) && Sound_SupportedFileFormat( COM_FileExtension( filename )))
```

macro:
https://github.com/FWGS/xash3d-fwgs/blob/master/engine/common/com_strings.h#L50=

engine/server/sv_init.c
```diff
@@ -295,8 +304,23 @@ void SV_ReadResourceList( const char *filename )
...
+		COM_FixSlashes( token );
+		restype = SV_DetermineResourceType( token );
+		Con_DPrintf( "  %s (%s)\n", token, COM_GetResourceTypeName( restype ));
+		switch( restype )
+		{
+			// TODO do we need to handle other resource types specifically too?
+			case t_sound:
+			{
+				const char *filepath = token;
+				filepath += 6; // skip "sound/" part
```
nekonomicon, line 316: And same here.

SNMetamorph:
More detailed explanation: all files listed in `mapname.res` are being added to resource list on server side. And therefore, when client connects to server, all these files will be downloaded for it. Because before this fix all files being processed as 'common' resource type, sounds/model are not being precached on client side after downloading, which causing late precache. In this fix I added condition for handling sound files as 'sound' resource type, and after that client will precache sound file after downloading it, as it should work in ideal.

But there is one opened question: should we handle model/sprite file as 'model' resource type? They also will precaching on client side immediately after downloading. But in other side, we never faced with models late precache problems yet.

SNMetamorph:
Another argument for handling model/sprite files as `t_model` resource type: there are separate resource lists for sounds/models/custom resources and they size is limited. When we handling all resources as `t_common`, "custom" resource list is being clogged, at same time sound/model resource lists are still almost empty.

a1batross:
I think having sound precache is enough for now. I don't see GoldSrc doing any of this, but having sound files precached early is a good practice.

PR #1305, opened 2023-03-23
PS Vita port patches

SNMetamorph:
In collaboration with @fgsfdsfgs

### DIFF
--- a/common/defaults.h
+++ b/common/defaults.h
@@ -191,6 +191,10 @@ Default build-depended cvar and constant values
 	#define DEFAULT_M_IGNORE "0"
 #endif // DEFAULT_M_IGNORE
 
+#ifndef DEFAULT_JOY_DEADZONE
+	#define DEFAULT_JOY_DEADZONE "4096"
+#endif // DEFAULT_JOY_DEADZONE
+
 #ifndef DEFAULT_DEV
 	#define DEFAULT_DEV 0
 #endif // DEFAULT_DEV
--- a/engine/common/launcher.c
+++ b/engine/common/launcher.c
@@ -91,9 +91,13 @@ _inline int Sys_Start( void )
 #if !XASH_WIN32
 int main( int argc, char **argv )
 {
+#if XASH_PSVITA
+	// inject -dev -console into args if required
+	szArgc = PSVita_GetArgv( argc, argv, &szArgv );
+#else
 	szArgc = argc;
 	szArgv = argv;
-
+#endif // XASH_PSVITA
 	return Sys_Start();
 }
 #else
--- a/engine/common/sys_con.c
+++ b/engine/common/sys_con.c
@@ -265,10 +265,18 @@ static void Sys_PrintStdout( const char *logtime, const char *msg )
 	IOS_Log( buf );
 #endif // TARGET_OS_IOS
 
-#if (XASH_NSWITCH && NSWITCH_DEBUG) || XASH_PSVITA // REMOVEME
+#if XASH_NSWITCH && NSWITCH_DEBUG
 	// just spew it to stderr normally in debug mode
 	fprintf( stderr, "%s %s", logtime, buf );
 #endif // XASH_NSWITCH && NSWITCH_DEBUG
+
+#if XASH_PSVITA
+	// spew to stderr only in developer mode
+	if( host_developer.value )
+	{
+		fprintf( stderr, "%s %s", logtime, buf );
+	}
+#endif
 
 #elif !XASH_WIN32 // Wcon does the job
 	Sys_PrintLogfile( STDOUT_FILENO, logtime, msg, XASH_COLORIZE_CONSOLE );
--- a/engine/platform/psvita/sys_psvita.c
+++ b/engine/platform/psvita/sys_psvita.c
@@ -156,7 +156,7 @@ void PSVita_Init( void )
 
 	if( vrtld_init( 0 ) < 0 )
 	{
-		Sys_Error( "Could not init vrtld: %s\n", vrtld_dlerror( ) );
+		Sys_Error( "Could not init vrtld:\n%s\n", vrtld_dlerror() );
 	}
 
 	// init vitaGL, leaving some memory for DLL mapping
--- a/engine/platform/psvita/sys_psvita.c
+++ b/engine/platform/psvita/sys_psvita.c
@@ -90,8 +90,7 @@ void PSVita_Init( void )
 		chdir( xashdir );
 	}
 
-	sceCtrlSetSamplingModeExt( SCE_CTRL_MODE_ANALOG_WIDE );
-	sceTouchSetSamplingState( SCE_TOUCH_PORT_FRONT, SCE_TOUCH_SAMPLING_STATE_START );
+	sceTouchSetSamplingState( SCE_TOUCH_PORT_BACK, SCE_TOUCH_SAMPLING_STATE_STOP );
 	scePowerSetArmClockFrequency( 444 );
 	scePowerSetBusClockFrequency( 222 );
 	scePowerSetGpuClockFrequency( 222 );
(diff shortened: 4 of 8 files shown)


common/defaults.h
```diff
@@ -191,6 +192,10 @@ Default build-depended cvar and constant values
 	#define DEFAULT_M_IGNORE "0"
 #endif // DEFAULT_M_IGNORE
 
+#ifndef DEFAULT_JOY_DEADZONE
+	#define DEFAULT_JOY_DEADZONE "0"
```
a1batross, line 196: I think default can be 4096, so we have somewhat sane values for all platforms.

(and somewhere in the future get an UI for this)

engine/common/sys_con.c
```diff
@@ -123,8 +123,14 @@ static void Sys_FlushLogfile( void )
 void Sys_InitLog( void )
 {
 	const char	*mode;
+#if XASH_PSVITA
+	// force-enable logging when developer is set to 2+
+	const qboolean force_log = ( host_developer.value > DEV_NORMAL );
```
a1batross, line 128: Why should we do this? If you somehow enabled developer mode (using `-dev` flag I suppose, not switching cvar), I think you can enable `-log` as well.

engine/platform/psvita/sys_psvita.c
```diff
@@ -75,32 +75,63 @@ const size_t __vrtld_num_exports = sizeof( aux_exports ) / sizeof( *aux_exports
...
-	sceCtrlSetSamplingModeExt( SCE_CTRL_MODE_ANALOG_WIDE );
-	sceTouchSetSamplingState( SCE_TOUCH_PORT_FRONT, SCE_TOUCH_SAMPLING_STATE_START );
+	sceAppUtilInit( &initParam, &bootParam );
+
+	// check if the user wants debug mode
+	if( PSVita_GetLaunchParameter( param ))
+	{
+		if( !Q_strcmp( param, "dev" ))
+		{
+			Con_Reportf( S_NOTE "Enabled developer mode from launcher.\n" );
```
a1batross, line 117: What if we append cmdline instead, append `-dev 2 -log` there and let the engine automatically enable developer mode with enabled logging without platform-specific behavior in `engine/common`?

engine/platform/psvita/sys_psvita.c
```diff
@@ -75,32 +75,63 @@ const size_t __vrtld_num_exports = sizeof( aux_exports ) / sizeof( *aux_exports
...
 	scePowerSetBusClockFrequency( 222 );
 	scePowerSetGpuClockFrequency( 222 );
 	scePowerSetGpuXbarClockFrequency( 166 );
 	sceSysmoduleLoadModule( SCE_SYSMODULE_NET );
 
-	if ( vrtld_init( 0 ) < 0 )
+	if( vrtld_init( 0 ) < 0 )
 	{
-		Sys_Error( "Could not init vrtld: %s\n", vrtld_dlerror( ) );
+		// Host_Error might not show a message box here since framecount is 0
```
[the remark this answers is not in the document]
a1batross: Yes, and you replaced Sys_Error with Sys_Error, which does in fact spawn a messagebox.

engine/platform/psvita/sys_psvita.c
a1batross: Well, `-dev` checked before `Platform_Init` is called.

What if you reimplmement `main()` from `common/launcher.c` to make it PSVita-specific that reads livearea parameters and sets the appropriate argc and argv?

a1batross: You can take out main() function from `common/launcher.c` (this file REALLY needs to be decoupled to platform-specific launchers, or removed completely, but we'll do that later =/), remove anything that's not relevant to `psvita` and insert your livearea check there.

common/defaults.h
SNMetamorph: Ok, fixed

engine/platform/psvita/sys_psvita.c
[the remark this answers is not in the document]
SNMetamorph: On PS Vita calling `Sys_Error` for unknown reason causes crash without any messages

engine/platform/psvita/sys_psvita.c
fgsfdsfgs: In relation to the previous review and this one, like you said, you can't trivially inject arguments in `Platform_Init` and I wanted to keep all Vita-related initialization in there. But if that's not a requirement then sure.

a1batross: Yeah, but we really need to override `main()` for platforms. I think it's a good compromise between hacking the engine functions, and have two stage initialization for platforms.

engine/platform/psvita/sys_psvita.c
```diff
@@ -75,32 +76,76 @@ const size_t __vrtld_num_exports = sizeof( aux_exports ) / sizeof( *aux_exports
...
+===========
+PSVita_GetArgv
+
+On the PS Vita under normal circumstances argv is empty, so we'll construct our own
+based on which button the user pressed in the LiveArea launcher.
+===========
+*/
+int PSVita_GetArgv( char ***out_argv )
+{
+	static const char *fake_argv[MAX_ARGV] = { "", NULL };
```
a1batross, line 107: argv[0] shouldn't be an empty string. You can write anything you want there, just as a fake executable name.

engine/common/launcher.c
```diff
@@ -91,9 +91,13 @@ _inline int Sys_Start( void )
 #if !XASH_WIN32
 int main( int argc, char **argv )
 {
+#if XASH_PSVITA
+	// argv is normally empty, so we generate our own
```
a1batross, line 95: You could just copypaste the whole main and Sys_Start, and disable compiling this file whatsoever. But ok, it will be split later then.

engine/common/sys_con.c
```diff
@@ -265,11 +265,19 @@ static void Sys_PrintStdout( const char *logtime, const char *msg )
...
 #endif // TARGET_OS_IOS
 
-#if (XASH_NSWITCH && NSWITCH_DEBUG) || XASH_PSVITA // REMOVEME
+#if XASH_NSWITCH && NSWITCH_DEBUG
 	// just spew it to stderr normally in debug mode
 	fprintf( stderr, "%s %s", logtime, buf );
 #endif // XASH_NSWITCH && NSWITCH_DEBUG
 
+#if XASH_PSVITA
+	// spew to stderr only in developer mode
```
a1batross, line 274: Why though? Does writing to stderr even matters on this platform?

Then it could be shared with NSwitch code as well! It doesn't look PSVita-specific.

a1batross: Either make it always print, either let NSwitch print to stderr with enabled host_developer.

But we shouldn't depend on host_developer here, all the optional console messages are rejected even before this function gets called.

fgsfdsfgs: > Why though? Does writing to stderr even matters on this platform?
> 
> Then it could be shared with NSwitch code as well! It doesn't look PSVita-specific.

Used to be slow, so I didn't want to make it always do that. Not sure about now. This can probably be restored to the old state then.

engine/platform/psvita/sys_psvita.c
fgsfdsfgs: It is an empty string in the actual `argv` when the game is started normally, so I wanted it to act the same in both cases.

engine/common/sys_con.c
a1batross: 🤔
Again, we shouldn't print too much if developer mode is turned off and if console was disabled... Which isn't true, `psvita` has DEFAULT_ALLOW_CONSOLE set to 1.

Maybe leave this code as is and do not turn it into a platform-specific mess, but disable DEFAULT_ALLOW_CONSOLE?

a1batross:
Well, let's merge this anyway.

I'm still very unhappy about syslog.

PR #1324, opened 2023-04-03
Add GPU timestamps and scopes to profiler

w23:
Things left:
- [ ] staging integration
  - can we combine staging cmdbuf management with framectl? does it need its own cmdbuf set? for what? texture uploading? building two sets of commands in parallel that need to run sequentially?
- [ ] BLAS build integration (requires staging)
- [ ] try "nesting" combuf scopes explicitly, e.g. all ray tracing under "yay tracing"


w23:
https://github.com/orgs/community/discussions/11729 :disappointed:

PR #1346, opened 2023-05-02
New vgui support api

a1batross:
- Drafts new vgui_support API and implements it.
- Adds preliminary VGUI2 support by @kungfulon


a1batross:
I need to stub chromehtml API, so we don't have this heavy dependency.

a1batross:
@mittorn mods should never load chromehtml.so, and in practice never do.

It's much easier to add chromehtml interface to vgui_support and return it in `CreateInterface`.

PR #1360, opened 2023-05-17
Implement simple speedhack detection and automatic kicking

a1batross:
Based on already existing logic of `clockwindow`.


tyabus (commented):
Would be great if this was a cvar.

a1batross:
Was merged.

Issue #1188, opened 2022-12-26
Default GL_FOG_MODE in GoldSource


FreeSlave:
When rendering fog, the modern GoldSource sets `GL_FOG_MODE` to `GL_EXP2`.

Reverse according to @a1batross 
```cpp
void R_RenderFinalFog( void )
{
qglEnable( GL_FOG );
qglFogi( GL_FOG_MODE, GL_EXP2 );
qglFogf( GL_FOG_DENSITY, flFogDensity );
qglHint( GL_FOG_HINT, GL_NICEST );
qglFogfv( GL_FOG_COLOR, flFinalFogColor );
qglFogf( GL_FOG_START, flFogStart );
qglFogf( GL_FOG_END, flFogEnd );
}
```

It probably was changed at some point, and I guess `GL_LINEAR` was used before. Currently [Xash3D uses GL_LINEAR](https://github.com/FWGS/xash3d-fwgs/blob/12bb0ca44b966b16c7ec3688ae07404b485ae455/ref/gl/gl_triapi.c#L290) too.

With `GL_EXP2` however the call to `gEngfuncs.pTriAPI->Fog` from a client library doesn't make sense on its own, because exponential fog doesn't use `GL_FOG_START` and `GL_FOG_END`. It requires `GL_FOG_DENSITY` which is set only via `gEngfuncs.pTriAPI->FogParams`. If density is set to 0 or not set at all (i.e. `FogParams` was not called), GoldSource doesn't render fog (again, because it's exponential). Xash3D however does render fog and doesn't need a call to `FogParams`, because it uses the linear fog.

I don't know if anything should be changed in the engine, I'm opening the issue just to increase the awareness of differencies between Xash3D and GoldSource, when using triapi fog from client library.
As a compromise Xash3D can use exponential fog if density is not zero, and continue using linear fog if density is zero. Note, however, that `GL_FOG_MODE` can be changed from a client library via direct call to `glFogi` (assuming the client dynamically loaded opengl functions) after or between calls to `pTriAPI->Fog` and `pTriAPI->FogParams`, so it's not clear when the engine should decide on the `GL_FOG_MODE`.

Velaron:
Видимо наткнулся на это в CS1.6, в стиме туман совсем другой:
![image](https://github.com/FWGS/xash3d-fwgs/assets/53096779/d2e99348-ad9d-4960-b3ff-40073ee92eb7)
[...]

PR #1365, opened 2023-05-20
ref: gl: Goldsrc compatible fog implementation

Velaron:
Fix https://github.com/FWGS/xash3d-fwgs/issues/1188


a1batross (approved):

FreeSlave:
Seems like assumes that pTriAPI->FogParams must be called before calling pTriAPI->Fog.
It might be a problem for mods where FogParams is called after the Fog.

PR #1367, opened 2023-05-20
Cvars allocation

a1batross:
This is an important branch, as it moves almost every convar from dynamic allocation to static, making it easier for the CPU to address such cvars, avoiding one pointer dereferencing, and, in case of CVAR_TO_BOOL macro, NULL pointer check.

In simple `timedemo` based benchmark this increases performance from 2700 FPS to 3100 FPS.

I would like to ask reviewers to carefully check these patches, because this was only half-automated and therefore there would be bugs.


a1batross:
I looked into this a few times but nobody really wanted to review this except me.

So... YOLO

PR #1381, opened 2023-06-09
Implemented correct answers to TSourceEngineQuery server queries

jeefo:
Corrected answer to A2S_INFO, and implemented A2S_RULES, A2S_PLAYER,   ~~A2A_PING~~ 

It's also fixes alot of problems with third-party server query stuff like SteamCondenser, Hlstats, HLSW etc.

was:
image

became:
image

### DIFF
--- a/engine/server/sv_client.c
+++ b/engine/server/sv_client.c
@@ -3063,64 +3063,6 @@ void SV_ExecuteClientCommand( sv_client_t *cl, const char *s )
 }
 
 /*
-==================
-SV_TSourceEngineQuery
-==================
-*/
-void SV_TSourceEngineQuery( netadr_t from )
-{
-	// A2S_INFO
-	char	answer[1024] = "";
-	int	count, bots;
-	int	index;
-	sizebuf_t	buf;
-
-	SV_GetPlayerCount( &count, &bots );
-
-	MSG_Init( &buf, "TSourceEngineQuery", answer, sizeof( answer ));
-
-	MSG_WriteLong( &buf, -1 ); // Mark as connectionless
-	MSG_WriteByte( &buf, 'm' );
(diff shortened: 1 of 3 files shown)


engine/server/sv_client.c
```diff
@@ -3150,7 +3092,8 @@ void SV_ConnectionlessPacket( netadr_t from, sizebuf_t *msg )
 	pcmd = Cmd_Argv( 0 );
 	Con_Reportf( "SV_ConnectionlessPacket: %s : %s\n", NET_AdrToString( from ), pcmd );
 
-	if( !Q_strcmp( pcmd, "ping" )) SV_Ping( from );
+	if (SV_SourceQuery_HandleConnnectionlessPacket ( pcmd, from ) ) return;
```
a1batross, line 3095: IMHO it should first check our packets, and then check Source ones.

Source packets have single letter header, and it clashes with our packets. Like `i` vs `info`.

jeefo: Agree about ``info`` but...

goldsrc-style A2A_PING really used anywhere now inside xash ? Probably it should be dropped in favor of actual goldsrc (source-style) query protocol ?

a1batross: These packets are still the part of the protocol.

A2A_PING would be nice to use in the next ServerBrowser iteration with accurate ping times, detailed server info and other, but then I would use Xash's own protocol anyway.

jeefo: okay, returned back old-style A2A_PING, source-query now doesn't handle these type of packet.

a1batross:
It's fine by me. I will squash merge it, and fix the styling issues myself.

a1batross:
Merged now.

PR #1414, opened 2023-08-29
Changed register names for NetBSD x86

Emru1:
Probably also needs change for x86_64


a1batross (approved):
Looks good to me! Thanks for the fix!

Emru1:
I think I'll try to provide some more NetBSD fixes in the nearest future

PR #1418, opened 2023-08-31
[WIP] GoldSrc protocol support

a1batross:
Because Xash can't connect to Steam to get certificate cookie, it requires `sv_lan 1` on HLDS side.

Almost nothing does work except joining to the server. Use `connect IP:Port goldsrc` to connect to GoldSrc server.


a1batross:
Had a good online match using Xash3D FWGS on ReHLDS-based server with GoldSrc clients. Friendship! And not a single crash!

Also recorded two demos to test protocol support on.

nekonomicon:
May be use fresh libbzip2 versions from gitlab?
https://gitlab.com/bzip2/bzip2/-/tags

mittorn:
What about ktx.bz2?

a1batross:
> May be use fresh libbzip2 versions from gitlab? https://gitlab.com/bzip2/bzip2/-/tags

Not a big issue, since the protocol impl is almost ready now.

At start, I just threw whatever was in ReHLDS to be sure it's exactly the same bz2 that they are using.

a1batross:
We can as well just link with system provided libbz2. :)

nekonomicon:
Yep, you're right.
And I didn't see any critical fixes or additions in new releases.
One small thing which annoying me - there yet another crc32 table.

a1batross:
> Yep, you're right. And I didn't see any critical fixes or additions in new releases. One small thing which annoying me - there yet another crc32 table.

I don't think we can do anything if the dependency carries its own CRC32 implementation.

SNMetamorph:
What is planned for PR completion? Maybe we can rebase and merge all this already? So people can easily test it

a1batross:
Let's rebase/merge this with current master at first... and then decide what to do.

When I last tried this branch, I disliked these things about it:
1. We had to connect to ReHLDS servers. Can't connect to Steam servers. I thought of an implementing some kind of broker that would help connect to Steam servers without explicitly linking Xash3D to Steam API DLLs. (which is absolute no go by itself)
Don't want to implement connection to pirated servers, either. This part of HL multiplayer history better be forgotten, IMHO.
2. I don't like how GoldSrc messaging quirks get into the common code. We only implement client connection here, so this better not touch common code if possible.

a1batross:
Some thoughs about 1.

* Broker can be a simple program that Xash starts on its own, which then will tell Steam that we run Half-Life and need a certificate blob and communicate with it through some kind of IPC. But this also limits connection to x86 Windows/Linux-only ports, since Steam doesn't exist on other platforms.
* The VAC. Steam doesn't send VAC blobs to GoldSrc clients anymore as far as I know. But technically speaking, VAC still exists.

SNMetamorph:
> 1. We had to connect to ReHLDS servers. Can't connect to Steam servers.

This is the thing that better to be reworked in future, but I think for now it's ok. It's more important to have ability to connect to GoldSrc servers from master branch, so it will could be done without playing with different branches and keeping it up-to-date.

mittorn:
It it possible to run broker on different machine? Basicly, it might be useful for network-connected platforms

a1batross:
Well, broker is something that only exists in my mind. I haven't implemented it yet.

At first stage, we better merge this branch. @SNMetamorph is right in that regard, as it's actually a working thing, and it wouldn't be fair to hide it here forever.

a1batross:
Reworked and merged.

PR #1423, opened 2023-09-12
Fixed macOS (Darwin) compilation

sofakng:
This fixes the compilation issues on macOS (Darwin).

1. Removed strip_on_install from Waf script (wscript) for Darwin hosts.
2. Enables DISABLE_WERROR by default for Darwin hosts.

### DIFF
--- a/wscript
+++ b/wscript
@@ -180,7 +180,11 @@ def configure(conf):
 	if conf.env.COMPILER_CC == 'msvc':
 		conf.load('msvc_pdb')
 
-	conf.load('msvs msdev subproject gitversion clang_compilation_database strip_on_install waf_unit_test enforce_pic cmake')
+	conf.load('msvs msdev subproject gitversion clang_compilation_database waf_unit_test enforce_pic cmake')
+
+	# macOS (Darwin) doesn't support objcopy so we need to disable binary stripping
+	if conf.env.DEST_OS != 'darwin':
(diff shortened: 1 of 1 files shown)


a1batross:
1. Thanks, made it optional in `strip_to_install` tool: https://github.com/FWGS/waifu/commit/7965cf1475ebb9153ae0974623ceaf5afce5a50d
2. It's better to go, and fix possible errors. Can you run `./waf clean build` without this patch applied?

sofakng:
Thanks so much for making these changes!

Here is the first error:
```
[207/405] Compiling engine/common/net_ws.c
../engine/common/net_ws.c:1069:10: error: implicit conversion from enumeration type 'qboolean' to different enumeration type 'net_gai_state_t' [-Werror,-Wenum-conversion]
```

Here is the compiler version:
```
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: arm64-apple-darwin23.0.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

```

a1batross:
You probably can run build with `-k` flag, that way Waf won't stop compiling at first error.

I made the changes, is it fine now? Interestingly enough, the -Wenum-conversion doesn't trigger under vanilla clang-15.

sofakng:
It configures and compiles perfectly now.  Thank you so much for the changes!

PR #1427, opened 2023-09-17
Update port maintainer list

sofakng:


a1batross:
I think you can move it higher in the list, so it's easily visible...

Actually, will merge it anyway. This file feels cluttered with stuff, better re-sort it.

And welcome to FWGS, I guess :)

sofakng:
Thanks and thanks! :)

PR #1446, opened 2023-10-13
gl2shim for core/gles2+ contexts

mittorn:
Limited FFP emulation with streaming buffers for gles/core contexts
Please check on nvidia/amd/intel proprietary drivers before merging, maybe need change some feature detection/blacklisting logic


engine/platform/sdl/vid_sdl.c
```diff
@@ -812,8 +812,13 @@ qboolean VID_CreateWindow( int width, int height, window_mode_t window_mode )
 	{
 		if( !glw_state.initialized )
 		{
-			if( !GL_CreateContext( ))
-				return false;
+			while( !GL_CreateContext( ))
```
a1batross, line 815: I don't get it. There is already function `VID_CreateWindowWithSafeGL` that iterates over safegl options.

Why there is a second loop?

ref/gl/gl_image.c
```diff
@@ -1070,7 +1070,7 @@ static void GL_TextureImageRAW( gl_texture_t *tex, GLint side, GLint level, GLin
 	}
 	else if( tex->target == GL_TEXTURE_2D_MULTISAMPLE )
 	{
-#if !defined( XASH_GLES ) && !defined( XASH_GL4ES )
+#if !defined( XASH_GL_STATIC ) || (!defined( XASH_GLES ) && !defined( XASH_GL4ES ))
```
a1batross, line 1073: What's the point for checking for XASH_GL_STATIC here? It doesn't seem to change anything for MSAA.

a1batross: If you're updating macros, don't forget to update comments after #else and #endif

ref/gl/gl_local.h
```diff
@@ -40,6 +40,9 @@ int VGL_ShimInit( void );
 void VGL_ShimShutdown( void );
 void VGL_ShimEndFrame( void );
 #endif
+#if !defined(XASH_GL_STATIC)
+#include "gl2_shim/gl2_shim.h"
```
a1batross, line 44: I don't think #if !defined(XASH_GL_STATIC) really needed here. Including gl2shim header won't break anything.

ref/gl/gl_opengl.c
```diff
@@ -247,6 +251,177 @@ static dllfunc_t drawrangeelementsextfuncs[] =
...
+//	{ GL_CALL( glVertexAttrib3fv ) },
+//	{ GL_CALL( glVertexAttrib4f ) },
+//	{ GL_CALL( glVertexAttrib4fv ) },
+//	{ GL_CALL( glVertexAttrib4ubv ) },
+{ NULL, NULL }
+};
+
+static dllfunc_t shaderobjectsfuncs_gles[] =
+{
+	{ "glDeleteShader"             , (void **)&pglDeleteObjectARB },
```
a1batross, line 341: Let's use GL_CALL macro here, it's very easy to pass wrong pointer here. You can create a second macro that will include ARB postfix.

ref/gl/gl_opengl.c
```diff
@@ -247,6 +251,177 @@ static dllfunc_t drawrangeelementsextfuncs[] =
...
+	{ "glGetProgramInfoLog"               , (void **)&pglGetProgramInfoLog },
+	//{ "glVertexAttrib4f"              , (void **)&pglVertexAttrib4fARB },
+	//{ "glVertexAttrib4fv"             , (void **)&pglVertexAttrib4fvARB },
+	//{ "glVertexAttrib4ubv"            , (void **)&pglVertexAttrib4ubvARB },
+{ NULL, NULL }
+};
+
+static dllfunc_t vaofuncs[] =
+{
+{ "glBindVertexArray"    , (void **)&pglBindVertexArray },
```
a1batross, line 400: Use GL_CALL here.

ref/gl/gl_opengl.c
```diff
@@ -362,8 +538,20 @@ qboolean GL_CheckExtension( const char *name, const dllfunc_t *funcs, const char
 		// functions are cleared before all the extensions are evaluated
 		if((*func->func = (void *)gEngfuncs.GL_GetProcAddress( func->name )) == NULL )
 		{
-			// one or more functions are invalid, extension will be disabled
-			GL_SetExtension( r_ext, false );
+			// HACK: fix ARB names
+			char *str = Q_strstr(func->name, "ARB");
```
a1batross, line 542: Add spaces in parentheses.

ref/gl/gl_opengl.c
```diff
@@ -556,12 +748,21 @@ void GL_InitExtensionsGLES( void )
...
 		{
 		case GL_ARB_VERTEX_BUFFER_OBJECT_EXT:
-			GL_SetExtension( extid, true );
+			GL_CheckExtension( "vertex_buffer_object", vbofuncs, "gl_vertex_buffer_object", extid, 1.0 );
 			break;
 		case GL_ARB_MULTITEXTURE:
-			GL_SetExtension( extid, true ); // required to be supported by wrapper
+			if(!GL_CheckExtension( "multitexture", multitexturefuncs, "gl_arb_multitexture", GL_ARB_MULTITEXTURE, 1.0) && glConfig.wrapper == GLES_WRAPPER_NONE  )
+#ifndef XASH_GL_STATIC
+				if( !GL_CheckExtension( "multitexture_es1", multitexturefuncs_es, "gl_arb_multitexture", GL_ARB_MULTITEXTURE, 1.0 ) )
```
a1batross, line 756: Can you just use && here instead of nested ifs?

ref/gl/gl_opengl.c
```diff
@@ -571,24 +772,48 @@ void GL_InitExtensionsGLES( void )
...
-			GL_CheckExtension( "GL_OES_texture_npot", NULL, "gl_texture_npot", extid );
+			GL_CheckExtension( "GL_OES_texture_npot", NULL, "gl_texture_npot", extid, 0 );
+			break;
+#ifndef XASH_GL_STATIC
+		case GL_SHADER_OBJECTS_EXT:
+			GL_CheckExtension( "ES2 Shaders", shaderobjectsfuncs_gles, "gl_shaderobjects", extid, 2.0 );
+			break;
+		case GL_ARB_VERTEX_ARRAY_OBJECT_EXT:
+			if(!GL_CheckExtension( "GL_OES_vertex_array_object", vaofuncs, "gl_vertex_array_object", extid, 3.0 ))
+				!GL_CheckExtension( "GL_EXT_vertex_array_object", vaofuncs, "gl_vertex_array_object", extid, 3.0 );
```
a1batross, line 796: Unused NOT operator here.

ref/gl/gl_opengl.c
```diff
@@ -739,31 +977,83 @@ void GL_InitExtensionsBigGL( void )
...
-	GL_CheckExtension( "OpenGL 1.1.0", opengl_110funcs, NULL, GL_OPENGL_110 );
+	GL_CheckExtension( "OpenGL 1.1.0", opengl_110funcs, NULL, GL_OPENGL_110, 1.0 );
 
 	// get our various GL strings
 	glConfig.vendor_string = (const char *)pglGetString( GL_VENDOR );
 	glConfig.renderer_string = (const char *)pglGetString( GL_RENDERER );
 	glConfig.version_string = (const char *)pglGetString( GL_VERSION );
 	glConfig.extensions_string = (const char *)pglGetString( GL_EXTENSIONS );
+
+	pglGetIntegerv(GL_MAJOR_VERSION, &major);
```
a1batross, line 1011: Spaces in parentheses please.

ref/gl/gl_opengl.c
```diff
@@ -739,31 +977,83 @@ void GL_InitExtensionsBigGL( void )
...
+	pglGetIntegerv(GL_MINOR_VERSION, &minor);
+	if( !major && glConfig.version_string )
```
a1batross, line 1013: Can this even happen? O_o?

ref/gl/gl2_shim/gl2_shim.h
```diff
@@ -0,0 +1,24 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+*/
+
+#pragma once
+
+// max verts in a single frame
+#define GL2_MAX_VERTS 32768
```
a1batross, line 19: It's pretty low for our frame. 

net_graph for example, easily exceeds this

ref/gl/gl_opengl.c
lewa-j: Yes. GL_MAJOR_VERSION was intoduced in gl 3.0

mittorn:
Because it is not called when GL_CreateContext failed. It is only called when it failed create window, but does not work in case window created, but context not. It just runs engine without valid context and it crashes on bad gl function calls.

14 октября 2023 г. 5:46:20 GMT+03:00, Alibek Omarov ***@***.***> пишет:

mittorn:
I do not want to have declaration of missing functtion present. It wont be present if gl2shim part not compiled at all (it have macro in source file), so why include this header?

14 октября 2023 г. 7:42:07 GMT+03:00, Alibek Omarov ***@***.***> пишет:

mittorn:
gl changed function names, so we need macro including two names in this case, ARB does not call all cases

14 октября 2023 г. 7:45:03 GMT+03:00, Alibek Omarov ***@***.***> пишет:

mittorn:
&& inside macro? I not sure if it would be better

14 октября 2023 г. 7:46:57 GMT+03:00, Alibek Omarov ***@***.***> пишет:

mittorn:
It seems to be some bug inside net_graph, it should not draw to many quads, but i'll add workaround for big GL_QUADS cases.

14 октября 2023 г. 8:13:09 GMT+03:00, Alibek Omarov ***@***.***> пишет:

a1batross:
> Because it is not called when GL_CreateContext failed. It is only called when it failed create window, but does not work in case window created, but context not. It just runs engine without valid context and it crashes on bad gl function calls.

It's better to be moved out of the scope of this patchset then.

>I do not want to have declaration of missing functtion present. It wont be present if gl2shim part not compiled at all (it have macro in source file), so why include this header?

It will be detected at the link time. I don't see a problem here.

> gl changed function names

Yeah, we discussed this already. It's very unfortunate, so let's keep it as is then.

> && inside macro? I not sure if it would be better

I think this code is not an example of a good code. It's not very obvious at which level `break` statement goes. The absence of code blocks even make it so much worse, that you even made a mistake in the checks. Try to take both macro paths, the one with defined XASH_GL_STATIC and the one where it's not defined.

mittorn:
This might be availiable in gles context without static gl. I just rechecked all XASH_GLES leftovers because many of them refered to XASH_GL_STATIC really, not to GLES context itself. But GLES without wrappers was not supported before. This commit first time add launching on pure gles not completely wrapped by nanogl/wes/gl4es, only having ffp implemented with gl2shim

14 октября 2023 г. 7:41:22 GMT+03:00, Alibek Omarov ***@***.***> пишет:

mittorn:
Alibek Omarov писал(а) 2023-10-15 02:41:

>> Because it is not called when GL_CreateContext failed. It is only 
>> called when it failed create window, but does not work in case window 
>> created, but context not. It just runs engine without valid context 
>> and it crashes on bad gl function calls.
> 
> It's better to be moved out of the scope of this patchset then.
> 
> Message ID: ***@***.***>

I need to test if i did not break older GL versions and found that i 
cannot even create context before 3.x with compatibility flag enabled 
and safegl was not set automaticly. I searched other places where it 
might be set, but it was unrelated: this function only calls 
SDL_CreateWindow. Even when this safegl level enough to create window, 
it might not enough to create context, so we need to increment it later
--=_f98e57ced455552daa412d819c95ca55
Content-Transfer-Encoding: quoted-printable
Content-Type: text/html; charset=UTF-8

<html><head><meta http-equiv=3D"Content-Type" content=3D"text/html; charset=
=3DUTF-8" /></head><body style=3D'font-size: 10pt; font-family: Verdana,Gen=
eva,sans-serif'>
<p id=3D"reply-intro">Alibek Omarov =D0=BF=D0=B8=D1=81=D0=B0=D0=BB(=D0=B0) =
2023-10-15 02:41:</p>
<blockquote type=3D"cite" style=3D"padding: 0 0.4em; border-left: #1010ff 2=
px solid; margin: 0">
<div id=3D"replybody1">
<p><br /></p>
<blockquote type=3D"cite" style=3D"padding: 0 0.4em; border-left: #1010ff 2=
px solid; margin: 0">
<p dir=3D"auto">Because it is not called when GL_CreateContext failed. It i=
s only called when it failed create window, but does not work in case windo=
w created, but context not. It just runs engine without valid context and i=
t crashes on bad gl function calls.</p>
</blockquote>
<p dir=3D"auto">It's better to be moved out of the scope of this patchset t=
hen.</p>
<p style=3D"font-size: small; -webkit-text-size-adjust: none; color: #666;"=
><span style=3D"color: transparent; font-size: 0; display: none; visibility=
: hidden; overflow: hidden; opacity: 0; width: 0px; height: 0px; max-width:=
 0; max-height: 0; mso-hide: all;">Message ID: <span>&lt;FWGS/xash3d-fwgs/p=
ull/1446/c1763221270</span><span>@</span><span>github</span><span>.</span><=
span>com&gt;</span></span></p>
</div>
</blockquote>
<p>I need to test if i did not break older GL versions and found that i can=
not even create context before 3.x with compatibility flag enabled and safe=
gl was not set automaticly. I searched other places where it might be set, =
but it was unrelated: this function only calls SDL_CreateWindow. Even when =
this safegl level enough to create window, it might not enough to create co=
ntext, so we need to increment it later</p>

</body></html>

--=_f98e57ced455552daa412d819c95ca55--

--=_4bdf773d7ffda6467a4f2456ec038f46
Content-Transfer-Encoding: base64
Content-Type: image/gif;
 name=blocked.gif
Content-Disposition: attachment;
 filename=blocked.gif;
 size=118

R0lGODlhZAAyAIAAAPrOzgAAACH5BAAAAAAALAAAAABkADIAAAJNhI+py+0Po5y02ouz3rz7D4bi
SJbmiabqyrbuC8fyTNf2jef6zvf+DwwKh8Si8YhMKpfMpvMJjUqn1Kr1is1qt9yu9wsOi8fksvls
KwAAOw==
--=_4bdf773d7ffda6467a4f2456ec038f46--

a1batross:
GitHub didn't close it automatically AGAIN.

Closing manually, merged.

PR #1455, opened 2023-10-17
engine: common: imagelib: add KTX2 support

w23:
Adds basic KTX2 support for a few compressed formats, as part of a still ongoing ref_vk effort. KTX2 essentially is a Vulkan-centric texture format that supports literally hundreds of pixel formats.
For now only support for these is added:
- `VK_FORMAT_BC4_UNORM_BLOCK`
- `VK_FORMAT_BC4_SNORM_BLOCK`
- `VK_FORMAT_BC5_UNORM_BLOCK`
- `VK_FORMAT_BC5_SNORM_BLOCK`
- `VK_FORMAT_BC6H_UFLOAT_BLOCK`
- `VK_FORMAT_BC6H_SFLOAT_BLOCK`
- `VK_FORMAT_BC7_UNORM_BLOCK`
- `VK_FORMAT_BC7_SRGB_BLOCK`

Adding more formats is relatively straightforward:
- Copy format definition from `VkFormat` enum in `vulkan_core.h`
- Add a new definition into `pixformat_t` enum.
- Add format size calculation into `Image_ComputeSize()`

While we're at it, also adds a few new formats to DDS:
- `PF_BC4_UNSIGNED` for BC4_UNORM
- `PF_BC4_SIGNED` for BC4_SNORM
- `PF_BC5_UNSIGNED` for BC5_UNORM
- `PF_BC5_SIGNED` for BC5_SNORM
- BC7 is expanded into `PF_BC7_UNORM` and `PF_BC7_SRGB`

ref_gl and ref_soft code is updated for updated definitions where it made sense, but they were not tested beyond compilation success. 
These new formats and KTX2 files were tested with the ref_vk fork.


engine/common/imagelib/img_dds.c
```diff
@@ -356,7 +362,9 @@ qboolean Image_LoadDDS( const char *name, const byte *buffer, fs_offset_t filesi
 			SetBits( image.flags, IMAGE_HAS_ALPHA );
 		else if( image.type == PF_DXT5 && Image_CheckDXT5Alpha( &header, fin ))
 			SetBits( image.flags, IMAGE_HAS_ALPHA );
-		else if ( image.type == PF_BC7 )
+		else if ( image.type == PF_BC5_SIGNED || image.type == PF_BC5_UNSIGNED )
```
a1batross, line 365: No space before parentheses, only inside.

public/ktx2.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
```
a1batross, line 1: I guess because `ktx2.h` is used by both renderer and imagelib, it's being put to `public`? libpublic only contains basic C string, math and buildinfo functions.

imagelib is planned to be split into separate library, so you can store such header in img_ktx2.h in imagelib itself, and include from renderer. Renderers can include private engine headers anyway, except functions, for obvious reason.

Also you missed copyright header here.

public/ktx2.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <stdint.h>
+
+#define KTX2_IDENTIFIER_SIZE 12
```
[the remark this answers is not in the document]
w23: Not sure about this one. The spec explicitly states the size to be 12, and this definition makes it more easily visible.
I can change it if you insist though.

engine/common/imagelib/img_dds.c
w23: Addressed where I could notice it (clang-format when)

engine/common/imagelib/img_ktx2.c
```diff
@@ -0,0 +1,204 @@
...
+
+		// If KTX2 to imagelib conversion failed, try passing the file as raw data.
+		// This is useful for ref_vk which can directly support hundreds of formats which we don't convert to pixformat_t here
+
+		Con_DPrintf( S_WARN "%s: (%s) could not be converted to supported imagelib format, passing as raw KTX2 data\n", __FUNCTION__, name );
+		// This is a catch-all for ref_vk, which can do this format directly and natively
+		image.type = PF_KTX2_RAW;
+
+		image.size = filesize;
+		//image.encode = TODO custom encode type?
```
w23, line 197: I wonder if `image.encode` would be a good place for format semantic flags like `SRGB`, `UNORM/UNSIGNED`, and `SNORM/UNSIGNED`.

a1batross: Not really sure what should go into `encode` field in this case. I'm fine with extending `pixformat_t` anyway.

PR #1470, opened 2023-10-27
Add gameinfo.txt key to make engine use larger lightmaps without modifying the DLL

a1batross:
The DLL requested features still takes precedence over this key. 

Should we warn if gameinfo.txt has use_large_lightmaps 1 but DLL doesn't request it?


a1batross:
cc @mittorn

a1batross:
There is nothing useful in this. 

If we need large lightmaps without modifying the DLL, we can as well make it into a cvar.

PR #1475, opened 2023-10-30
Software ripples for GL renderer

a1batross:
We just update textures here.

This version differs from my older version I did for Yamagi Quake 2, just to be sure, I rechecked it with decompile and compared it against original `sw.dll`, that's why it uses only 64x64 despite drawing texture at 128x128.

I want this merged with the current CPU-based solution, as I believe someone smart will later make a GPU-based solution, and we would be able to compare if it will be any better.

### DIFF
--- a/engine/client/ref_common.c
+++ b/engine/client/ref_common.c
@@ -16,6 +16,7 @@ CVAR_DEFINE_AUTO( gl_msaa_samples, "0", FCVAR_GLCONFIG, "samples number for mult
 CVAR_DEFINE_AUTO( gl_clear, "0", FCVAR_ARCHIVE, "clearing screen after each frame" );
 CVAR_DEFINE_AUTO( r_showtree, "0", FCVAR_ARCHIVE, "build the graph of visible BSP tree" );
 static CVAR_DEFINE_AUTO( r_refdll, "", FCVAR_RENDERINFO, "choose renderer implementation, if supported" );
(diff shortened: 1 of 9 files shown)


a1batross:
Comparison:

Usual `ref_gl` water:

https://github.com/FWGS/xash3d-fwgs/assets/3213697/3aaf3bf6-aa84-4b27-ac8c-5feede862412

Software water in WON Half-Life:

https://github.com/FWGS/xash3d-fwgs/assets/3213697/ee3cb8c9-064c-41aa-bd1e-5fe25b277eeb

Software water in `ref_gl` (with nearest filter):

https://github.com/FWGS/xash3d-fwgs/assets/3213697/f883a860-48cf-4f84-a810-8272ed1f8106 

Software water in `ref_gl` (with linear filter):

https://github.com/FWGS/xash3d-fwgs/assets/3213697/4637ea67-ee10-4447-81ea-976c24bad686

SNMetamorph:
I think we need to add "authentic water" option checkbox to video settings.

Magic-Nipples:
Works like a charm!! Thanks so much for getting this ported to xash and making it work in gl!!

a1batross:
Made faster version that only uploads texture of a selected size. So on green sludge texture which is 128x128, it will generate 128x128 texture (viewable in full glory with `r_ripple 2`), and usual water texture which is 64x64, will render in 64x64.

In `mainui_cpp` there is a checkbox in Video Options now that toggles `r_ripple` between 0 and 1.

That's why I said better not take this code yet :)

PR #1504, opened 2023-11-22
Fix building on OpenBSD 7.4

hgojf:
I tried to build this on OpenBSD 7.4 but ran into a few errors, these patches allow it to build.

### DIFF
--- a/wscript
+++ b/wscript
@@ -409,7 +409,7 @@ def configure(conf):
 			for i in a:
 				conf.check_cc(lib = i)
 	else:
-		conf.check_cc(lib='dl')
+		conf.check_cc(lib='dl', mandatory = False)
 		conf.check_cc(lib='m')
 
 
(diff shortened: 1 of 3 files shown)


wscript
```diff
@@ -405,6 +408,9 @@ def configure(conf):
 		else:
 			for i in a:
 				conf.check_cc(lib = i)
+	elif conf.env.DEST_OS == 'openbsd':
+		#libdl is a part of libc on this platform
```
nekonomicon, line 412: linkage works with `conf.check_cc(lib='dl', mandatory=False)`?

hgojf: Yes that works

a1batross: Could you revert this and add `mandatory=False` flag to libdl check then? I don't think making libdl optional will hurt anything.

a1batross:
Other than `libdl` check, I'm fine with these changes.

hgojf:
Applied that change

a1batross (approved):

a1batross:
Thank you! <3

PR #1518, opened 2023-12-02
Lightmapped water

a1batross:
TODO:
- [x] basic renderer for lightmapped opaque water surfaces
- [x] move renderer into a similar to EmitWaterPolys function to apply lightmaps on waves
- [ ] detect presense of lightmap on water surface, do not build texcoords where it's not needed
- [x] lightmaps on other rendermodes? how?
- [ ] ...

### DIFF
--- a/Documentation/extensions/litwater.md
+++ b/Documentation/extensions/litwater.md
@@ -1,0 +1,19 @@
+## Lightmapped water
+
+Xash3D FWGS supports lightmapped water, as an extension. It adds three new cvars and new worldspawn key values.
+
+### For level designers:
+
+If you're a level designer and intend to make your level to have lightmapped water, you can put these keyvalues to worldspawn entity description (always first entity in entities list):
+
+| Key                    | Value   | Description |
+| ---------------------- | ------- | ----------- |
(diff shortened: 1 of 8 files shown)


a1batross:
[x] Added lightmaps on wavey water and transparent water
[x] Simplified code a lot, easier diff to read :)

a1batross:
Despite that in most cases there is ready to use lightmap for water surfaces, a lot of maps weren't really made with lightmapped water in mind, and it's quite hard to see through it. See water at `gasworks.bsp`, the room isn't lit enough for water doesn't receive the light at all.

So I'm looking for ways to make lightmap on water surfaces less dark in general but keep it the same when it's properly lit. Even simple addition or scale make the situation better, but again, it also makes some lit spots too intense. Like some sort of hyperbolic scaling depending on lightmap value.

Later it must be possible to configure it through `worldspawn` key values.

a1batross:
>Later it must be possible to configure it through worldspawn key values.

Done.

>So I'm looking for ways to make lightmap on water surfaces less dark in general but keep it the same when it's properly lit

Done.

a1batross:
There is a bug though: as we don't draw back faces, the water surface doesn't get drawn when the camera is underwater.

PR #1530, opened 2023-12-13
Legacy android port

mittorn:
This restores old non-sdl android port
Fix unstable engine in background mode (mostly, TODO: debug some edge cases, gles3compat will crash on any frame without context)
Workaround showing messageboxes on crash (every jni call from handler silently exited process)
Fix host-clang on recent clang versions
Fix android_sleep mode
Refactor jni code, restore egl support (EGL core now can be reused in other platforms, move to separate platform section if needed)

Enabled with --enable-android-legacy option
To build apk, add game code project and xash3d-android-project(android-legacy branch) with --extra-projects option, otherwise only native libraries enabled

- Revert "engine: partially remove legacy Android port, in preparation of new port merge"
- waifulib/xcompile: host clang experiments (ndk ld)
- platform/android_nosdl: fix build (ndk10, host clang15)
- wscript: add external projects support (updated waf required)
- waifulib/xcompile: fix host clang16
- platform/android: rewrite egl part to separate not android-specific file
- platform/android: disable swbuffer resize hack, wrong buffer size really is an error that should be fixed on Activity side, not here
- platform/android: debug surface enabling/disabling, add notification for event_set_pause, fix wrong host.status, prevent rendering while no surface
- platform/eglutil: prevent rendering while no surface
- ref_soft,ref_gl: fix unaligned read in sprite code
- imagelib: fix for loop scope
- platform/android: implement dumb dladdr fallback (only searches in server library)
- platform/posix: check for non-zero dladdr ret
- waf/compiler_optimizations: remove no-semantic-interposition for gcc4
- waf/xcompile: fix ld path for host clang
- platform/android: prevent rendering without context, fix android_sleep using surfaceless mode or dummy surface if possible
- ref_soft: fix crash when y < -height
- waifu: set external waifulibs for extra-projects
- wscript: check if extra-projects was set in configuration
- gl2shim: fix gcc 4.9 build
- gl2shim: add SOFTFP_LINK support for gles3compat
- waifulib/xcompile: add android-from-none for new host-clang versions, support hardfloat abi in host-clang
- gl4es: update submodule
- waifulib/xcompile: disable stpcpy on host-clang
- waifulib/xcompile: add objcopy tool to android configuration, allow switching to ndk binutils with host clang
- ref_gl: workaround using fake arb functions on GLES (Mali400 has stub glBindBufferARB somehow\!)
- crashhandler: set host.crashed before messagebox to allow platform code handle crashed state in different way
- vid/android: Remove pixelFormat selection code, it only used to set desktop bits per pixel
- waifulib/compiler_optimizations: disable no-semantic-interposition for gcc4 in release builds too
- platform/android-nosdl: Move all bindings to XashBindings class, workaround broken messageBox from crashhandler
- platform/android-nosdl: fix declaration style (werror)
- platform/android-nosdl: add Platform_Shutdown for preShutdown callback
- engine/system: add asm-based raise implementation for arm, making Sys_DebugBreak more relevant
- platform/android-nosdl: optionally wait gdb in nativeInit, enable dumpable flag early as possible
- engine/system: add aarch64 INLINE_RAISE implementation, use tgkill instead of kill
- platform/android-nosdl: restore SIGTRAP handler
- platform/android-nosdl: add workarounds for yama bullshit
- wscript: fix options in external projects
- scripts/waifulib: save zip task to make it output reuseable
- wscript: re-enable xash-extras on android
- fs/android: only enable aasset code in android+sdl build
- ref_gl: fix loop scope in ripple code

### DIFF
--- a/engine/platform/android/android_nosdl.c
+++ b/engine/platform/android/android_nosdl.c
@@ -23,6 +23,11 @@ GNU General Public License for more details.
 #include <sys/prctl.h>
 #include <setjmp.h>
 
+#ifndef PR_SET_PTRACER
+#define PR_SET_PTRACER 0x59616d61
+#define PR_SET_PTRACER_ANY ((unsigned long)-1)
+#endif
+
 #ifndef JNICALL
 #define JNICALL // a1ba: workaround for my IDE, where Java files are not included
 #define JNIEXPORT
@@ -249,6 +254,7 @@ DECLARE_JNI_INTERFACE( int, nativeInit, jobject array )
 	}
 	argv[argc] = NULL;
 	prctl(PR_SET_DUMPABLE, 1);
+	prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0);
 
 	/* Init callbacks. */
 
@@ -605,6 +611,7 @@ DECLARE_JNI_INTERFACE( int, nativeTestWritePermission, jstring jPath )
 JNIEXPORT jint JNICALL JNI_OnLoad( JavaVM *vm, void *reserved )
 {
 	prctl(PR_SET_DUMPABLE, 1);
+	prctl(PR_SET_PTRACER, PR_SET_PTRACER_ANY, 0, 0, 0);
 	return JNI_VERSION_1_6;
 }
 
--- a/engine/platform/android/android_nosdl.c
+++ b/engine/platform/android/android_nosdl.c
@@ -212,6 +212,12 @@ nativeSetPause
 #define DECLARE_JNI_INTERFACE( ret, name, ... ) \
 	JNIEXPORT ret JNICALL Java_su_xash_engine_XashBinding_##name( JNIEnv *env, jclass clazz VA_ARGS(__VA_ARGS__) )
 
+static int _debugger_present = -1;
+static void _sigtrap_handler(int signum)
+{
+	_debugger_present = 0;
+}
+
 DECLARE_JNI_INTERFACE( int, nativeInit, jobject array )
 {
 	int i;
@@ -280,8 +286,18 @@ DECLARE_JNI_INTERFACE( int, nativeInit, jobject array )
 		return 127; // unreach
 	}
 
-	/* Run the application. */
+	if( getenv( "XASH3D_GDB_WAIT" ))
+	{
+		signal(SIGTRAP, _sigtrap_handler);
+		while(_debugger_present <= 0)
+		{
+			_debugger_present = 1;
+			INLINE_RAISE( SIGTRAP );
+			INLINE_NANOSLEEP1();
+		}
+	}
 
+	/* Run the application. */
 	status = Host_Main( argc, argv, getenv("XASH3D_GAMEDIR"), false, NULL );
 
 	/* Release the arguments. */
@@ -586,6 +602,7 @@ DECLARE_JNI_INTERFACE( int, nativeTestWritePermission, jstring jPath )
 
 JNIEXPORT jint JNICALL JNI_OnLoad( JavaVM *vm, void *reserved )
 {
+	prctl(PR_SET_DUMPABLE, 1);
 	return JNI_VERSION_1_6;
 }
 
--- a/engine/platform/android/android_nosdl.c
+++ b/engine/platform/android/android_nosdl.c
@@ -265,6 +265,7 @@ DECLARE_JNI_INTERFACE( int, nativeInit, jobject array )
 	jni.getGLAttribute = (*env)->GetStaticMethodID(env, jni.bindcls, "getGLAttribute", "(I)I");
 	jni.deleteGLContext = (*env)->GetStaticMethodID(env, jni.bindcls, "deleteGLContext", "()Z");
 	jni.getSurface = (*env)->GetStaticMethodID(env, jni.bindcls, "getNativeSurface", "(I)Landroid/view/Surface;");
+	jni.preShutdown = (*env)->GetStaticMethodID(env, jni.bindcls, "preShutdown", "()V");
 
 	// jni fails when called from signal handler callback, so jump here when called messagebox from handler
 	if( setjmp( crash_frame ))
@@ -272,8 +273,7 @@ DECLARE_JNI_INTERFACE( int, nativeInit, jobject array )
 		// note: this will destroy stack and shutting down engine with return-from-main
 		// will be impossible, but Sys_Quit works
 		(*jni.env)->CallStaticVoidMethod( jni.env, jni.bindcls, jni.messageBox, (*jni.env)->NewStringUTF( jni.env, "crash" ), (*jni.env)->NewStringUTF( jni.env , crash_text ) );
-
-		// java shutdown callback here?
+		(*jni.env)->CallStaticVoidMethod( jni.env, jni.bindcls, jni.preShutdown );
 
 		// UB, but who cares, we already crashed!
 		longjmp( restore_frame, 1 );
@@ -603,7 +603,9 @@ void Android_Init( void )
 
 void Android_Shutdown( void )
 {
-
+	if( host.crashed )
+		return;
+	(*jni.env)->CallStaticVoidMethod( jni.env, jni.bindcls, jni.preShutdown );
 }
 
 /*
--- a/engine/platform/android/android_nosdl.c
+++ b/engine/platform/android/android_nosdl.c
@@ -879,6 +879,7 @@ void Platform_RunEvents( void )
 			{
 //				(*jni.env)->CallStaticVoidMethod( jni.env, jni.actcls, jni.toggleEGL, 0 );
 //				(*jni.env)->CallStaticVoidMethod( jni.env, jni.actcls, jni.toggleEGL, 1 );
+				Android_UpdateSurface( false );
 				Android_UpdateSurface( true );
 				SetBits( gl_vsync.flags, FCVAR_CHANGED ); // set swap interval
 				VID_SetMode();
--- a/engine/platform/android/android_nosdl.c
+++ b/engine/platform/android/android_nosdl.c
@@ -21,6 +21,7 @@ GNU General Public License for more details.
 #include "errno.h"
 #include <pthread.h>
 #include <sys/prctl.h>
+#include <setjmp.h>
 
 #ifndef JNICALL
 #define JNICALL // a1ba: workaround for my IDE, where Java files are not included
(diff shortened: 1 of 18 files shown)


a1batross:
After a quick look, I don't have any objections. Will look further a bit later.

mittorn:
File server with this issues:
Deleting files implemented, but does not work for some reason
Deleting folders unimplemented.

File server TODO:
check zip download/upload on real devices
Uploading folders seems to work ok and very fast, but listing files slow. Why?

Recent build (not published one) have strange regression: barney cannot move after c0a0d

Android 12 issues:
1. Fileserver does not work (fixed, use accept4 syscall first)
2. GDB does not work (fixed, rebuild with musl)
3. Back button sometimes closes engine during hiding console. This seems not reproduce on android 8, but very similar to issues on early new engine port. Maybe device-specific issue?
4. Sometimes getting SIGABRT on exit. It seems to be after exit() or device-specific, but need debug

Still missing:
1. Mod support. I not very interested in ndk20x abi due to platform19 requirement and port not fully compatible with any mods. aarch64 version should be compatible with new mod abi, but armeabi-v7a uses old abi with different file names, so it compatible with nothing
2. Create shortcut feature is ugly and does not support new launchers/android itself
3. Cannot close console when virtual keybpard active
4. CI/testing/update channels?
5. Version information should be collected fromall projects
6. Localization for new features
7 New launcher interface? There are two ideas: mod list in the launcher and calling laucher from mod with extra data
8. Installation instructions for internal folder

filesystem/android.c
```diff
@@ -15,7 +15,7 @@ GNU General Public License for more details.
 
 #include "port.h"
 
-#if XASH_ANDROID
+#if XASH_ANDROID && XASH_SDL
```
a1batross, line 18: There is nothing non-SDL in these functions. Could you either provide stubs on Java side or stub the functions inside `android.c`?

ref/gl/gl_opengl.c
```diff
@@ -585,6 +589,9 @@ qboolean GL_CheckExtension( const char *name, const dllfunc_t *funcs, const char
 				if(( f = gEngfuncs.GL_GetProcAddress( name )))
 				{
 					// GL_GetProcAddress prints errors about missing functions, so tell user that we found it with different name
+#ifdef XASH_GLES
+					if(i != 0)
```
a1batross, line 593: This is so ugly. I don't know what to do about this, but this better be refactored somehow.

ref/gl/gl_sprite.c
```diff
@@ -200,22 +202,25 @@ void Mod_LoadSpriteModel( model_t *mod, const void *buffer, qboolean *loaded, ui
 
 	for( i = 0; i < mod->numframes; i++ )
 	{
-		frametype_t frametype = pframetype->type;
+		frametype_t frametype;
+
+		memcpy( &dframetype, pframetype, sizeof( dframetype ));
```
a1batross, line 207: Do you really need `dframetype` in outer scope? 

Maybe its scope could be lowered to this block?

ref/soft/r_sprite.c
```diff
@@ -200,22 +201,25 @@ void Mod_LoadSpriteModel( model_t *mod, const void *buffer, qboolean *loaded, ui
 
 	for( i = 0; i < mod->numframes; i++ )
 	{
-		frametype_t frametype = pframetype->type;
+		frametype_t frametype;
+
+		memcpy( &dframetype, pframetype, sizeof( dframetype ));
```
a1batross, line 206: Same as in ref_gl.

scripts/waifulib/xcompile.py
```diff
@@ -240,9 +241,22 @@ def strip(self):
 			return os.path.join(self.gen_binutils_path(), 'llvm-strip')
 		return os.path.join(self.gen_binutils_path(), 'strip')
 
+	def objcopy(self):
+		if self.is_host() and not self.ndk_binutils:
+			environ = getattr(self.ctx, 'environ', os.environ)
```
a1batross, line 246: You don't need `getattr` here. 

Context already should have environ (we're literally overriding compilers and tools through it)

wscript
```diff
@@ -102,6 +102,42 @@ REFDLLS = [
 	RefDll('gles3compat', False),
 ]
 
+def process_extra_projects_opts(ctx):
```
a1batross, line 105: IMO extra projects thing should migrate to waifulib/subproject.py after a while...

a1batross (changes requested):

engine/platform/android/android_nosdl.c
```diff
@@ -0,0 +1,1068 @@
...
+*/
+#define VA_ARGS(...) , ##__VA_ARGS__ // GCC extension
+#define DECLARE_JNI_INTERFACE( ret, name, ... ) \
+	JNIEXPORT ret JNICALL Java_su_xash_engine_XashBinding_##name( JNIEnv *env, jclass clazz VA_ARGS(__VA_ARGS__) )
+
+static volatile int _debugger_present = -1;
+static void _sigtrap_handler(int signum)
+{
+	_debugger_present = 0;
+	__android_log_print( ANDROID_LOG_ERROR, "XashGDBWait", "It's a TRAP!" );
```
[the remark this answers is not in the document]
nekonomicon: Lol!

TheKingFireS:
Crash report on HTC Desire 610.
Engine: [android-legacy-port](https://github.com/mittorn/xash3d-android-project/releases/tag/android-legacy-prerelease) (Not only this, also very same exactly happened to all Xash3D forks that use FWGS's xash3d-android-project)
To reproduce crash: Just tap on Xash3D app icon to launch, then it crashed instantly.
ROM: HTC Sense 6.0
OS: Android 4.4.2
Logcat: [cat.log](https://github.com/FWGS/xash3d-fwgs/files/13692029/cat.log) (That repeated same xash activity is me attempt to launch Xash3D multiple times.)

mittorn:
Strange layout inflater bug on HTC. This dialog works OK even on 1.5, 2.3, 4.2, 7.0 and 8.1. I'll include workaround to next build

mittorn:
Fixed dlsym-weak codestyle. I not sure using auto in non-deduced type and not before very complex casts is good idea.
RTLD_NEXT does not seems to search library in linker, and it is even unclear for me, what RTLD_NEXT should do in this case. dlopen just returns soinfo stub from linker on android, so it seems to be best solution, in never unloads anyway.

engine/platform/android/dlsym-weak.cpp
```diff
@@ -78,21 +114,75 @@ static unsigned elfhash(const char* _name) {
...
  */
-static Elf_Sym* dlsym_handle_lookup(soinfo* si, const char* name) {
-    return soinfo_elf_lookup(si, elfhash(name), name);
+static Elf_Sym* dlsym_handle_lookup( soinfo* si, const char* name )
+{
+	return soinfo_elf_lookup( si, elfhash( name ), name );
 }
 
-extern "C" void* dlsym_weak(void* handle, const char* symbol) {
+#include "lib_android.h"
```
a1batross, line 122: Includes better be at the start of the file.

engine/platform/android/dlsym-weak.cpp
```diff
@@ -78,21 +114,75 @@ static unsigned elfhash(const char* _name) {
...
+
+static int dladdr_fallback( const void *addr, Dl_info *info )
+{
+	static soinfo *server_info;
+	Elf_Sym *sym;
+
+	if( !server_info )
+		server_info = ( soinfo* )ANDROID_GetServerLibrary();
```
a1batross, line 130: You called `ANDROID_` here when the function is called `Android_`.

I think it won't compile.

engine/platform/platform.h
```diff
@@ -247,4 +251,95 @@ void VoiceCapture_Shutdown( void );
 qboolean VoiceCapture_Activate( qboolean activate );
 qboolean VoiceCapture_Lock( qboolean lock );
 
+#ifdef XASH_LINUX
+// this allows to make break in this line, not somewhere in libc
+// libc builds with -fomit-frame-pointer may just eat stack frame (hello, glibc), making this useless
+// calling syscalls directly allows to make break like if it was asm("int $3") on x86
+#ifdef __arm__
+// this multi-line macro hides single line under it, multi-line macro will be expanded to single-line anyway
```
a1batross, line 259: Unuseful comment. Everybody who ever looked into `gcc -E` output knows how C preprocessor works.

engine/platform/platform.h
```diff
@@ -247,4 +251,95 @@ void VoiceCapture_Shutdown( void );
...
+		int raise_sig = (x); \
+		__asm__ volatile (  \
+			"mov r7,#268\n\t" \
+			"mov r0,%0\n\t" \
+			"mov r1,%1\n\t" \
+			"mov r2,%2\n\t" \
+			"svc 0\n\t" \
+			: \
+			:"r"(raise_pid), "r"(raise_tid), "r"(raise_sig) \
+			: "r0", "r1", "r2", "r7", "memory"
```
a1batross, line 273: Forgotten `\` at the end of line.

a1batross (changes requested):

engine/platform/android/dlsym-weak.cpp
```diff
@@ -32,63 +32,71 @@
...
     Elf_Sym* symtab = si->symtab;
     const char* strtab = si->strtab;
 
 	if( si->nbucket == 0 )
 		return NULL;
 
-    for (unsigned n = si->bucket[hash % si->nbucket]; n != 0; n = si->chain[n]) {
+	for( unsigned n = si->bucket[hash % si->nbucket]; n != 0; n = si->chain[n] )
+	{
         Elf_Sym* s = symtab + n;
```
lewa-j, line 48: spaces mixed with tabs

a1batross: I don't think it should be fixed right now. 

The original file used spaces and nobody noticed. We can convert it after merge.

amirwolf512lo:
When will you release it officially?
Delete the old version

a1batross:
@amirwolf512lo what do you even mean

amirwolf512lo:
> @amirwolf512lo what do you even mean

You do not have players statistics
They all use this [version](https://github.com/FWGS/xash3d)
If you restrict access to it, they will be forced to install the new version
Or you don't do it on purpose

nekonomicon:
> They all use this version
We know.

@amirwolf512lo 
Probably  in the next half year after stable release will be disabled access to an old servers.

a1batross:
It will be decided after a first truly stable Android release of the new engine.

It depends on how soon players and server owners will migrate to the new protocol. I hope for some cooperation in this.

a1batross:
Anyway, it's not the place to discuss such things. If you want to give your ideas on post-first-Android-stable-release, make a new issue.

a1batross:
Not viable to support.

Android is a dying platform, and we do not have enough developers who can work on Android.

I kept the branch, just in case.

PR #1544, opened 2023-12-27
Fix engine/renderer timings synchronization. Bump to RefAPI 6

a1batross:
Discussion on SourceHut: https://lists.sr.ht/~a1batross/xash3d-fwgs/patches/48142

This set of patches fixes the earliest bug since renderers split: timings.

It eliminates the need to synchronize renderers timings by passing a pointer from engine to renderer in a form of API structure. The structure in the API was designed to be compatible with internal engine structures that holds these values.

It also fixes another bug: truncated client time. For some reason it was in single-precision float all this time and nobody noticed.

Possible Questions:

Q: Why have structures when it's possible to pass separate pointers to double?

A: I had this in the non-published version of RefAPI 6 however turning `gpGlobals->time` into a pointer enables dereferencing it. The `->` operator does two things here: it will be always evaluated FIRST (as of C standard) and it already dereferences the pointer. I thought it's more aesthetically pleasant to work with a structure than a single variable pointer. The downside of this is a set of static assertions in the engine, however it's an engine problem to keep it compatible.

Q: Why not to rename everything in the engine to use these shared structures?

A: I considered this too but didn't want to go with this, as replacing every `cl.{time,oldtime}` and `host.{realtime,frametime}` into something else would probably not only touch hundreds of lines in renderers but thousands in the engine. Also calling client time `cl.time` just makes sense.

Changes since v1:
* As suggested by mittorn, we now use PARMs to get engine pointers.
* It also eliminates older functions that were introduced in RefAPI 5.
* ref_client_t and ref_host_t now expose more data to avoid excessive EngineGetParm calls. Though not everything was replaced, it's outside of the scope of this patch set and can be easily fixed later without breaking compatibility. 

Changes since v2:
* No changes. Sorry, I put wrong in-reply-to.

### DIFF
--- a/engine/client/cl_parse.c
+++ b/engine/client/cl_parse.c
@@ -1642,9 +1642,6 @@ void CL_RegisterResources( sizebuf_t *msg )
 
 			CL_ClearWorld ();
 
-			// update the ref state.
-			R_UpdateRefState ();
-
 			// tell rendering system we have a new set of models.
 			ref.dllFuncs.R_NewMap ();
 
--- a/engine/client/cl_parse_48.c
+++ b/engine/client/cl_parse_48.c
@@ -660,9 +660,6 @@ void CL_LegacyPrecache_f( void )
 	if( clgame.entities )
 		clgame.entities->model = cl.worldmodel;
 
-	// update the ref state.
-	R_UpdateRefState ();
-
 	// tell rendering system we have a new set of models.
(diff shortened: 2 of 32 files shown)


a1batross:
Can somebody review that here? 

@nekonomicon @SNMetamorph @w23

a1batross:
Applied.

PR #1560, opened 2024-01-05
engine: platform: linux: Add gettid definition for old systems

rpiasetskyi:
The `gettid()` library support was added in glibc 2.30. Earlier glibc versions did not provide a wrapper for this system call, necessitating the use of `syscall(2)`.

Also, put `_GNU_SOURCE` definition in the guard to avoid the warning:
```
  ../engine/platform/linux/sys_linux.c:16: warning: "_GNU_SOURCE" redefined
   #define _GNU_SOURCE

  <command-line>: note: this is the location of the previous definition
```

### DIFF
--- a/engine/platform/linux/sys_linux.c
+++ b/engine/platform/linux/sys_linux.c
@@ -13,7 +13,10 @@ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
 */
 
+#ifndef _GNU_SOURCE
 #define _GNU_SOURCE
+#endif
+
 #include <time.h>
 #include <stdlib.h>
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


a1batross:
_GNU_SOURCE could be added from command line, sometimes even implicitly thanks to pkg-config files. The definition is intended to be added at the first lines of the source file, so I don't think this warning should be fixed.

Otherwise, I'm fine with these changes.

rpiasetskyi:
Thanks for the review!
Does it make sense to make something like this to avoid the warning?
```
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
```

If no, I will just return the original line.

a1batross:
Yes, it's fine to me.

a1batross (approved):

Issue #1559, opened 2024-01-05
IPv6 in sv_downloadurl


d47081:
Does `sv_downloadurl` in `valve/server.cfg` support IPv6?

I can download resources trough IPv4 protocol only - just need IPv6 for Yggdrasil connections.

mittorn:
No, it's ipv4-only

a1batross:
The HTTP client implementation is in net_ws.c file. I believe, it only needs to specify any matching protocol family instead of just IPv6 for our getaddrinfo/gethostbymame wrapper function.

sv_downloadurl however only allows to set only one address.

mittorn:
Old xash3d allowed to specify multiple urls in sv_downloadurl delimited with spaces. Is not new engine allow this?

oooo-ps:
Is this option not officially implemented to work with IPv6?

If no, can I disable this feature on the client side, to prevent external (internet) connections?

nekonomicon:
> Is this option not officially implemented to work with IPv6?

It's fastdl %)
And new cvar from this PR is completely useless.

PR #1577, opened 2024-01-15
IPv6 support for sv_downloadurl (sv_downloadurl_ipv6)

d47081:
Solution compatible with previous protocol / clients as implements new option `sv_downloadurl_ipv6`

Details described in #1559

Client-side already tested - I can provide IPv6 links in `server.cfg` and download resources from there, but stuck on server-side part, as don't know how to check remote connection type and return `sv_downloadurl_ipv6` for IPv6 clients instead of `sv_downloadurl`

Here is the subject with example:
https://github.com/FWGS/xash3d-fwgs/issues/1559#issuecomment-1892163611


a1batross:
I'm not sure whether having two sv_downloadurl cvars is actually needed.

In my idea, by having both AAAA and A DNS records pointing at the same domain, the client would be free to choose whether it wants to use IPv6 connection or IPv4 connection.

Also, the way svc_resourcelocation parser works, it actually allows multiple locations to be set, by listing all available URLs in `sv_downloadurl` separated by space.

a1batross:
The same goes for having multiple locations in `sv_downloadurl`.

It's the client's choice whether they want to use IPv6 or IPv4.

a1batross:
#1862 now allows the usage of IPv6 for HTTP resources location, by passing AF_UNSPEC to resolver and then passing resolved address family to `socket()`. I haven't tested it yet, though.

But the idea is still the same: whatever resolver picks is going to be used here.

PR #1628, opened 2024-02-26
Connection bandwidth test fixes

SNMetamorph:
- Fixed bug when `svs.testpacket_buf` was not being created on server at all
- Reduced overall server connection time (test probe delay reduced to 1.0 seconds)
- Count of test probes changed to 5, probe sizes are 64000, 32000, 10666, 5200, 1400
- Allowed to use bandwidth test on listen servers (why it even was disabled?)
- Fixed reconnection attempts after bandwidth test (they wasn't working as intended - just sending `bandwidth` packets instead of `getchallenge`)

### DIFF
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -28,7 +28,6 @@ GNU General Public License for more details.
 #define MAX_CMD_BUFFER		8000
 #define CONNECTION_PROBLEM_TIME	15.0	// 15 seconds
 #define CL_CONNECTION_RETRIES		10
-#define CL_TEST_RETRIES_NORESPONCE	3
 #define CL_TEST_RETRIES		5
 
 CVAR_DEFINE_AUTO( mp_decals, "300", FCVAR_ARCHIVE, "decals limit in multiplayer" );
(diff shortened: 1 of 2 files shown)


engine/client/cl_main.c
```diff
@@ -1072,6 +1071,22 @@ static void CL_SendConnectPacket( void )
...
+/*
+=================
+CL_GetTestFragmentSize
+
+Returns bandwidth test fragment size
+=================
+*/
+static int CL_GetTestFragmentSize( void )
+{
+	const int fragmentSizes[CL_TEST_RETRIES] = { 64000, 32000, 10666, 5200, 1400 };
```
a1batross, line 1083: Huh. Ideally we could save all CRCs for these values on server for quicker check. But that's something that could be done later.

a1batross (approved):

engine/client/cl_main.c
SNMetamorph: Also, instead of relying on gfx.wad file content, it's better to use pseudorandom sequence of bytes with "pre-shared" seed. This should be more reliable.

mittorn:
> Allowed to use bandwidth test on listen servers (why it even was 
> disabled?)

Maybe it was disabled because of possible flood attacks while engine is 
running and server port active?

PR #1629, opened 2024-02-29
utils: mdldec: boneweights fix.

nekonomicon:


utils/mdldec/qc.c
```diff
@@ -601,6 +601,12 @@ void WriteQCScript( void )
 	fputs( "$scale 1.0\n", fp );
 	fputs( "\n", fp );
 
+	if( model_hdr->flags & STUDIO_HAS_BONEINFO )
+	{
+		if( model_hdr->flags & STUDIO_HAS_BONEWEIGHTS )
```
lewa-j, line 606: `if(( model_hdr->flags & STUDIO_HAS_BONEINFO ) && ( model_hdr->flags & STUDIO_HAS_BONEWEIGHTS ))`

nekonomicon: What about jiggle bones?

PR #1643, opened 2024-03-16
Minor logging improvements

SNMetamorph:
Provides better experience processing users feedback about bugs. Earlier, we couldn't be able to know are user turned on `-dev` feature for more verbose logs or not. Also, we wasn't be able to know are user downloaded engine build from "master" branch rather than experimental one (that's common problem, on GitHub it's kinda confusing for normies)


a1batross:
Merged, though, I changed output format to make it shorter and easier to read.

VCS branch detection is implemented too.

PR #1653, opened 2024-04-03
Gl4

HH45137:


SNMetamorph:
Nice to see that thing. Are you planning to completely avoid compatibility profile and using core profile instead? This is what we need in some cases, because old versions of OpenGL with compatibility profile works pretty bad.

a1batross:
@SNMetamorph there is already a core context mode, enabled with `-glcore`. It automatically enables gl2_shim and might decrease performance in some ways, but it might be very useful for custom renderers like yours.

As for this PR, it's got closed, and source repository got deleted. I guess that wasn't supposed to be published yet. :)

Issue #1662, opened 2024-04-18
kRenderFxGlowShell rendering incorrectly


Hedgefog:
Entities with `kRenderFxGlowShell` effect render as transparent

![image](https://github.com/FWGS/xash3d-fwgs/assets/7786660/aad3f310-c2a8-4a97-a2a8-eb448d423ff6)

It also has an issue related to color, if the entity has `kRenderFxGlowShell` render effect and `rendercolor` set to `{0, 0, 0}` rendered glow shell will be white.

nekonomicon:
All good, but can you provide info on which mods it may be tested and show screenshot how it look under GoldSource?

Hedgefog:
> All good, but can you provide info on which mods it may be tested and show screenshot how it look under GoldSource?

It shouldn't be transparent. 

![20240419044021_1](https://github.com/FWGS/xash3d-fwgs/assets/7786660/a480e447-3d2c-481f-92ec-cb5bfce1c7c4)
![20240419044040_1](https://github.com/FWGS/xash3d-fwgs/assets/7786660/f61512d7-c4c9-4504-bca9-f0ba799118de)

The mod I have issue with: https://github.com/Hedgefog/cs-halloween-mod

a1batross:
I see some weird ordering problem on screenshot above.

It's like the green bridge gets drawn on top of skeleton with glowshell effect.

Hedgefog:
> I see some weird ordering problem on screenshot above.
> 
> It's like the green bridge gets drawn on top of skeleton with glowshell effect.

I've tested the engine with other maps and mods and it looks like the depth buffer is broken not only for glowshell entities. Any transparent entities behind may be visible through the other entities even if these entities have no transparency or render mode.

a1batross:
What's the value of gl_vbo and which renderer do you use?

Hedgefog:
> What's the value of gl_vbo and which renderer do you use?

I tried with a different renderer, it has the same issue with the phone and PC. I used the default `gl_vbo` value which equals `0`.

PR #1665, opened 2024-04-22
ref: if the entity is opaque, check the renderfx too, it might become translucent

a1batross:
Attempt to fix https://github.com/FWGS/xash3d-fwgs/issues/1662

cc @Hedgefog


a1batross:
Not sure if it's the right fix for the issue, because it doesn't happen under GoldSrc and GoldSrc doesn't check renderfx when deciding where to put the entity in solid entities list or in transparent entities list.

But I'll merge it. Hopefully nothing breaks.

PR #1685, opened 2024-06-08
Support building for Windows-on-ARM targets

tuldok89:
I had this lying around since 2021. Might as well upload this, considering that MS is pushing hard for Windows on ARM devices.

These are the minimum required changes to make the codebase build for both ARM32 and ARM64 targets on Windows. Tested to work on both ARM and ARM64 targets.

References:
- [Overview of ARM64 ABI conventions](https://learn.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions)
- [Overview of ARM32 ABI Conventions](https://learn.microsoft.com/en-us/cpp/build/overview-of-arm-abi-conventions)


a1batross:
Thank you very much!

PR #1687, opened 2024-06-10
More verbose reporting of engine build information to masterserver

SNMetamorph:
This can be pretty useful for statistical purposes and for tracking how many users using outdated/vulnerable Xash3D builds, and we can enforce such users to immediately update their Xash3D installation, for example. 

But masterserver should be updated to support such things too.

Reported information:
1. Xash3D version string
2. Build number
3. VCS commit hash
4. VCS branch name
5. Platform name
6. Architecture name

### DIFF
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -1635,9 +1635,17 @@ static size_t NONNULL CL_BuildMasterServerScanRequest( char *buf, size_t size, u
 #ifndef XASH_ALL_SERVERS
 	Info_SetValueForKey( info, "gamedir", GI->gamefolder, remaining );
 #endif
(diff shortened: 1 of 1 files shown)


a1batross:
Use Info_SetValueForKey, so we can continue parsing it on master server
without changing the logic of using infostrings and acquire data based on
predefined set of keys.

You can look up which keys we're already sending to server during signon
(so called useragent).

SNMetamorph:
> Use Info_SetValueForKey, so we can continue parsing it on master server without changing the logic of using infostrings and acquire data based on predefined set of keys. 

Done.

a1batross (approved):
It can be improved (like there is Info_SetValueForKeyf function that can be used instead of Q_snprintf here) but that would be nitpicking. :)

LGTM.

PR #1689, opened 2024-06-11
Refapi 8 logo

a1batross:
Now that in RefAPI 8 we have a less intrusive (to smelly decal code) way to control decal scale sizes, I brought back @SNMetamorph's suggestion.

Except few things:
* `hpk_maxsize` is increased to 8 megabytes, however I think most players won't get this, since Valve sets this cvar to 4 in HL and CS (and for some reason it's 15.4 in TFC?). Not to mention, some less tech-savvy players set this cvar to absolutely random values, expecting it to magically improve their shooting skills. Shall we rename it and enforce our own defaults here? 
* Logo max size is now 768 pixels, which corresponds to 2.25 megabytes of VRAM. In the worst possible case, 32 players spamming high quality clan logos will take 72 megabytes. I think, this isn't too much for modern PCs and Android devices.
* Customization and HPAK entry size is limited to 256 kilobytes. Again, in the worst possible case, with HPAK limit set to 8 megabytes, the engine will be able to cache logos at least for 32 players, which neatly fits our maxplayers limit and 16 players with Valve's `hpk_maxsize` default. 

In my opinion, 768px logo re-compressed with `pngquant` looks pretty awesome. If using this tool feels inconvenient (strangely, I wasn't able to find an online version of it), turns out it has a C API and can be easily linked to the menu.dll where player logo UI resides anyway.


a1batross:
Of course, this intended to be merged after `refapi-8` branch, as depends on one new function from it.

a1batross:
Merged.

PR #1762, opened 2024-08-03
Enable vsync by default

glebm:
This is the first setting I change after installing xash3d-fwgs. Some distros even patch xash3d-fwgs to enable vsync by default (e.g. [Batocera](https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/ports/xash3d/xash3d-fwgs/0001-Enable-VSync-by-default.patch)).

Let's enable vsync by default here.

### DIFF
--- a/engine/client/ref_common.c
+++ b/engine/client/ref_common.c
@@ -10,7 +10,7 @@ ref_globals_t refState;
 
 static const char* r_skyBoxSuffix[SKYBOX_MAX_SIDES] = { "rt", "bk", "lf", "ft", "up", "dn" };
 
-CVAR_DEFINE_AUTO( gl_vsync, "0", FCVAR_ARCHIVE,  "enable vertical syncronization" );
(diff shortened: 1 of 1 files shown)


SNMetamorph:
Instead, I think, we need some kind of platform-specific default configs. Not only v-sync required to be enabled by default sometimes: such platforms as Nintendo Switch and PlayStation Vita required fully customized controls layout, for example. For now we don't have some proposal for it.

glebm:
Enabled is a better default for the vast majority of platforms, so I think it should be on by default, and specific platforms that benefit from having it off can turn it off (if there even are such platforms).

a1batross:
Oh, I totally forgot about this PR, sorry.

I don't mind enabling VSync by default, it's good for everyone. Also, it will make the experience smoother for those who have high refresh rate displays.

For those who don't like it, myself included, there is an option menu sitting there for years.

Enabling VSync, if it's not supported by the platform, won't do anything as `SDL_GL_SetSwapInterval` would just print an error to the console. If there will be platforms that better have gl_vsync disabled by default, we already have `defaults.h` for that.

glebm:
Thank you!

Issue #1743, opened 2024-07-28
[OpenGL] Some decals are rendered with artifacts


OlegAckbar:
Here are examples of such decals
My system:
Operating System: Arch Linux 
KDE Plasma Version: 6.1.3
KDE Frameworks Version: 6.4.0
Qt Version: 6.7.2
Kernel Version: 6.10.2-2-cachyos (64-bit)
Graphics Platform: Wayland
Processors: 8 × Intel® Core™ i7-7700K CPU @ 4.20GHz
Memory: 15.6 ГиБ of RAM
Graphics Processor: AMD Radeon RX 6700 XT

![Снимок экрана_20240728_154220](https://github.com/user-attachments/assets/3c9806e6-abb1-4711-90b0-d755523dd2b5)
![Снимок экрана_20240728_154143](https://github.com/user-attachments/assets/1e522700-7656-495e-a76a-50ab0f2b5795)
![Снимок экрана_20240728_153828](https://github.com/user-attachments/assets/61b7d048-8935-4034-8b48-9d8b1396ae50)
![Снимок экрана_20240728_153811](https://github.com/user-attachments/assets/abc16982-0043-4328-8384-391f6482352a)
![Снимок экрана_20240728_153801](https://github.com/user-attachments/assets/f17cafe1-78cc-402a-8e2b-a838a258979c)

a1batross:
Does `gl_vbo` is enabled?

Which renderer do you use? `r_refdll_loaded` in console.

Also, run engine with `-dev 2 -gldebug -log` arguments and post engine.log here.

OlegAckbar:
@a1batross 

1. `gl_vbo` was not enabled
2. `r_refdll_loaded" is "gl" ( "" )`
3. [engine.log](https://github.com/user-attachments/files/16404559/engine.log)

TacoDeBoss:
Pitching in some info -- I experienced this exact same thing last night and spent a few hours debugging it. This issue manifests after compiling on my Arch Linux machine with gcc 14.x. A bisect was impossible because no matter how far I went back, it was still happening.

![xash-arch](https://github.com/user-attachments/assets/784c1c66-5996-4e6f-a393-78f9e2c310ae)


However, I discovered that the latest master branch compiled in an Ubuntu 20.04 VM with gcc 9.3.0 produces a build without the issue.

![xash-ubuntu](https://github.com/user-attachments/assets/55463f43-bed9-4adf-926e-87d5ce48d502)

[...]

PR #1767, opened 2024-08-08
Use C99 flexible array members

a1batross:
Now that we're on C99, nothing stops us from normally using the standard, non-UB, non-compiler-specific way for flexible array members.

The only somewhat breaking change is the `glpoly_t` struct redefinition. But because it's never allocated on client side, nor it's never a part of another struct, it doesn't break binary compatibility. Ideally, mods better take this change too, because it less confuses the compiler.

Fixes https://github.com/FWGS/xash3d-fwgs/issues/1743

cc @mittorn

### DIFF
--- a/common/com_model.h
+++ b/common/com_model.h
@@ -127,13 +127,22 @@ typedef struct
 	int		flags;		// sky or slime, no lightmap or 256 subdivision
 } mtexinfo_t;
 
+// a1ba: changed size to avoid undefined behavior. Check your allocations if you take this header!
+// For example:
+//  before: malloc( sizeof( glpoly_t ) + ( numverts - 4 ) * VERTEXSIZE * sizeof( float ))
+//  after (C): malloc( sizeof( glpoly_t ) + numverts * VERTEXSIZE * sizeof( float ))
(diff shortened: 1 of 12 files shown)


mittorn:
Is not it break when trying to build non-c99 mod with engine headers? 
Maybe add new separate definition?
If this does not break mods, it's ok, i have not really checked how it's 
used outside of engine, but it is in shared headers, so need think twice 
before changing it's definition

a1batross:
It might break building in C++, however most mainstream compilers that support C99, allow `[]` in C++ mode.

Or in `com_model.h` we can do this:
```
#ifdef __cplusplus 
int foo[1];
#else
int foo[];
#endif 
```

a1batross:
I modified the declaration in header, that's shared between C and C++ code.

Seems OK so far. If no objections will be made, I will merge it.

mittorn:
How will sizeof() work in that case?

a1batross:
@mittorn I literally explained it in the comment

a1batross:
Merged.

glpoly_t has been renamed to glpoly2_t.

PR #1810, opened 2024-09-02
ref: disabled verbose reporting about tracer invalid color index

SNMetamorph:
Fixes messages spamming in mods such as Sven Co-op 4.8

### DIFF
--- a/common/r_efx.h
+++ b/common/r_efx.h
@@ -56,6 +56,8 @@ color24 gTracerColors[] =
 };
 */
 
+#define TRACER_COLORINDEX_DEFAULT 4
+
 // Temporary entity array
 #define TENTPRIORITY_LOW	0
 #define TENTPRIORITY_HIGH	1
--- a/engine/client/cl_efx.c
+++ b/engine/client/cl_efx.c
@@ -250,7 +250,7 @@ static particle_t *R_AllocTracer( const vec3_t org, const vec3_t vel, float life
 	VectorCopy( vel, p->vel );
 	p->die = cl.time + life;
 	p->ramp = tracerlength.value;
-	p->color = 4; // select custom color
+	p->color = TRACER_COLORINDEX_DEFAULT; // select custom color
 	p->packedColor = 255; // alpha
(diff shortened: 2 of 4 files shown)


a1batross:
What's the point of this? Why not to turn it into DPrintf or Reportf, if it's bothering you?

We still have bugs like #388 and it won't fix the issue for them, just like it less noticeable and harder to debug.

SNMetamorph:
> We still have bugs like #388 and it won't fix the issue for them, just like it less noticeable and harder to debug.

It's not fixable from engine side, because this is bug of mod itself. GoldSrc does not any validation of color index, but still have buffer with 12 colors just like as we have. Therefore, any index above 12 will cause undefined behavior in such cases, while no reporting anything about invalid color index. This PR clarifies behavior in such cases, and fixes not reason, but consequences of #388

a1batross:
So you're telling GoldSrc just reads whatever is out of bounds color array and interprets as an RGB...

Still, it would be nice to have such message for mod developers at higher dev level, though it might be better to catch it not when it draws, because it will be printed each frame, but when it creates. Then it should be in cl_efx.c. Also, I don't think we will increase tracers color set anyway, as there is no real reason for this for now. :)

SNMetamorph:
> So you're telling me, GoldSrc just reads whatever is out of bounds color array and interprets as an RGB...

Exactly. In function `R_DrawParticles`:
изображение

SNMetamorph:
> Still, it would be nice to have such message for mod developers at higher dev level, though it might be better to catch it not when it draws, because it will be printed each frame, but when it creates

Yes, this is good idea. But somehow we should have option to disable these messages for mods that won't be fixed anyway because they was abandoned.

a1batross:
I'll merge it for now, restoring the message might be done later.

GoldSrc seems to print it only in R_UserTracerParticle impl.

a1batross:
I wonder if we could replicate the broken values that GoldSrc from those versions had. If they don't get changed, of course.

PR #1825, opened 2024-10-04
Fix CenterPrint Wrong X Position

xiaodo1337:


a1batross:
If it's measured in virtual coordinates, then you could also fix CL_AdjustYPos and check that returned value is checked against virtual coordinates too plus converted to real coordinates.

CenterPrint could always use real coordinates anyway, it's completely an engine-side feature.

Also, please follow the CONTRIBUTING.md.

xiaodo1337:
Sorry, I will notice that next time. :o

a1batross (approved):

PR #1830, opened 2024-10-07
Added handling SIGTERM signal for proper stopping dedicated server

SNMetamorph:


engine/platform/posix/sys_posix.c
```diff
@@ -145,6 +146,16 @@ void Posix_Daemonize( void )
...
 }
 
+static void Posix_SigtermCallback( int signum )
+{
+	Sys_Quit();
+}
+
+void Posix_SetupSigtermHandling( void )
+{
+	signal( SIGTERM, Posix_SigtermCallback );
```
a1batross, line 156: Better use `sigaction(2)`, it's more portable than `signal(2)`.

a1batross (changes requested):

engine/platform/posix/sys_posix.c
a1batross: The example is in `crashhandler.c` btw.

SNMetamorph: Ok, I did it.

SNMetamorph:
Unfortunately, PS Vita knows nothing about `sigaction`, so I decided to add `#if` condition to skip it there. Anyway, there is no possibility on PS Vita to properly working with processes and signals like on other normal OSes.

a1batross (approved):

PR #1836, opened 2024-10-09
Documentation: added article about GoldSrc network protocol support

SNMetamorph:


Documentation/goldsrc-protocol-support.md
```diff
@@ -0,0 +1,14 @@
+# Support for GoldSrc network protocol
+This feature is still work-in-progress, but for now it's available for all users, and we appreciate any bug-reports and contributions around it.
+
+For connecting to GoldSrc-based servers, use this command:
+```
+connect ip:port goldsrc
```
a1batross, line 6: There is an alternative shorter form:

connect ip:port gs

Documentation/goldsrc-protocol-support.md
```diff
@@ -0,0 +1,14 @@
...
+```
+
+But keep in mind, there are requirement for server to be able to accept connections from Xash3D-based clients: it should accept HLTV connections.
+Without this requirement, you will just get "Steam validation rejected" error on connecting.
+
+That is because proper authorization with Steam API is not implemented in engine yet (but we have plans on it), instead, on authorization stage Xash3D presenting itself for server as HLTV.
```
a1batross, line 12: It's wasn't really an intention to present ourselves as HLTV..?

We're just sending bogus Steam ticket, it's all zeros.

PR #1843, opened 2024-10-13
Restore r_shadows

54ac:
This restores the r_shadows cvar as it's also been re-enabled in HL25 and seems to be working fine.

### DIFF
--- a/ref/gl/gl_studio.c
+++ b/ref/gl/gl_studio.c
@@ -154,8 +154,7 @@ void R_StudioInit( void )
 
 	Matrix3x4_LoadIdentity( g_studio.rotationmatrix );
 
-	// g-cont. cvar disabled by Valve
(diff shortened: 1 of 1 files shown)


a1batross:
They still look ugly as hell, I don't understand why people want them.

54ac:
They're not too bad with the Goldsrc z-fighting fixed IMO.

a1batross:
They _are_ bad on small or leaned surfaces, they also bad because it only approximate lighting direction from the lightmap. There is just no information for that.

nekonomicon:
Emm... And how currently works mods that uses dirty hack to engine?

nekonomicon:
@Qwertyus3D what you can say?

Qwertyus3D:
Zombie-X DLE, maybe some HL Alpha mods...

PR #1875, opened 2024-11-09
Documentation updates and additions

SNMetamorph:


Documentation/goldsrc-protocol-support.md
```diff
@@ -6,9 +6,10 @@ For connecting to GoldSrc-based servers, use this command:
 connect ip:port gs
 ```
 
-But keep in mind, there are requirement for server to be able to accept connections from Xash3D-based clients: it should accept HLTV connections.
+But keep in mind, there are requirement for server to be able to accept connections from Xash3D-based clients: it should use Reunion or Dproto.
```
a1batross, line 9: Dproto by default enables fake client detection. I tried to fix it, it just drops player after a while.

Reunion is fine because it doesn't do anything against fake clients, instead it's done in ReAuthCheck, which is okay-ish, though sometimes it also kicks the player anyway.

Btw since we're here, maybe add GoldSrc protocol support to the README.md?

Documentation/nat-bypass-usage.md
```diff
@@ -0,0 +1,32 @@
...
+But this leads to a problem for clients: they cannot accept direct
+incoming connections anymore. This means that if you are behind provider's NAT and will try to setup Xash3D FWGS server - nobody will be able to connect to it,
+and the server will not show up in servers public list.
+
+## Is it possible to avoid this problem?
+In most cases, it is possible to bypass NAT with UDP hole punching, and Xash3D FWGS uses this method too. But this method is not 100% guaranteed to work - it depends
+on NAT configurations on both server and client side, and there is no way to control it.
+
+First of all, server should not be behind symmetric NAT. You can check your NAT type on [this page](https://www.checkmynat.com/).
+If you get "Symmetric NAT" result in this test, that means you cannot setup publically available server with internet connection that you are using.
```
a1batross, line 14: publicly

Documentation/nat-bypass-usage.md
```diff
@@ -0,0 +1,32 @@
...
+| Restricted Cone      | ✔️        | ✔️              | ✔️                    | ✔️        |
+| Port Restricted Cone | ✔️        | ✔️              | ✔️                    | ❌        |
+| Symmetric            | ✔️        | ✔️              | ❌                   | ❌        |
+
+## How to use NAT bypass feature?
+If you are starting server within the game, you need to enable option "*Use NAT Bypass instead of direct mode*" in bottom left corner of screen.
+
+If you are starting dedicated server, you should add console variable `sv_nat 1` to server.cfg file, or add `+sv_nat 1` to server startup parameters.
+
+If you need to connect to server behind NAT, there is separate tab named *NAT* for such servers in server browser. Note that it is useless to add NAT server to favorites,
```
a1batross, line 31: Well, mainui shouldn't allow adding NAT servers to favorites anyway. But it's good to mention the technical reason behind this anyway.

Documentation/entity-tools.md
```diff
@@ -1,60 +1,68 @@
-# There are few new commands availiable in xash3d fork:
+# Entity tools
+
+For some features described below, you need to enable [console scripting](extensions/console-scripting.md) with command `cmd_scripting 1` in console.
```
a1batross, line 3: I wonder if `entity-tools.md` should be in extensions folder too. This feature is kinda unique to us.

Meanwhile, the root `Documentation` folder is getting a bit cluttered. I will work on this later.

Documentation/entity-tools.md
```diff
@@ -117,34 +124,29 @@ alias ent_last_cb "ent_fire \$ent_last_inst use"
...
 ```
 if $ent_last_name == my_ent1_$name
 :(ent1 actions)
 if $ent_last_name == my_ent2_$name
 :(ent2 actions)
 ```
 Note that scripting cannot be blocking. You cannot wait for server answer and continue. But you can use small scripts, connected with ent_last_cb command. The best usage is user interaction. You can add touch buttons to screen or call user command menu actions by callbacks.
-##  Server side
+
+## Configurating on server side
```
a1batross, line 148: Maybe `Server side configuration`?

Documentation/entity-tools.md
a1batross: If you're gonna decide to move it, keep it as a symlink, or just make a new file with a link, in case if anybody using the direct URL somewhere

Documentation/goldsrc-protocol-support.md
SNMetamorph: > Btw since we're here, maybe add GoldSrc protocol support to the README.md?

Maybe I misundertood, but it's already there.

PR #1947, opened 2025-01-07
filesystem: massively rework how scanning game directories work

a1batross:
* No more conversion from liblist.gam to gameinfo.txt. We are using liblist.gam directly now. `gameinfo.txt` being native format to Xash3D takes priority over `liblist.gam`.
* Quake game directories now don't receive autogenerated `gameinfo.txt`.
* Empty directories don't get `gameinfo.txt` either, finally making it easier to support HD addon folders.
* If the user still wishes to generate gameinfo.txt, there is now a command `fs_make_gameinfo` that creates `gameinfo.txt` for the currently running game.
* No more creating empty folders for RoDir. They are now created on demand.


TheKingFireS:
is `commandargs "-example 1234 -example example"` supported? It can be seen in Cry of Fear's `liblist.gam`.

a1batross:
No, this is an unofficial extension. I think it's used by its custom launcher, not by the engine.

PR #1972, opened 2025-01-22
Documentation: Improve mod installation instructions

catskull:
~Also improved uplink setup and used a better source for mod files.~

### DIFF
--- a/Documentation/supported-mod-list.md
+++ b/Documentation/supported-mod-list.md
@@ -142,7 +142,7 @@ For mappacks - place *.bsp files to **valve/maps** folder, *.wad files to **valv
 90. [Half-Life C.A.G.E.D.](https://store.steampowered.com/app/679990/HalfLife_Caged/) (set **-num edicts 4096** in command-line parameters for successful start)
 90. [Half-Life: Dreamcast v1.1](https://www.moddb.com/mods/half-life-dreamcast)
 91. [Half-Life: OPS](https://www.runthinkshootlive.com/posts/half-life-ops/) 
-92. [Half-Life: Uplink](https://www.halflifeuplink.com) (download [hlulsl-2.1.zip](https://www.halflifeuplink.com/pub/hlulsl/hlulsl-2.1.zip), extract, rename `src` to `uplink` and move to base mod directory)
+92. [Half-Life: Uplink](https://www.fileplanet.com/10360/10000/fileinfo/Half-Life:-UpLink-Demo)
 92. [Half-Quake](https://www.moddb.com/mods/halfquake-amen)
 93. [Half-Quake 2: Amen](https://www.moddb.com/mods/halfquake-amen)
 94. [Hard 2](https://www.runthinkshootlive.com/posts/hard-2/) (set **fps_max** to **60** to prevent a possible sticking of Barney on an elevator in the beginning of the mod)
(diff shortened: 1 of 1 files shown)


Documentation/supported-mod-list.md
```diff
@@ -135,7 +142,7 @@ For mappacks - place *.bsp files to **valve/maps** folder, *.wad files to **valv
 90. [Half-Life C.A.G.E.D.](https://store.steampowered.com/app/679990/HalfLife_Caged/) (set **-num edicts 4096** in command-line parameters for successful start)
 90. [Half-Life: Dreamcast v1.1](https://www.moddb.com/mods/half-life-dreamcast)
 91. [Half-Life: OPS](https://www.runthinkshootlive.com/posts/half-life-ops/) 
-92. [Half-Life: Uplink](https://www.fileplanet.com/10360/10000/fileinfo/Half-Life:-UpLink-Demo)
+92. [Half-Life: Uplink](https://www.halflifeuplink.com) (download [hlulsl-2.1.zip](https://www.halflifeuplink.com/pub/hlulsl/hlulsl-2.1.zip), extract, rename `src` to `uplink` and move to base mod directory)
```
nekonomicon, line 145: For what purpose?
There already 4 links for different versions of this mod as you can see:
https://github.com/FWGS/xash3d-fwgs/blob/master/Documentation/supported-mod-list.md?plain=1#L273
https://github.com/FWGS/xash3d-fwgs/blob/master/Documentation/supported-mod-list.md?plain=1#L274

Also, the latest steam version already contains uplink files from further data.

nekonomicon (changes requested):

Documentation/supported-mod-list.md
a1batross: I don't think it really make sense to list `uplink` here at all, since it's shipped with the Steam version of Half-Life now. And we even check for `hldemo1.bsp` existence to show the checkbox in the menu now.

Other than that, clarifying how to run mods (once again) is fine.

Documentation/supported-mod-list.md
```diff
@@ -39,7 +39,14 @@
 # List of Half-Life-based mods
 Originally this list was written by *Qwertyus*(*Qortez*)
 
-To run mods from this category - place mod folder to folder where is **valve** folder located and launch xash3d fwgs with command-line parameter `-game <modfolder>`.
+To install these mods:
+- place mod folder in the same directory as your **valve** folder is located
```
a1batross, line 43: Capitalize the first letter on the list.

Plus, it could be turned from bulleted to numbered list.

Documentation/supported-mod-list.md
nekonomicon: I think need to add separate file in Documentation with instruction how to load first Uplink map in 25th update.

This list originally was written for [old engine](https://github.com/FWGS/xash3d/wiki/List-of-mods-which-work-on-Android-and-other-non-Windows-platforms-without-troubles) using mod list by @Qwertyus3D from [moddb](https://www.moddb.com/engines/xash3d-engine/features/list-of-tested-half-life-singleplayer-mods).
Currently, need to fix broken links somehow and syncronize with fresh @Qwertyus3D's lists anyway.
And just actualize this list for new engine.

Documentation/supported-mod-list.md
nekonomicon: An old instruction was universal for all platforms, but new instruction looks more for PC.
`Custom game` menu is not always available anyway.
I think "To install and run these mods" will be more correct here.

Documentation/supported-mod-list.md
catskull: I undid this change.

nekonomicon (approved):

PR #1980, opened 2025-01-25
Experimental: UTF-8 multilingual font rendering using FreeType

YuanBaian:
It is already possible to implement drawing, but the width and height retrieval needs to be redesigned. The current version contains an unknown bug.

Note: Since I am not familiar with the source code, any path-related parts need to be modified:

You need to manually modify "engine/wscript[line:153]" to specify the folder where "freetype.lib" is stored.
You need to manually modify "cl_font.c[line:120]" to specify the path of the TTF font file.


nekonomicon:
For what purpose you need to upload freetype headers?

nekonomicon:
> Note: Since I am not familiar with the source code, any path-related parts need to be modified:

You can check how it was made in mainui_cpp.

YuanBaian:
Because I don't have enough experience to implement it in multiple ways, I hope to quickly obtain support for UTF8🥺

YuanBaian:
> > Note: Since I am not familiar with the source code, any path-related parts need to be modified:
> 
> 
> 
> You can check how it was made in mainui_cpp.

Because I don't have enough experience to implement it in multiple ways, I hope to quickly obtain support for UTF8🥺

a1batross:
Sorry for long reply.

For now, it definitely works as proof of concept, but code cleanup is definitely required. There is no need to include FreeType headers, nor hardcode library paths.

Also some stuff better be ported from mainui_cpp's text renderer, it at least supports outline effect required for console font.

Instead of hardcoding fonts paths you can try to load ones included in extras.pk3. It's Wine's Tahoma font and Fira Sans: https://github.com/FWGS/xash-extras/tree/master/gfx/fonts. Not sure how they support Chinese though.

PR #1984, opened 2025-01-27
RefAPI 10, hot fix for ref_vk

a1batross:
Now R_EndFrame accepts argument that tells whether engine filled screen with valid data to prevent garbage on screen. But still lets renderer to cleanup it's state.


w23:
This turns out to be a rather inconvenient fix for Vulkan. It's not trivial to just "do nothing" under a boolean flag, as vk needs to still update a bunch of internal state.

The current workaround for multiple begin-frame-calls seems to work just fine, given the assumption that there are no rendering calls in-between them.

Arguably better solution would be not to call the begin-end frame if there's nothing to draw. But it's not really required now, given the reasonable workaround we have now.

a1batross:
@w23 shall this be closed then?

w23:
> @w23 shall this be closed then?

In this particular form, yes.

PR #1991, opened 2025-02-01
Multilingual font rendering

YuanBaian:
﻿
﻿put your ttf file into "3rdparty\extras\xash-extras\gfx\fonts\"

change cl_font.c
FONT_COUNT
g_FontsPath
g_FontsSize

### DIFF
--- a/ref/gl/gl2_shim/gl2_shim.c
+++ b/ref/gl/gl2_shim/gl2_shim.c
@@ -192,7 +192,7 @@ static void APIENTRY GL2_BindTexture( GLenum tex, GLuint obj )
 static char *GL_PrintInfoLog( GLhandleARB object, qboolean program )
 {
 	static char	msg[8192];
-	int		maxLength = 0;
+	GLuint maxLength = 0;
 
 	if( program && pglGetProgramiv )
 		pglGetProgramiv( object, GL_OBJECT_INFO_LOG_LENGTH_ARB, &maxLength );
@@ -282,8 +282,8 @@ static GLuint GL2_GenerateShader( gl2wrap_prog_t *prog, GLenum type )
 
 static gl2wrap_prog_t *GL2_GetProg( const GLuint flags )
 {
-	int i, loc, status;
-	GLuint vp, fp, glprog;
+	int i, loc;
+	GLuint status = 0, vp, fp, glprog;
 	gl2wrap_prog_t *prog;
 
 	// try to find existing prog matching this feature set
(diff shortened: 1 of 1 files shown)


engine/client/console.c
```diff
@@ -619,17 +619,71 @@ Convert utf char to current font's single-byte encoding
 */
 int Con_UtfProcessCharForce( int in )
 {
-	// TODO: get rid of global state where possible
```
a1batross, line 622: Why you did that?

YuanBaian: This can accurately process characters, as previous functions were unable to handle them correctly

YuanBaian: Sorry, I haven't read Q-DecodeUTF8 function. If Con-UtfProcessCharForce is not modified, the Chinese will become "?"

a1batross: Q_DecodeUTF8 must be able to decode Chinese symbols (as long as they're encoded in UTF-8). If it doesn't, it's a bug in Q_DecodeUTF8, and **not** in this function.

engine/client/cl_font.c
```diff
@@ -278,8 +301,11 @@ int CL_DrawString( float x, float y, const char *s, rgba_t color, cl_font_t *fon
 			continue;
 		}
 
+		int ch = Con_UtfProcessChar( (unsigned char)*s );
+
 		// skip setting rendermode, it was changed for this string already
-		draw_len += CL_DrawCharacter( x + draw_len, y, (byte)*s, current_color, font, flags | FONT_DRAW_NORENDERMODE );
+		if( ch )
```
a1batross, line 307: You also absolutely don't have to do this. 

If you don't try to override CL_DrawCharacter with your own function (i.e. Font_DrawChar) it will actually decode UTF-8 for you.

engine/client/console.c
YuanBaian: image
image
After this setting, Chinese can be displayed normally

a1batross: I think I have a better solution for this problem.

a1batross: Nevermind, setting it up to zero is absolutely fine for what we're trying to achieve here.

This code will be removed once engine will completely draw text on it's own, without relying on provided bitmap fonts (which are often in CP1251 or CP1252, thus the logic)

engine/client/cl_font.c
YuanBaian: I don't think this is a good approach. When I ported this part of the code to Font_DrawChar, it didn't produce the correct results. Additionally, CL_DrawCharacter is often called independently, and performing Con_UtfProcessChar operations within it would likely introduce performance overhead
image

YuanBaian: image
The number parameter in CL_DrawCharacter may receive a Chinese character. If CL_DrawCharacter is not called from CL_DrawString, it may fail to display Chinese characters

engine/client/console.c
YuanBaian: I believe the core issue now is whether the implementation of my code is correct. Currently, multilingual rendering in the engine does not appear to be inherently difficult. Let’s prioritize finalizing this feature as soon as possible   XD

engine/client/cl_font.c
YuanBaian: image
cs16-client called CL_DrawCharacter

engine/client/console.c
a1batross: There is absolutely no need to hurry. I didn't even start on the thing, that your code doesn't even compile.

engine/client/cl_font.c
a1batross: That's a valid concern actually.

PR #1993, opened 2025-02-02
Multilingual font rendering

YuanBaian:
put your ttf file into "3rdparty\extras\xash-extras\gfx\fonts"
Change "cl_font.c" file (FONT_COUNT, g_FontsPath, g_FontsSize) to add your ttf font

### DIFF
--- a/engine/client/cl_font.c
+++ b/engine/client/cl_font.c
@@ -17,6 +17,17 @@ GNU General Public License for more details.
 #include "filesystem.h"
 #include "client.h"
 #include "qfont.h"
+#define STB_TRUETYPE_IMPLEMENTATION
+#define STBTT_STATIC
+#include "cl_font.h"
+
+#define FONT_COUNT 2
+
+static Font* g_currentFont;
+static Font g_Fonts[FONT_COUNT];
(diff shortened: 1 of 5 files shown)


engine/client/cl_font.c
```diff
@@ -378,3 +409,178 @@ void CL_DrawStringLen( cl_font_t *font, const char *s, int *width, int *height,
...
+	char texname[30];
+
+	int iCallBack = Font_CheckCharExists(self, ch);
+	if (iCallBack > -1) return iCallBack;
+
+	buf = stbtt_GetCodepointBitmap( &self->m_FontInfo, self->scale, self->scale, ch, &bm_width, &bm_height, &bm_left, &bm_top );
+
+	Q_snprintf( texname, sizeof( texname ), "font_%i", Font_GetTexIndex(self, ch));
+
+	if (!(id = ref.dllFuncs.GL_FindTexture(texname)))
```
a1batross, line 473: Something that would be really nice to have, before this gets merged, is to load the characters in the atlas texture, like mainui_cpp, to eliminate texture switching. 

Maybe allocate 1024x1024 texture and fill it with characters as needed, and once it gets filled up, allocate another large texture.

And then cache the result, something I did in `mainui_cpp` because drawing glyphs takes a lot of time.

PR #2008, opened 2025-02-07
Fix multimonitor issues with SDL backend

OlegAckbar:
This is my attempt on fixing multimonitor issues with SDL backend. Instead of setting 0 xpos and ypos we let platform to decide where to put our window. Instead of setting displayIndex at 0 we get display index from our window (to enumerate display modes properly).
It works just fine for me on Linux with `sdl2-compat` but I haven't tested it on another platforms.

### DIFF
--- a/ref/gl/gl_local.h
+++ b/ref/gl/gl_local.h
@@ -388,7 +388,7 @@ void R_TextureReplacementReport( const char *modelname, int gl_texturenum, const
 void CL_RunLightStyles( lightstyle_t *ls );
 void R_PushDlights( void );
 void R_GetLightSpot( vec3_t lightspot );
-void R_MarkLights( dlight_t *light, int bit, mnode_t *node );
+void R_MarkLights( const dlight_t *light, int bit, const mnode_t *node );
 colorVec R_LightVec( const vec3_t start, const vec3_t end, vec3_t lightspot, vec3_t lightvec );
 colorVec R_LightPoint( const vec3_t p0 );
 
@@ -812,6 +812,7 @@ extern convar_t r_ripple;
 extern convar_t r_ripple_updatetime;
 extern convar_t r_ripple_spawntime;
 extern convar_t r_large_lightmaps;
+extern convar_t r_dlight_virtual_radius;
 
 //
 // engine shared convars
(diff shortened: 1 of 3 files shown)


engine/platform/sdl/vid_sdl.c
```diff
@@ -708,7 +712,7 @@ static qboolean VID_CreateWindowWithSafeGL( const char *wndname, int xpos, int y
 	while( glw_state.safe >= SAFE_NO && glw_state.safe < SAFE_LAST )
 	{
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
-		host.hWnd = SDL_CreateWindow( wndname, xpos, ypos, w, h, flags );
+		host.hWnd = SDL_CreateWindow( wndname, SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED, w, h, flags );
```
a1batross, line 715: You don't need to do this here because `xpos` and `ypos` are passed from `VID_CreateWindow`.

By default, it's SDL_WINDOWPOS_UNDEFINED, but tries to take the window position saved in the config.

engine/platform/sdl/vid_sdl.c
```diff
@@ -1093,8 +1097,12 @@ rserr_t R_ChangeDisplaySettings( int width, int height, window_mode_t window_mod
 {
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
 	SDL_DisplayMode displayMode;
+	int displayIndex = SDL_GetWindowDisplayIndex(host.hWnd);
```
a1batross, line 1100: `host.hWnd` might be NULL here, which seems to be fine for SDL and it will return -1 in this case.

However, is `0` the desired value here?

Similarly to how engine remembers it's window position, why not try to deduce the display index from that? There seems to be `SDL_GetPointDisplayIndex` function for that.

engine/platform/sdl/vid_sdl.c
a1batross, line 715: You don't need to do this here because xpos and ypos are passed from VID_CreateWindow.

By default, it's SDL_WINDOWPOS_UNDEFINED, but tries to take the window position saved in the config.

engine/platform/sdl/vid_sdl.c
```diff
@@ -257,9 +257,13 @@ static void R_InitVideoModes( void )
 {
 	char buf[MAX_VA_STRING];
 #if SDL_VERSION_ATLEAST( 2, 0, 0 )
-	int displayIndex = 0; // TODO: handle multiple displays somehow
+	int displayIndex = SDL_GetWindowDisplayIndex(host.hWnd);
 	int i, modes;
 
+    if (displayIndex < 0) {
```
a1batross, line 263: Please reformat your code.

Opening bracket must be on new line, and there should be spaces in parentheses.

There is provided Uncrustify config in the repository root for that.

ref/gl/gl_local.h
```diff
@@ -388,7 +388,7 @@ void R_TextureReplacementReport( const char *modelname, int gl_texturenum, const
 void CL_RunLightStyles( lightstyle_t *ls );
 void R_PushDlights( void );
 void R_GetLightSpot( vec3_t lightspot );
-void R_MarkLights( dlight_t *light, int bit, mnode_t *node );
+void R_MarkLights( const dlight_t *light, int bit, const mnode_t *node );
```
a1batross, line 391: Sorry for that, I had to fix the commit description and force-pushed the changes.

Cherry-pick your latest commit on top of current `master`.

a1batross (changes requested):

PR #2014, opened 2025-02-09
ci: upgrade freebsd tasks.

nekonomicon:


a1batross:
Oh, nice, they package it!

I will merge this after I will add check for system-provided `libbacktrace`.

nekonomicon:
Oh... there no dynamic linking yet...

> Oh, nice, they package it!

Yep, because FreeBSD does not have glibc.

a1batross:
> Oh... there no dynamic linking yet...
> 
> > Oh, nice, they package it!
> 
> Yep, because FreeBSD does not have glibc.

The `backtrace` from glibc (which is in `execinfo.h`) is not `libbacktrace` by Ian Lance Taylor. The latter actually can print file and line from debug info.

Because we don't link to zlib, I didn't enable compression right away, it's either we should switch to `zlib` or add support for `miniz` in `libbacktrace`. And again I quite like `miniz`, so I'm looking how to add support for it, which shouldn't be complicated as `miniz` provides compatible interface.

a1batross:
I think we should prefer `libbacktrace` if it turns out to be stable enough to be used in signal handler.

a1batross:
Had to manually apply it because it seems you commited changes to `xash-extras` and `mainui`.

Merged, thanks!

PR #2019, opened 2025-02-11
engine: platform: introduce for Platform_NanoSleep, to be used for better sleeping in between frames for lowering CPU usage

a1batross:
There were reports about high CPU usage with high framerates, especially on the servers.

As it seems, it might've been caused by insufficient accuracy of `Platform_Sleep`, which only accepts milliseconds. If we imagine game running at 500+ FPS, `Platform_Sleep` instantly becomes useless.

With this PR, `sleeptime` cvar resolution changes from milliseconds down to microseconds.

### DIFF
--- a/engine/common/host.c
+++ b/engine/common/host.c
@@ -698,17 +698,17 @@ static qboolean Host_Autosleep( double dt, double scale )
 		static double timewindow; // allocate a time window for sleeps
 		static int counter; // for debug
 		static double realsleeptime;
-		const double sleeptime = sleep * 0.001;
+		const double sleeptime = sleep * 0.000001;
 
 		if( dt < targetframetime * scale )
 		{
 			// if we have allocated time window, try to sleep
 			if( timewindow > realsleeptime )
 			{
-				// Platform_Sleep isn't guaranteed to sleep an exact amount of milliseconds
+				// Platform_Sleep isn't guaranteed to sleep an exact amount of microseconds
 				// so we measure the real sleep time and use it to decrease the window
 				double t1 = Sys_DoubleTime(), t2;
-				Platform_Sleep( sleep ); // in msec!
+				Platform_NanoSleep( sleep * 1000 ); // in usec!
 				t2 = Sys_DoubleTime();
 				realsleeptime = t2 - t1;
 
--- a/engine/common/net_ws.c
+++ b/engine/common/net_ws.c
@@ -1512,7 +1512,7 @@ static int NET_SendLong( netsrc_t sock, int net_socket, const char *buf, size_t 
 				total_sent += size;
 			len -= size;
 			packet_number++;
-			Platform_Sleep( 1 );
+			Platform_NanoSleep( 100 * 1000 );
 		}
 
 		return total_sent;
--- a/engine/platform/platform.h
+++ b/engine/platform/platform.h
@@ -70,8 +70,9 @@ void Android_Shutdown( void );
 #endif
 
 #if XASH_WIN32
-void Win32_Init( void );
+void Win32_Init( qboolean con_showalways );
 void Win32_Shutdown( void );
+qboolean Win32_NanoSleep( int nsec );
 void Wcon_CreateConsole( qboolean con_showalways );
 void Wcon_DestroyConsole( void );
 void Wcon_InitConsoleCommands( void );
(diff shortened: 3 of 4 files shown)


a1batross:
Not sure if Windows version of nanosleep will even work. `SetWaitableTimer` wants resolution of 100 ns, which is fine by us.

SNMetamorph:
Interesting stuff according to this topic:
https://blog.bearcats.nl/perfect-sleep-function/

a1batross:
On my Linux server, that expectedly helped me to reduce CPU load from 100% to 20-ish% with `sys_ticrate` set to 1000.

SNMetamorph:
> On my Linux server, that expectedly helped me to reduce CPU load from 100% to 20-ish% with `sys_ticrate` set to 1000.

Pretty good result. This is very important in case when hosting more than 1 server on the VPS/VDS.

a1batross:
> > On my Linux server, that expectedly helped me to reduce CPU load from 100% to 20-ish% with `sys_ticrate` set to 1000.
> 
> Pretty good result. This is very important in case when hosting more than 1 server on the VPS/VDS.

What should we do on Windows though, is there a reliable way to at least have a microsecond timer? That will also help decrease load on the CPU.

SNMetamorph:
> > > On my Linux server, that expectedly helped me to reduce CPU load from 100% to 20-ish% with `sys_ticrate` set to 1000.
> > 
> > 
> > Pretty good result. This is very important in case when hosting more than 1 server on the VPS/VDS.
> 
> What should we do on Windows though, is there a reliable way to at least have a microsecond timer? That will also help decrease load on the CPU.

CreateWaitableTimerEx with `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` flag, provides accuracy something like 1-5 nanoseconds according to benchmarks. But it's main downside that it's available only since Windows 10, version 1803 (cheers for anybody who still uses Windows XP in 2025). So we need to somehow check Windows version in runtime to decide should we use this flag or shouldn't.

a1batross:
> > > > On my Linux server, that expectedly helped me to reduce CPU load from 100% to 20-ish% with `sys_ticrate` set to 1000.
> > > 
> > > 
> > > Pretty good result. This is very important in case when hosting more than 1 server on the VPS/VDS.
> > 
> > 
> > What should we do on Windows though, is there a reliable way to at least have a microsecond timer? That will also help decrease load on the CPU.
> 
> CreateWaitableTimerEx with `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` flag, provides accuracy something like 1-5 nanoseconds according to benchmarks. But it's main downside that it's available only since Windows 10, version 1803 (cheers for anybody who still uses Windows XP in 2025). So we need to somehow check Windows version in runtime to decide should we use this flag or shouldn't.

Right now, I'm getting CreateWaitableTimerEx with GetProcAddress and passing `3` as flags, which equals to that high resolution flag + auto reset.

I think it should be fine for those who use Windows XP or 7.

SNMetamorph:
For Windows versions older that mentioned, we can fallback to `timeBeginPeriod`. It provides decent accuracy, but their main flaw is that it affects OS scheduler to tick more frequently, and consequently it affects other running programs too. Sometimes this can lead to noticable worse battery usage on laptops.

SNMetamorph:
I'm not sure about this.

```cpp
#include <windows.h>
#include <chrono>
#include <math.h>

void timerSleep(double seconds) {
    using namespace std::chrono;

    static HANDLE timer = CreateWaitableTimer(NULL, FALSE, NULL);
    static double estimate = 5e-3;
    static double mean = 5e-3;
    static double m2 = 0;
    static int64_t count = 1;

    while (seconds - estimate > 1e-7) {
        double toWait = seconds - estimate;
        LARGE_INTEGER due;
        due.QuadPart = -int64_t(toWait * 1e7);
        auto start = high_resolution_clock::now();
        SetWaitableTimerEx(timer, &due, 0, NULL, NULL, NULL, 0);
        WaitForSingleObject(timer, INFINITE);
        auto end = high_resolution_clock::now();

        double observed = (end - start).count() / 1e9;
        seconds -= observed;

        ++count;
        double error = observed - toWait;
        double delta = error - mean;
        mean += delta / count;
        m2   += delta * (error - mean);
        double stddev = sqrt(m2 / (count - 1));
        estimate = mean + stddev;
    }

    // spin lock
    auto start = high_resolution_clock::now();
    auto spinNs = int64_t(seconds * 1e9);
    auto delay = nanoseconds(spinNs);
    while (high_resolution_clock::now() - start < delay);
}
```

изображение

SNMetamorph:
Ah, I'm wrong here. This benchmark uses spin-lock for providing better accuracy, so it isn't relevant here :)

a1batross:
It turns out Uncle Mike already uses waitable timers in his XashNT engine, and it seems to be working fine for him.

PR #2048, opened 2025-02-23
engine: client: enable half-texel hack also for drawing text touch buttons

Velaron:


Velaron:
nvm, it kinda actually works only 80% of the time

PR #2050, opened 2025-02-24
ExternalProject for CMake build for Android

a1batross:
This will probably resolve all issues with too complex Android build process.

So far, it gets installed into the correct folder and get packaged into APK, but binaries built with Waf are dependent on `libc++_shared` for some reason, while ones that were using CMake doesn't. This tells me that propagating flags from CMake to Waf is important here.

However, instead, I'm actually thinking of just re-using Android NDK support in `xcompile.py` here.

What I also don't like, is that the build directory is too far away from the source, and the compiler generates long relative paths here, i.e. what was `../public/build.c` becomes `../../../../../public/build.c`. I can try using `file-prefix-map` compiler flag here, but not sure if this will break debugging or not. It's something I will work on before this will get merged.

cc @Velaron

### DIFF
--- a/android/app/CMakeLists.txt
+++ b/android/app/CMakeLists.txt
@@ -5,13 +5,13 @@ cmake_minimum_required(VERSION 3.6)
 project(XASH_ANDROID)
 
 # armeabi-v7a requires cpufeatures library
-include(AndroidNdkModules)
-android_ndk_import_module_cpufeatures()
+if(ANDROID)
+    include_directories(${ANDROID_NDK}/sources/android/cpufeatures)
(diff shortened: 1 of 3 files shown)


lvonasek:
FYI: This PR breaks debugging in Android Studio. CSVR will probably use codebase before the PR got merged as having the debugger is essential for problems solving. I tried to revert this but then I went into other issues.

a1batross:
Create a bugreport. It should work, and it works on Linux machine.

сб, 14 июн. 2025 г., 23:20 Luboš V. ***@***.***>:

PR #2063, opened 2025-03-01
engine: client: lower the bandwidth test fragment sizes to 1400 and less, as using 64k, 32k and more is unreliable

### LINKED ISSUE #2060: Server connection issues on 49 protocol servers
The issue was on the bandwidth test. When I tried connecting the server with a larger fragment size (64000, 32000, 10666, 5200), it always stuck on the console, and nothing happened. When I set it to 1400 fragment size only, it can join the server quickly and never stuck.

Image

a1batross:
Fixes #2060

Until we don't have a better way to determine path MTU, just lower the bandwidth test values.

### DIFF
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -1183,7 +1183,11 @@ Returns bandwidth test fragment size
 */
 static int CL_GetTestFragmentSize( void )
 {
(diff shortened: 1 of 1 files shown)


a1batross:
@xiaodo1337 can you check and verify it helps you?

xiaodo1337:
Yes, I've been testing it for 2 days, and it works on all servers except the OLDSCHOOL CSDM.

PR #2075, opened 2025-03-06
Unique HWID implementation for PS Vita platform

SNMetamorph:


a1batross:
@fgsfdsfgs could you verify on your device that this syscall really returns unique IDs?

SNMetamorph:
@fgsfdsfgs could you please validate that with this patch engine generates unique ID in `ux0:data/xash3d/value/.xashid` file? Mine starts with `77FC` and ends with `11F5`. In your case this should be different value, and same ID should persists even after deleting `.xashid` file and starting engine again.

PR #2101, opened 2025-03-26
engine: client: fix TE_BLOODSPRITE behaviour to GoldSrc

yumirak:
Right now when you land a headshot, enemy will launch blood so high since blood height are scaled with damage.
This is different from goldsrc.

Below tested using M3 in CSCZ for worst case scenario.

Before:
before2

After:
after

GoldSrc:
goldsrc


a1batross:
Looks good to me. Thanks!

PR #2113, opened 2025-04-01
Add experimental modern TUI main menu

numas13:
Every day more and more users are switching to TUI apps. It brings many advantages over traditional GUIs, such as reduced memory usage, reduced CPU usage, and improved navigation with a keyboard. **TUI is the future of UX/UI**.

Today I propose to add TUI-based main menu to prepare the engine for the future.

This PR adds a TUI-based main menu that can be enabled with `--enable-tui` option.

main-new-game

The menu is written in Rust language to prevent dangerous vulnerabilities such as remote code execution (RCE) with malicious server and map names.


a1batross:
Good, merging!

PR #2117, opened 2025-04-02
Fix nightly builds for PS Vita

SNMetamorph:
Because modern Ubuntu versions provides CMake 4.x from apt, and this version breaks a lot of stuff.


a1batross:
Installing exact version might work for a while, but it might disappear from the repos, if some minor patch comes out.

a1batross:
Let's keep it the way it is for now, in future we could just fix psvita SDL branch to build with CMake 4.0.

PR #2131, opened 2025-04-15
New Emscripten port by @okhmanyuk-ev

a1batross:
This is a very specific piece of code that's made to run on web clients and Linux servers.

It cannot be merged in the current state (you can probably see that by diffstat) however for preserving and for possible upstream Emscripten support reasons it's now here.

Was taken from https://github.com/okhmanyuk-ev/customized-xash.


nekonomicon:
unordered_map?
Noise :(

PR #2172, opened 2025-05-22
ref_api: R_GetWindowHandle method

TheEVolk:
### **Added `R_GetWindowHandle` to `ref_api`**  
- **Purpose**: Allows renderers to fetch the native window handle for direct integration (e.g., Vulkan/D3D/Metal).  
- **Returns**: Opaque pointer to the platform-specific handle (`HWND`, `NSWindow*`, `X11 Window`, etc.).  
- **Unsupported platforms**: Returns `NULL`
- **Supported platforms**: SDL2 only now (WIN32, X11, WAYLAND, MACOS)

```c 
// Usage example (renderer-side):  
window_handle_t window_handle = ref_api->R_GetWindowHandle();  
if (window_handle.type == 0) {  
    Com_Error(ERR_FATAL, "Window handle not supported on this platform");  
}  
// ... create surface/swapchain with the handle  
```

### DIFF
--- a/engine/platform/dos/vid_dos.c
+++ b/engine/platform/dos/vid_dos.c
@@ -543,12 +543,9 @@ qboolean SW_CreateBuffer( int width, int height, uint *stride, uint *bpp, uint *
 	return true;
 }
 
-window_handle_t R_GetWindowHandle( void )
+qboolean R_GetWindowHandle( void **handle, int type )
 {
-	window_handle_t handle;
-	handle.type = REF_WINDOW_TYPE_NONE;
-	handle.handle = NULL;
-	return handle;
+	return FALSE;
 }
 
 #endif
--- a/engine/platform/dos/vid_dos.c
+++ b/engine/platform/dos/vid_dos.c
@@ -543,9 +543,12 @@ qboolean SW_CreateBuffer( int width, int height, uint *stride, uint *bpp, uint *
 	return true;
 }
 
-window_handle_t *R_GetWindowHandle( void )
+window_handle_t R_GetWindowHandle( void )
 {
-	return NULL;
+	window_handle_t handle;
+	handle.type = REF_WINDOW_TYPE_NONE;
+	handle.handle = NULL;
+	return handle;
 }
 
 #endif
--- a/engine/platform/dos/vid_dos.c
+++ b/engine/platform/dos/vid_dos.c
@@ -543,4 +543,9 @@ qboolean SW_CreateBuffer( int width, int height, uint *stride, uint *bpp, uint *
 	return true;
 }
 
+window_handle_t *R_GetWindowHandle( void )
+{
+	return NULL;
+}
+
 #endif
--- a/engine/platform/platform.h
+++ b/engine/platform/platform.h
@@ -312,21 +312,6 @@ typedef enum
 	rserr_unknown
 } rserr_t;
 
-enum
-{
-	REF_WINDOW_TYPE_NONE = 0, // NULL
-	REF_WINDOW_TYPE_WIN32 = 1, // HWND
-	REF_WINDOW_TYPE_X11 = 2, // Display*
-	REF_WINDOW_TYPE_WAYLAND = 3, // wl_display*
-	REF_WINDOW_TYPE_MACOS = 4, // NSWindow*
-};
-
-typedef struct window_handle_s
-{
-	int type;
-	void* handle;
-} window_handle_t;
-
 struct vidmode_s;
 typedef enum window_mode_e window_mode_t;
 // Window
@@ -345,7 +330,7 @@ void *SW_LockBuffer( void );
 void SW_UnlockBuffer( void );
 qboolean SW_CreateBuffer( int width, int height, uint *stride, uint *bpp, uint *r, uint *g, uint *b );
 void Platform_Minimize_f( void );
-window_handle_t R_GetWindowHandle( void );
+qboolean R_GetWindowHandle( void **handle, int type );
 
 //
 // in_evdev.c
--- a/engine/platform/platform.h
+++ b/engine/platform/platform.h
@@ -312,6 +312,20 @@ typedef enum
 	rserr_unknown
 } rserr_t;
 
+enum
+{
+	REF_WINDOW_TYPE_WIN32 = 1, // HWND
+	REF_WINDOW_TYPE_X11 = 2, // Display*
+	REF_WINDOW_TYPE_WAYLAND = 3, // wl_display*
+	REF_WINDOW_TYPE_MACOS = 4, // NSWindow*
+};
+
+typedef struct window_handle_s
+{
+	int type;
+	void* handle;
+} window_handle_t;
+
 struct vidmode_s;
 typedef enum window_mode_e window_mode_t;
 // Window
@@ -330,6 +344,7 @@ void *SW_LockBuffer( void );
 void SW_UnlockBuffer( void );
 qboolean SW_CreateBuffer( int width, int height, uint *stride, uint *bpp, uint *r, uint *g, uint *b );
 void Platform_Minimize_f( void );
+window_handle_t *R_GetWindowHandle( void );
 
 //
 // in_evdev.c
--- a/engine/platform/platform.h
+++ b/engine/platform/platform.h
@@ -314,6 +314,7 @@ typedef enum
 
 enum
 {
+	REF_WINDOW_TYPE_NONE = 0, // NULL
 	REF_WINDOW_TYPE_WIN32 = 1, // HWND
 	REF_WINDOW_TYPE_X11 = 2, // Display*
 	REF_WINDOW_TYPE_WAYLAND = 3, // wl_display*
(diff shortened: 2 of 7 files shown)


TheEVolk:
I slightly revised my decision regarding memory allocation. Now the service is always returned, and you can check it already by the pointer inside it, or by type.

mittorn:
Why even allocate heap pointer, just pass struct pointer from renderer to engine
Also, do you even need struct?
Pass enum from engine and return void* handle or null if different window system used.
Also you may add some generic handles like SDL_Window in that case, which may be useful to implement ref over SDL APIs
If you need return struct, pass pointer from renderer

engine/platform/sdl2/vid_sdl2.c
```diff
@@ -1164,6 +1165,38 @@ qboolean VID_SetMode( void )
...
+	handle.handle = NULL;
+
+	SDL_SysWMinfo wmInfo;
+    SDL_VERSION(&wmInfo.version);
+
+    if (SDL_GetWindowWMInfo(host.hWnd, &wmInfo) != SDL_TRUE) {
+        return handle;
+    }
+
+    #if defined(SDL_VIDEO_DRIVER_WINDOWS)
```
a1batross, line 1181: This isn't correct way to get handle to the window.

SDL2 provides SDL_SYSWM_TYPE enumeration to get currently used window subsystem: https://wiki.libsdl.org/SDL2/SDL_SYSWM_TYPE

Also, hiding them under macro this way isn't right either.

I suggest you changing this to something like:
```
switch( wmInfo.subsystem )
{
case FOO:
#if defined( SDL_VIDEO_DRIVER_FOO )
	handle.type = REF_WINDOW_TYPE_FOO;
	handle.handle = wmInfo.info.foo.window;
#endif
	break;
case BAR...   
...
}
```

Also, I think what if function signature would be changed to:
`qboolean R_GetWindowHandle( window_handle_t *handle )`

For example, this way we can return non-zero if no error happened and still use NULL and REF_WINDOW_TYPE_NULL. And return zero if it's completely not supported by the platform-specific code or if SDL was misconfigured somehow.

engine/platform/platform.h
```diff
@@ -312,6 +312,21 @@ typedef enum
 	rserr_unknown
 } rserr_t;
 
+enum
```
a1batross, line 315: Move enumeration and structure to the `ref_api.h`.

a1batross (changes requested):

engine/platform/sdl2/vid_sdl2.c
a1batross: Also, as @mittorn said above, it might be useful to request specifically SDL_Window pointer.

SDL3 for example adds SDL3_gpu which I believe is initialized from SDL_Window pointer.

a1batross: This is also something that a better function signature should take care of. However it also feels dirty if SDL specifics would go into RefAPI.

a1batross:
Also, do you have any new renderer in development that needs this export?

Sure, ref_vk guys (hi @w23) might be interested in it, but don't they already have some Vulkan-related extensions to RefAPI?

SNMetamorph:
> Also, do you have any new renderer in development that needs this export?

He is doing new renderer based on WGPU

mittorn:
>window_handle_t R_GetWindowHandle( void )
I do not think aggregate return is good solution.
maybe something like:
void * R_GetWindowHandle(enum handle_type type)
or qboolean R_GetWindowHandle(struct window_handle_s *out, enum handle_type 
type)

This allows request exact window type if it's currently supported. So you can 
select SDL_Window for sdlrenderer/sdlgpu ref implementations

Also, i'm not sure struct return is useful in other languages like Rust

>Sure, ref_vk guys (hi @w23) might be interested in it, but don't they already 
have some Vulkan-related extensions to RefAPI?

ref_vk is using Vulkan types in their API. Vulkan is complex and creating 
context on engine side will require pass many information from renderer. 
Vulkan behaviour is hardly depended on enabled version and extensions, 
selected only when creating instance/device.
I think, moving context creation to ref is preferred in that case.
Also, OpenXR graphics bindings will need to be implemented in ref too (because 
openxr vulkan v2 wraps context/device create calls), and keeping all internal 
vulkan/opengl/d3d/metal data inside ref looks more consistent. Engine only 
provides platform egl/wgl/glx abstraction, so ref does not need 
platform-specific data and engine does not need graphics api specific data

TheEVolk:
> Also, do you have any new renderer in development that needs this export?
> 
> Sure, ref_vk guys (hi @w23) might be interested in it, but don't they already have some Vulkan-related extensions to RefAPI?

It's true, the guys from vulkan extended the method and made XVK_CreateSurface.  My renderer uses a wgpu that is abstracted from the backend.

TheEVolk:
> I was thinking about either a separate method for this, or making the
> useAbstractHandle flag or something like that Why adding some strange flags when you already have enum for that? Just ask which exact handle you would to get. Other way is passing some set/array of expected handles, but i do not think 3-4 GetHandle calls on swapchain initialization affect performance too much

So be it, we pass the type of the desired handle. The engine will either return null or a pointer to the handle. As for whether the method will accept a pointer or return it, I will implement this method that you will like. I'm waiting for your feedback.

a1batross:
So far I like `qboolean R_GetWindowHandle(struct window_handle_s *out, enum handle_type
type)` more.

You can request specific handle type. It really does make sense for renderer that expects very specific kind of handle (would it be `drm_fd`, `HWND` or `SDL_Window`), and for engine to fill the struct by pointer. And simple return type still can be used to report an error.

engine/platform/dos/vid_dos.c
```diff
@@ -543,4 +543,9 @@ qboolean SW_CreateBuffer( int width, int height, uint *stride, uint *bpp, uint *
 	return true;
 }
 
+qboolean R_GetWindowHandle( void **handle, int type )
+{
+	return FALSE;
```
a1batross, line 548: Oh, you can use just `false` in engine and renderer code. Not sure where `FALSE` could've came from.

engine/platform/sdl2/vid_sdl2.c
```diff
@@ -1164,6 +1165,54 @@ qboolean VID_SetMode( void )
...
+{
+	SDL_SysWMinfo wmInfo;
+	SDL_VERSION( &wmInfo.version );
+
+	if ( SDL_GetWindowWMInfo( host.hWnd, &wmInfo ) != SDL_TRUE )
+	{
+		return FALSE;
+	}
+
+	switch( type )
```
a1batross, line 1178: Again, you can not be sure that requested window handle type matches whatever SDL used to create the window.

I think the correct logic is like this:

```
switch( type )
{
case REF_WINDOW_TYPE_WIN32:
#ifdef SDL_VIDEO_DRIVER_WINDOWS
	if( wmInfo.info.subsystem == SDL_SYSWM_WINDOWS )
	{
		*handle = (void*)wmInfo.info.win.window;
		return true;
	}
	else
		return false;
#else
	return false;
#endif
```

Also `case` doesn't indent, it stays on the same level as `switch`.

engine/ref_api.h
```diff
@@ -57,6 +57,7 @@ GNU General Public License for more details.
 //    CL_RunLightStyles now accepts lightstyles array.
 //    Removed R_DrawTileClear and Mod_LoadMapSprite, as they're implemented on engine side
 //    Removed FillRGBABlend. Now FillRGBA accepts rendermode parameter.
+// 10. Added R_GetWindowHandle.
 #define REF_API_VERSION 9
```
a1batross, line 61: You either add to the changelog _and_ bump the version, or don't mention that in the changelog at all, don't bump the version and just add your changes to the file.

Changelog in file is only for migration from older RefAPI versions, everything else could be and is handled by Git.

engine/platform/sdl2/vid_sdl2.c
a1batross: Also, on some platforms like Linux it might make sense to query window type FIRST and only then fetch handle.

Imagine the situation when we don't know if X11 or Wayland is used, though renderer supports both of them.

Returning a `struct` with handle _and_ window type was a good idea (I don't know why you decided to go with the void-pointer though), this way renderer might not ask for specific window type.

engine/ref_api.h
```diff
@@ -103,6 +104,15 @@ typedef enum
 	DEMO_QUAKE1
 } demo_mode;
 
+enum
+{
+	REF_WINDOW_TYPE_WIN32 = 1, // HWND
```
a1batross, line 109: There is no point to add numbers here. :)

Also it would be nice if you gave this enum a type and used it instead of abstract `int`.

a1batross (changes requested):

engine/platform/dos/vid_dos.c
a1batross: Grepping engine source code for `FALSE` and `TRUE` sounds like yet another Win32 API retardness.

We don't do that here.

PR #2194, opened 2025-06-08
Clean-room reverse engineered SDK headers, part 1

### LINKED ISSUE #63: License issues
I really do like the idea of a GPL gldsrc clone, however after some research I've discovered a bit of an issue.

The engine is somewhat notorious for making use of leaked source code, and indeed Valve as supposedly remarked that the engine as it stands is illegal (though I have yet to personally corroborate that.) This is troublesome because I want to use xash3D for a commercial project, so as it currently stands that dream is unworkable. But what if it wasn't?

I turned to the codebase and I've picked out all the obvious files that violate Valve's copyright:

- common/weaponinfo.h
- common/beamdef.h
- common/entity_state.h
- common/net_api.h
- common/r_studioint.h
- common/demo_api.h
- common/r_efx.h
- common/kbutton.h
- common/const.h
- common/netadr.h
- common/con_nprint.h
- common/usercmd.h
- common/studio_event.h
- common/pmtrace.h
- common/cvardef.h
- common/ref_params.h
- common/wadfile.h
- common/screenfade.h
- common/event_args.h
- common/mathlib.h
- common/qfont.h
- common/triangleapi.h
- common/cl_entity.h
- common/event_flags.h
- common/event_api.h
- common/dlight.h
- common/hltv.h
- common/particledef.h
- common/entity_types.h
- common/ivoicetweak.h
- engine/alias.h
- engine/studio.h
- engine/keydefs.h
- engine/customentity.h
- engine/anorms.h
- engine/custom.h
- engine/edict.h
- engine/cdll_int.h
- engine/progdefs.h
- engine/shake.h
- engine/sprite.h
- engine/eiface.h
- pm_shared/pm_defs.h
- pm_shared/pm_movevars.h
- pm_shared/pm_info.h

I'm not sure about any snippets of code that might violate Valve's copyright, so I'll perform a more indepth analysis to check if any exist. I doubt much if any such snippets exist, so barring that; If we rewrite these headers, the legal and licensing issues associated with this project should be resolved and full GPL compliance will be achieved.

freed00m21:
Partially resolves #63


common/beamdef.h
```diff
@@ -1,60 +1,83 @@
...
-	float		frameRate;
-	float		frame;
-	int		segments;
-	int		startEntity;
-	int		endEntity;
-	int		modelIndex;
-	int		frameCount;
-	struct model_s	*pFollowModel;
-	struct particle_s	*particles;
+    FBEAM_STARTENTITY = 1 << 0,
```
nekonomicon, line 35: We have such macro:
https://github.com/FWGS/xash3d-fwgs/blob/master/common/xash3d_types.h#L51

nekonomicon (changes requested):

SNMetamorph:
Nice work! Though, it's very helpful in terms of licensing issues

a1batross:
I'm not sure that clean room means generate data types from DWARF. But it's better than nothing.

I would add static assertions on struct sizes though, at least for ILP32 model (aka 32-bit). It does not guarantee compatibility, but again better than nothing.

common/dlight.h
```diff
@@ -1,31 +1,53 @@
...
+*/
 
 #ifndef DLIGHT_H
 #define DLIGHT_H
+#include "xash3d_types.h"
+#include "const.h"
+
+typedef struct dlight_s dlight_t;
+
+struct dlight_s {
```
a1batross, line 35: As far as I remember, compatible dlight struct can be taken from Quake.

freed00m21:
I made changes, is anything else required?

a1batross:
As it builds now, I should check it that nothing potentially got broken, and then merge it. Thanks.

a1batross:
After all, I'm quite suspicious if clean-room definition applies to the generated code from DWARF debugging information.

Not only that, it's also weird that preprocessor variables only got refactored into the enumeration, retaining same names. Preprocessing is only the first stage of compiling C code and while _some_ of it information gets preserved in debugging info, like included files, the defines unfortunately do not and I don't know any other sources of information where they could be taken from.

SNMetamorph:
@freed00m21 i need to clarify some things, so I have some questions for you:
1. Have you even been familiar with Xash3D or HLSDK source code before doing this pull request?
2. In your code we found things that couldn't be obtained directly from binaries, as some bit flags, like `FBEAM_*`.
Where did you get these data? It's ok as long as it's obtained from some another clean-room reverse-engineering project, so we need to know exact source of that data to be completely sure.

PR #2212, opened 2025-07-01
GoldSrc voice chat

rootcreep:
## Voice Packet Structure

| Offset (bytes) | Length (bytes) | Field Name           | Description                                                    |
|----------------|---------------|----------------------|----------------------------------------------------------------|
| 0              | 8             | SteamID              | Unique user identifier (Steam)                                 |
| 8              | 1             | VPC_SETSAMPLERATE    | Command type: always 11                                        |
| 9              | 2             | Sample Rate          | Sample rate, Little Endian (e.g., 24000)                       |
| 11             | 1             | VPC_VDATA_OPUS_PLC   | Command type: [see code](https://github.com/FWGS/xash3d-fwgs/blob/cca1732245a6d90a13a14b05859d4132d3fe60ab/engine/client/voice.h#L54)                     |
| 12             | 2             | Data Length          | Length of the following data (Opus or Silence), Little Endian  |
| 14             | N             | Data                 | Opus data (N bytes) or silence (N samples)                     |
| 14+N           | 4             | CRC32                | CRC32 checksum of all previous bytes in the packet             |

**Notes:**
- If **VPC_VDATA_OPUS_PLC** (6), the following N bytes are Opus-encoded audio data.
- If **VPC_VDATA_SILENCE** (0), the following N samples are silence (usually just zeros).
- Sample Rate is almost always 24000.
- CRC32 is calculated over all bytes except the last 4 (the CRC32 itself).

## Opus Payload Structure

The Opus data (payload) inside a voice packet consists of a sequence of Opus frames, each with the following format:

| Offset (from start of Opus payload) | Length (bytes) | Field Name   | Description                                                        |
|-------------------------------------|---------------|--------------|--------------------------------------------------------------------|
| 0                                   | 2             | frame_size   | Size of the Opus frame in bytes (uint16, little-endian)            |
| 2                                   | 2             | sequence     | Sequence number/counter (uint16, little-endian), not used in code  |
| 4                                   | N             | frame_data   | Opus-encoded data, N bytes                                         |

Then the next frame follows (if any), and so on until the end of the Opus payload.

### Special values for frame_size:
- **0** — means "silence" (add 160 zero samples to the output buffer).
- **0xFFFF** — Opus decoder reset command; after this, parsing of the Opus payload stops.

### Example

| Offset | Length | Value    | Description                  |
|--------|--------|----------|------------------------------|
| 0      | 2      | 0x0032   | frame_size = 50 bytes        |
| 2      | 2      | 0x0001   | sequence = 1                 |
| 4      | 50     | ...      | Opus data (50 bytes)         |
| 54     | 2      | 0x0000   | frame_size = 0 (silence)     |
| 56     | 2      | 0x0002   | sequence = 2                 |
| 58     | 0      | -        | no data, just 160 zeros      |

### DIFF
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -1029,6 +1029,10 @@ static void CL_WriteSteamTicket( sizebuf_t *send )
 	crc = CRC32_Final( crc );
 	i = GenerateRevEmu2013( buf, s, crc );
 	MSG_WriteBytes( send, buf, i );
+
+	// RevEmu2013: pTicket[1] = revHash (low), pTicket[5] = 0x01100001 (high)
+	*(uint32_t*)cls.steamid = LittleLong( ((uint32_t*)buf)[1] );
+	*(uint32_t*)(cls.steamid + 4) = LittleLong( ((uint32_t*)buf)[5] );
 }
 
 /*
--- a/engine/client/cl_parse.c
+++ b/engine/client/cl_parse.c
@@ -2011,7 +2011,7 @@ CL_ParseVoiceData
 void CL_ParseVoiceData( sizebuf_t *msg, connprotocol_t proto )
 {
 	int size, idx, frames = 0;
-	byte received[8192];
+	byte received[VOICE_MAX_DATA_SIZE];
 
 	idx = MSG_ReadByte( msg ) + 1;
 
@@ -2021,7 +2021,7 @@ void CL_ParseVoiceData( sizebuf_t *msg, connprotocol_t proto )
 	if( proto == PROTO_GOLDSRC )
 	{
 		size = MSG_ReadShort( msg );
-		if ( size > 4096 )
+		if ( size > VOICE_MAX_GS_DATA_SIZE )
 		{
 			Con_Printf( S_ERROR "Voice data size is too large: %d bytes (max: %d)\n", size, VOICE_MAX_GS_DATA_SIZE );
 			return;
@@ -2031,7 +2031,7 @@ void CL_ParseVoiceData( sizebuf_t *msg, connprotocol_t proto )
 	{
 		frames = MSG_ReadByte( msg );
 		size = MSG_ReadShort( msg );
-		if (size > 8192 )
+		if (size > VOICE_MAX_DATA_SIZE )
 		{
 			Con_Printf( S_ERROR "Voice data size is too large: %d bytes (max: %d)\n", size, VOICE_MAX_DATA_SIZE );
 			return;
--- a/engine/client/cl_parse.c
+++ b/engine/client/cl_parse.c
@@ -2010,43 +2010,40 @@ CL_ParseVoiceData
 */
 void CL_ParseVoiceData( sizebuf_t *msg, connprotocol_t proto )
 {
-	int size, idx, frames;
+	int size, idx, frames = 0;
 	byte received[8192];
 
 	idx = MSG_ReadByte( msg ) + 1;
 
-	if( proto == PROTO_GOLDSRC )
-	{
-		size = MSG_ReadShort( msg );
-		MSG_SeekToBit( msg, size << 3, SEEK_CUR ); // skip the entire buf, not supported yet
-
-#if 0 // shall we notify client.dll if nothing can be heard?
-		// must notify through as both local player and normal client
-		if( idx == cl.playernum + 1 )
-			Voice_StatusAck( &voice.local, VOICE_LOOPBACK_INDEX );
-
-		Voice_StatusAck( &voice.players_status[idx], idx );
-#endif
-		return;
-	}
-
-	frames = MSG_ReadByte( msg );
-	size = MSG_ReadShort( msg );
-	size = Q_min( size, sizeof( received ));
-
-	MSG_ReadBytes( msg, received, size );
-
 	if ( idx <= 0 || idx > cl.maxclients )
 		return;
 
-	// must notify through as both local player and normal client
-	if( idx == cl.playernum + 1 )
-		Voice_StatusAck( &voice.local, VOICE_LOOPBACK_INDEX );
+	if( proto == PROTO_GOLDSRC )
+	{
+		size = MSG_ReadShort( msg );
+		if ( size > 4096 )
+		{
+			Con_Printf( S_ERROR "Voice data size is too large: %d bytes (max: %d)\n", size, VOICE_MAX_GS_DATA_SIZE );
+			return;
+		}
+	}
+	else
+	{
+		frames = MSG_ReadByte( msg );
+		size = MSG_ReadShort( msg );
+		if (size > 8192 )
+		{
+			Con_Printf( S_ERROR "Voice data size is too large: %d bytes (max: %d)\n", size, VOICE_MAX_DATA_SIZE );
+			return;
+		}
+	}
 
-	Voice_StatusAck( &voice.players_status[idx], idx );
+	size = Q_min( size, sizeof( received ));
 
 	if ( !size )
 		return;
+
+	MSG_ReadBytes( msg, received, size );
 
 	Voice_AddIncomingData( idx, received, size, frames );
 }
--- a/engine/client/client.h
+++ b/engine/client/client.h
@@ -638,6 +638,7 @@ typedef struct
 
 	// server's build number (might be zero)
 	int build_num;
+	uint8_t steamid[8];
 } client_static_t;
 
 #ifdef __cplusplus
--- a/engine/client/voice.c
+++ b/engine/client/voice.c
@@ -509,7 +509,7 @@ Get compressed voice data for GoldSrc mode
 */
 static uint Voice_GetGSCompressedData( byte *out, uint maxsize, uint *frames )
 {
-	uint ofs, size = 0;
+	uint ofs = 0, size = 0;
 	const uint      frame_size_samples = GS_DEFAULT_FRAME_SIZE;
 	const uint      frame_size_bytes = GS_DEFAULT_FRAME_SIZE * voice.width;
 	static uint16_t sequence = 0;
@@ -1124,7 +1124,11 @@ static void Voice_Shutdown( void )
 	voice.initialized = false;
 	voice.is_recording = false;
 	voice.device_opened = false;
+	voice.goldsrc = false;
 	voice.start_time = 0.0;
+	voice.samplerate = 0;
+	voice.frame_size = 0;
+	voice.width = 0;
 
 	voice.input_buffer_pos = 0;
 	voice.input_file_pos = 0;
@@ -1206,13 +1210,6 @@ qboolean Voice_Init( const char *pszCodecName, int quality, qboolean preinit )
 		// unsupported codec
 		Con_Printf( S_WARN "Server requested unsupported voice codec: %s\n", pszCodecName );
 		Voice_Shutdown();
-		voice.goldsrc = false;
-		voice.codec[0] = 0;
-		voice.quality = 0;
-		voice.samplerate = 0;
-		voice.frame_size = 0;
-		voice.width = 0;
-		voice.initialized = false;
 		return false;
 	}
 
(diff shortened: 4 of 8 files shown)


engine/client/cl_main.c
```diff
@@ -1029,6 +1029,10 @@ static void CL_WriteSteamTicket( sizebuf_t *send )
 	crc = CRC32_Final( crc );
 	i = GenerateRevEmu2013( buf, s, crc );
 	MSG_WriteBytes( send, buf, i );
+
+	// RevEmu2013: pTicket[1] = revHash (low), pTicket[5] = 0x01100001 (high)
+	*(uint32_t*)cls.steamid = LittleLong( ((uint32_t*)buf)[1] );
+	*(uint32_t*)(cls.steamid + 4) = LittleLong( ((uint32_t*)buf)[5] );
```
a1batross, line 1035: In future, this might be better redone to make MultiEmulator return it's generated steamid as an argument, rather filling the buffer and then parsing it back. But it's fine for now...

engine/client/cl_parse.c
```diff
@@ -2010,44 +2010,41 @@ CL_ParseVoiceData
...
-
-#if 0 // shall we notify client.dll if nothing can be heard?
-		// must notify through as both local player and normal client
-		if( idx == cl.playernum + 1 )
-			Voice_StatusAck( &voice.local, VOICE_LOOPBACK_INDEX );
-
-		Voice_StatusAck( &voice.players_status[idx], idx );
-#endif
-		return;
+		if ( size > 4096 )
```
a1batross:
```c
if( size > VOICE_MAX_GS_DATA_SIZE )
```

engine/client/cl_parse.c
```diff
@@ -2010,44 +2010,41 @@ CL_ParseVoiceData
...
+		{
+			Con_Printf( S_ERROR "Voice data size is too large: %d bytes (max: %d)\n", size, VOICE_MAX_GS_DATA_SIZE );
+			return;
+		}
+	}
+	else
+	{
+		frames = MSG_ReadByte( msg );
+		size = MSG_ReadShort( msg );
+		if (size > 8192 )
```
a1batross: Similarly,
```c
if( size > VOICE_MAX_DATA_SIZE )
```

engine/client/voice.c
```diff
@@ -31,11 +33,84 @@ CVAR_DEFINE_AUTO( voice_maxgain, "5.0", FCVAR_PRIVILEGED|FCVAR_ARCHIVE, "automat
...
+*/
+
+/*
+=========================
+Voice_IsGoldSrcMode
+
+Check if codec is GoldSrc mode
+=========================
+*/
+qboolean Voice_IsGoldSrcMode( const char *codec )
```
a1batross: If this function is defined before everything else, it's better to remove declaration above.

From this:
```c
static qboolean Voice_IsGoldSrcMode( const char *codec );
...
qboolean Voice_IsGoldSrcMode( const char *codec ) { ... }
```

To this:
```c
static qboolean Voice_IsGoldSrcMode( const char *codec ) { ... }
```

Make sure to do the same with other functions as well.

engine/client/voice.c
```diff
@@ -131,35 +223,66 @@ static qboolean Voice_InitOpusEncoder( int quality )
...
 	for( int i = 0; i < MAX_CLIENTS; i++ )
 	{
-		if( !voice.decoders[i] )
-			continue;
-
-		opus_custom_decoder_destroy( voice.decoders[i] );
-		voice.decoders[i] = NULL;
+		if( voice.goldsrc )
+		{
+			if( voice.gs_decoders[i] )
```
a1batross, line 226: Please reformat code, we don't use spaces for indentation.

a1batross: There is uncrustify config at the root of the repository to help you with that, btw

engine/client/voice.c
```diff
@@ -249,63 +510,274 @@ static uint Voice_GetOpusCompressedData( byte *out, uint maxsize, uint *frames )
...
+
+    *frames = 0;
+
+    while (voice.input_buffer_pos - ofs >= frame_size_bytes)
+    {
+        int bytes;
+
+        int is_silence = 1;
+        int16_t *samples = (int16_t *)(voice.input_buffer + ofs);
+        for (uint i = 0; i < frame_size_samples; ++i) {
```
a1batross: Again, you can just reformat whole file using uncrustify. It's very easy to use from console. Grab the _latest_ version and run:

`uncrustify -c uncrustify.cfg --replace --no-backup engine/client/voice.c`

engine/client/voice.c
```diff
@@ -249,63 +510,274 @@ static uint Voice_GetOpusCompressedData( byte *out, uint maxsize, uint *frames )
...
+    uint16_t sample_rate;
+    uint8_t vpc_type;
+    uint16_t data_len;
+    OpusDecoder *decoder;
+    size_t opus_offset;
+    int decoded;
+    size_t silence_samples;
+    uint16_t frame_size;
+
+    if (!data || size < 18 || size < 4 || ent <= 0 || ent > cl.maxclients)
```
a1batross: What's the point of `size < 18 || size < 4`?

At first, there shouldn't be any magic numbers without the explanation.

At second, `size < 4` implies `size < 18`.

engine/client/voice.c
```diff
@@ -249,63 +510,274 @@ static uint Voice_GetOpusCompressedData( byte *out, uint maxsize, uint *frames )
...
+        Con_Printf( S_WARN "Invalid voice packet type: %d\n", data[offset] );
+        return 0;
+    }
+    offset++;
+
+    if (offset + 4 > size - 4)
+        return 0;
+
+    sample_rate = LittleShort(*(uint16_t*)(data + offset));
+    offset += 2;
```
a1batross: Use `sizeof( uint16_t )` instead.

engine/client/voice.c
```diff
@@ -249,63 +510,274 @@ static uint Voice_GetOpusCompressedData( byte *out, uint maxsize, uint *frames )
...
+        if (is_silence) {
+            ofs += frame_size_bytes;
+            continue;
+        }
+
+		if ( !voice.input_file )
+			Voice_ApplyGainAdjust(samples, frame_size_samples, voice_transmit_scale.value);
+
+        bytes = opus_encode( voice.gs_encoder, samples,
+            frame_size_samples, out + size + 4, maxsize - 4 );
```
a1batross: Same thing about magic numbers here.

If `4` represents 32-bit number, use `sizeof( uint32_t )`. If it represents two separate 16-bit values, use `sizeof( uint16_t ) * 2` to keep the code readable and maintainable.

engine/client/voice.c
```diff
@@ -249,63 +510,274 @@ static uint Voice_GetOpusCompressedData( byte *out, uint maxsize, uint *frames )
...
+            Con_Printf( S_WARN "No decoder available for entity %d\n", ent );
+            return 0;
+        }
+        opus_offset = 0;
+        while (opus_offset + 4 <= data_len) {
+            frame_size = LittleShort(*(uint16_t*)(data + offset + opus_offset));
+            opus_offset += 4;
+
+            if (frame_size == 0) {
+                if (output_samples + 160 > GS_MAX_DECOMPRESSED_SAMPLES) {
```
a1batross: What does `160` is supposed to mean here?

engine/client/voice.c
```diff
@@ -249,63 +510,274 @@ static uint Voice_GetOpusCompressedData( byte *out, uint maxsize, uint *frames )
...
+		memcpy(out + offset, cls.steamid, 8);
+	} else {
+		memset(out + offset, 0, 8); // fallback: 0
+	}
 
-		for( i = 0; i < blockSize; ++i )
-		{
-			int sample = samples[blockOffset + i];
-			int absSample = abs( sample );
+	offset += 8;
```
a1batross: Again, use `sizeof` instead of magic numbers.

SteamID here is 64-bit number, so `sizeof( uint64_t )` or just `sizeof( cls.steamid )` could've done the job.

engine/client/client.h
```diff
@@ -638,6 +638,7 @@ typedef struct
 
 	// server's build number (might be zero)
 	int build_num;
+	uint8_t steamid[8];
```
a1batross, line 641: Actually, I'm not sure if `steamid` should be in `cls`. `cls` state is kept between connections, whereas `cl` is specific for current connection. I see I did the same mistake with `build_num` above, but still, I think it's connection-specific actually.

engine/client/voice.c
```diff
@@ -585,14 +1108,26 @@ static void Voice_Shutdown( void )
 			Voice_Status( i, false );
 	}
 
-	memset( &voice, 0, sizeof( voice ));
+	voice.initialized = false;
```
a1batross, line 1124: Why memset is gone?

engine/client/voice.c
```diff
@@ -605,15 +1140,19 @@ void Voice_Idle( double frametime )
 
 		if( voice_enable.value )
 		{
-			if( cls.state == ca_active && COM_CheckString( voice_codec_init ) && voice_quality_init != 0 )
-				Voice_Init( voice_codec_init, voice_quality_init, false );
+			// if( cls.state == ca_active && COM_CheckString( voice.codec ) && voice.quality != 0 )
```
a1batross: Do not comment out code, we use Git for that.

If the check has been moved, you can just remove it instead of commenting it out.

engine/client/voice.c
```diff
@@ -605,15 +1140,19 @@ void Voice_Idle( double frametime )
...
+				Voice_Init( voice.codec, voice.quality, false );
 		}
 		else
 			Voice_Shutdown();
 	}
 
 	// update local player status first
-	Voice_StatusTimeout( &voice.local, VOICE_LOOPBACK_INDEX, frametime );
+	if ( voice.goldsrc )
+		Voice_StatusTimeout( &voice.local, cl.playernum + 1, frametime );
```
a1batross: What so special about goldsrc protocol when notifying client.dll?

engine/client/voice.c
```diff
@@ -628,67 +1167,53 @@ Initialize the voice subsystem
...
-			Con_Printf( S_ERROR "Can't create decoders, voice chat is disabled.\n" );
 			Voice_Shutdown();
 			return false;
 		}
+	}
+	else
+	{
+		// unsupported codec
+		Voice_Shutdown();
+		voice.goldsrc = false;
```
a1batross: Why? If the voice hasn't been initialized, cleaning up the whole structure is much more fool-proof.

a1batross (changes requested):

engine/client/voice.c
Buggem: Following on a1batross, you should probably comment it our just plain remove it

engine/client/voice.c
```diff
@@ -628,67 +1167,53 @@ Initialize the voice subsystem
...
+	{
+		if( !Voice_InitGoldSrcMode( quality ) )
 		{
-			// no reason to init encoder and open audio device
-			// if we can't hear other players
 			Voice_Shutdown();
 			return false;
 		}
-
-		// we can hear others players, so it's fine to fail now
```
Buggem, line 662: Please keep these comments, it's easier for people to understand how this code works with them

engine/client/voice.c
rootcreep: In fact, clearing the entire structure is not entirely reliable, for example, if a player turns off the `voice_enable 0` voice chat while in the game and then turns on `voice_enable 1` again, he will not be able to properly initialize the voice system. Your option used separate variables to save the codec name and sound quality, it looked slightly inappropriate. I tried to redo it to use everything in one structure.

And I agree that selective cleaning of the structure is not exactly what it should be, it may be worth looking for another compromise or returning to your option.

engine/client/voice.c
rootcreep: 160 is the standard pause duration that must be inserted when receiving a silent frame. Maybe it should be to in voice.h

a1batross (approved):

PR #2221, opened 2025-07-09
GoldSrc spray support

pwd491:


engine/client/cl_main.c
```diff
@@ -1317,19 +1318,18 @@ static void CL_CreateResourceList( void )
 	cl.num_resources = 0;
 	memset( rgucMD5_hash, 0, sizeof( rgucMD5_hash ));
 
+	// sanitize cvar value
+	if( Q_strcmp( cl_logoext.string, "bmp" ) && Q_strcmp( cl_logoext.string, "png" ))
```
SNMetamorph, line 1322: Engine supports many other texture formats, rather than just BMP and PNG. Also, it could load TGA, DDS, KTX2 textures.

a1batross: I think that check was added by me, probably got moved from somewhere else. It's fine.

rootcreep: > Engine supports many other texture formats, rather than just BMP and PNG. Also, it could load TGA, DDS, KTX2 textures.

I'm not sure if these formats are actually supported, at least the engine ignores them. Most likely, they are also converted to png/bmp.

common/wadfile.h
```diff
@@ -83,4 +83,22 @@ typedef struct mip_s
 	unsigned int	offsets[4];	// four mip maps stored
 } mip_t;
 
+typedef struct
+{
+	char identification[4]; // "WAD3"
+	int  numlumps;
+	int  infotableofs;
+} wad3header_t;
```
a1batross: These structs are defined in `filesystem/wad.c`. You can just move them from that file to this header.

engine/client/cl_spray.c
```diff
@@ -0,0 +1,442 @@
...
+			FS_FreeImage( tmp );
+		if( scaled_image )
+		{
+			FS_FreeImage( image );
+			image = scaled_image;
+		}
+	}
+
+	// generate gradient palette
+	color = LogoColor_FindByName( logocolor );
```
a1batross: You don't have to do this. If you take `remapped.bmp`, palette can be taken from that file. 

`imagelib` can preserve palette and won't transform indexed image to RGBA with `KEEP_8BIT` flag.

As for `remapped.png`, it's always fullcolor and can be quantized into indexed image (we also have quantizer in `imagelib`).

engine/client/cl_spray.c
```diff
@@ -0,0 +1,442 @@
...
+		scaled_image->size = *width * *height * 4;
+		memcpy( scaled_image->buffer, resampled_buf, *width * *height * 4 );
+		Con_DPrintf( S_GREEN "Created scaled image %dx%d\n", *width, *height );
+	}
+	*scaled_image_out = scaled_image;
+	return image;
+}
+
+// write wad3 file
+static qboolean CL_WriteSprayWAD3( const byte *indexed, int width, int height, const byte *palette, int lump_type )
```
a1batross: Ideally, such things must be implemented in `imagelib`.

Maybe add handler for `.wad` extension in `imagelib` that will load and save tempdecal?

engine/common/net_chan.c
```diff
@@ -1080,6 +1120,10 @@ int Netchan_CreateFileFragments( netchan_t *chan, const char *filename )
 			// Write filename
 			MSG_WriteString( &buf->frag_message, filename );
 
+			// write compressor name and uncompressed size
+			MSG_WriteString( &buf->frag_message, compressor );
```
a1batross: This needs check for gs_netchan or it will break protocol for Xash3D 49.

a1batross (changes requested):

pwd491:
To be honest, I have no idea why the build failed. Since the last build in my fork, I have not changed anything in **net_chan.c**, but for some reason linux 22 cannot find definitions from **bzlib.h** Well, that's funny, because a week ago I successfully compiled the same code. I also checked history from upstream and didn't find anything suspicious.

common/wadfile.h
```diff
@@ -77,10 +90,73 @@ typedef struct lmp_s
...
+} dlumpinfo_t;
+
+/*
+========================================================================
+
+WAD file structure	(Internal WAD handling)
+
+========================================================================
+*/
+struct wfile_s
```
a1batross: `wfile_t` and `wadtype_t` are private to WADs mounting to filesystem. 

The (mostly unspoken Quake heritage) convention is that only structs that start with `d` are used to parse files. `d` here stands for data.

engine/common/custom.c
```diff
@@ -17,10 +17,125 @@ GNU General Public License for more details.
...
+	}
+	return NULL;
+}
 
 static rgbdata_t *CustomDecal_LoadImage( const char *path, void *raw, int size )
 {
 	const char *testname;
+	rgbdata_t *wadimg;
+
+	if( raw && size > 16 && memcmp( raw, "WAD3", 4 ) == 0 )
```
a1batross: The point of loading WAD files as images in `imagelib` is that you can do this:

```c
wadimg = FS_LoadImage( "tempdecal.wad", raw, size );
```

See, when do you pass second and third argument to FS_LoadImage it will actually use provided buffer and parse the image from it. The filename is used as a hint to tell what kind of image is encoded in the buffer. Here it will see `.wad` extension and go straight directly to parsing the buffer as the `wad` that only contains a single image.

engine/common/imagelib/img_wad.c
```diff
@@ -549,3 +549,115 @@ qboolean Image_LoadMIP( const char *name, const byte *buffer, fs_offset_t filesi
...
 }
+
+/*
+============
+Image_LoadWAD
+============
+*/
+qboolean Image_LoadWAD( const char *filename, const char *lumpname, rgbdata_t **out_image )
+{
+	// TODO: implement this plz
```
a1batross: The implementation of it is basically your `CustomDecal_LoadHLWADSpray`.

engine/common/imagelib/imagelib.h
```diff
@@ -157,13 +158,15 @@ qboolean Image_LoadFNT( const char *name, const byte *buffer, fs_offset_t filesi
 qboolean Image_LoadLMP( const char *name, const byte *buffer, fs_offset_t filesize );
 qboolean Image_LoadPAL( const char *name, const byte *buffer, fs_offset_t filesize );
 qboolean Image_LoadKTX2( const char *name, const byte *buffer, fs_offset_t filesize );
+qboolean Image_LoadWAD( const char *name, const char *lumpname, rgbdata_t **out_image );
```
a1batross: Why does the function declaration differs from these above? Same for `SaveWAD`.

a1batross: My idea was to add `wad` handling to the imagelib specifically for `tempdecal.wad` kind of file. It can and should be specifically tailored to loading and saving this kind of WAD file.

common/wadfile.h
a1batross: Same for `// WAD errors` macros above, they only matter in filesystem.

a1batross:
Thanks, it's perfect now, merged.

PR #2227, opened 2025-07-17
android: update port

Velaron:
This is still kinda W.I.P. but I'm gonna put this here:
* build now is using waf instead of cmake (may need some more work)
* the game files are now read from sd card (needs some testing and work)

Planned features:
* downloading game content
* download game libraries from a repository
* app update feature
* dedicated server mode?
* need to set up a server for ACRA
* basedir file picker
* and more...

Suggestions welcome!


README.md
```diff
@@ -84,6 +89,15 @@ This repository contains our fork of HLSDK and restored source code for Half-Lif
 
 * Clone this repostory: `$ git clone --recursive https://github.com/FWGS/xash3d-fwgs`.
 
+#### Android (Windows/Linux/macOS)
+* Install [Android Studio](https://developer.android.com/studio) (or the command line tools).
+* Install [Python](https://python.org) (at least 2.7, latest is better).
```
a1batross: I don't think it really makes sense to mention 2.7.

We support it for legacy systems but on development machine nothing stops from using latest version. I doubt that Android Studio can be ran and used on anything legacy.

android/app/build.gradle.kts
```diff
@@ -1,130 +1,137 @@
...
+		applicationId = "su.xash.engine"
+		versionName = "0.21-" + getGitHash()
+		versionCode = getBuildNum()
+		minSdk = 21
+		targetSdk = 35
+
+		externalNativeBuild {
+			val engineRoot = projectDir.parentFile.parent
+
+			experimentalProperties["ninja.abiFilters"] = setOf("armeabi-v7a", "arm64-v8a")
```
a1batross: Are we dropping x86? About time!

a1batross (approved):
Looks good enough to me.

PR #2233, opened 2025-07-22
[emscripten] use for macro EXPORT __attribute__(( used ))

ololoken:
In general it allows emscripten compiler to keep exports.

### DIFF
--- a/common/xash3d_types.h
+++ b/common/xash3d_types.h
@@ -12,6 +12,10 @@
 #include <wchar.h> // off_t
 #endif // _WIN32
 
+#if XASH_EMSCRIPTEN
+#include <emscripten.h>
+#endif
+
 #include <sys/types.h> // off_t
 #ifdef STDINT_H
 #include STDINT_H
@@ -73,6 +77,9 @@ typedef int qboolean;
 	#if defined( __i386__ )
 		#define EXPORT         __attribute__(( visibility( "default" ), force_align_arg_pointer ))
 		#define GAME_EXPORT    __attribute__(( force_align_arg_pointer ))
+	#elif defined(XASH_EMSCRIPTEN)
+		#define EXPORT EMSCRIPTEN_KEEPALIVE
+		#define GAME_EXPORT
 	#else
 		#define EXPORT         __attribute__(( visibility ( "default" )))
 		#define GAME_EXPORT
--- a/common/xash3d_types.h
+++ b/common/xash3d_types.h
@@ -12,10 +12,6 @@
 #include <wchar.h> // off_t
 #endif // _WIN32
 
-#if XASH_EMSCRIPTEN
-#include <emscripten.h>
-#endif
-
 #include <sys/types.h> // off_t
 #ifdef STDINT_H
 #include STDINT_H
@@ -78,7 +74,7 @@ typedef int qboolean;
 		#define EXPORT         __attribute__(( visibility( "default" ), force_align_arg_pointer ))
 		#define GAME_EXPORT    __attribute__(( force_align_arg_pointer ))
 	#elif defined(XASH_EMSCRIPTEN)
-		#define EXPORT EMSCRIPTEN_KEEPALIVE
+		#define                __attribute__((used)) //tell llvm to keep this export, see EMSCRIPTEN_KEEPALIVE
 		#define GAME_EXPORT
 	#else
 		#define EXPORT         __attribute__(( visibility ( "default" )))


a1batross:
Where does EMSCRIPTEN_KEEPALIVE is supposed to get from? If you add an attribute, you can keep it as an attribute, without hiding it under some other macro.

Please follow contribution requirements in CONTRIBUTING.md.

ololoken:
`EMSCRIPTEN_KEEPALIVE` is equivalent of `__attribute__((used))`.
There is issue with exports in emscripten, everything gets stripped with visibility("default"). Theoretically it can be fixed with `-Wl,-export-dynamic`, but I didn't test it.

a1batross:
That's what I mean, you can use that attribute directly, without the macro, making it so platform-specific emscripten.h doesn't get included globally.

a1batross:
Again, can you read CONTRIBUTING.md?

a1batross:
Why did you close it?

ololoken:
Well some scripts get working, but still there is bug with guard not opening door of wagon before c1a0 map.
For now I don't have ideas how to fix it.
Also I managed to run emscripten build with gl4es renderer https://github.com/ololoken/xash3d-fwgs/pull/1/files#diff-ddfbb6df870ac2fd70eca5fb6df7269fba4278fb3435bae64fe7e7c2da55a5b4R516
Would you give an advice how to avoid these #if#else as I need to use `gl4es_GetProcAddress` and `gEngfuncs.GL_GetProcAddress` returns result of default `SDL_GL_GetProcAddress`?

ololoken:
image
One additional question, what may cause this rendering issue. Texture is flickering when angle of view gets changed.

nekonomicon:
> One additional question, what may cause this rendering issue.

It's probably such named `mapper's bug`, when map author or map compiler makes mistakes.

ololoken:
> It's probably such named `mapper's bug`, when map author or map compiler makes mistakes.

hmm, it appears only with gl4es renderer on any map.

nekonomicon:
It appears with any renderer and even under GoldSource, just not such visible.

nekonomicon:
> Well some scripts get working, but still there is bug with guard not opening door of wagon before c1a0 map.

Nodes generation/loading is working?

ololoken:
> Nodes generation/loading is working?

How to check it? I'm totally new in the project.

nekonomicon:
https://github.com/FWGS/hlsdk-portable/blob/master/dlls/nodes.cpp#L2369

a1batross:
@ololoken I think it's kinda unfair to call us names for simply asking to follow the commit message format we all use, and then come back asking for help.

ololoken:
@a1batross sorry, man. No offence. My apologises. 
I just want to make working things and then prepare good PR. It was my mistake to expect that the fix is so easy.

ololoken:
image
so, I managed to build gl4es without ugly if/else. No more walls texture bug, but now same happens explosions marks.
suspect there is some mess between gl4es_GetProcAddress and SDL_GL_GetProcAddress...
For now I don't understand how pass gl4es_GetProcAddress from gl4es renderer and use in gEngfuncs.GL_GetProcAddress

ololoken:
@nekonomicon 

last logs messages are:
[01:59:44] Firing: (barnwalk)
[01:59:44] Found: scripted_sequence, firing (barnwalk)
[01:59:44] script "barnwalk" using monster "barney1"

ololoken:
@nekonomicon FYI problem was related `dladdr` in emscripten is just stub. I managed to get scripts work with `XASH_ALLOW_SAVERESTORE_OFFSETS=1`.

a1batross:
@ololoken yeah, it's a fallback option for platforms where reverse symbol search isn't possible or not implemented.

Keep in mind though that saves are intended to be crossplatform and using offsets will break that feature. Offsets themselves might also change after recompiling the server library.

As per graphical glitches, there is a logic in GL4ES that decides whether it should use low or high precision in it's generated shaders, and it might got stuck on low precision.

There is a difference between gl4es_GetProcAddress and SDL_GL_GetProcAddress. gl4es_GPA will return wrapped GL 1.x functions, whereas SDL_GL_GPA returns driver functions. GL4ES should retrieve driver function, as it interacts with GLES, and engine should get wrapped functions, as the renderer is designed for GL1.x.

ololoken:
there are two builds
https://turch.in/xash3d-hl-gl4es/index.html - uses gl4es_GPA (GL static linking is off) (bugs: transparency 100% of weapongs; flickering wall textures)
https://turch.in/xash3d-hl-gl2shim/index.html - uses built in gl2shim (bugs: flickering texture of glowing lamps, flickering texture of shot marks)

nothing special, just work in progress, but you can look and maybe "in vivo" it will be easier to locale an issue.

ololoken:
https://turch.in/xash3d-hl/index.html here is demo with fixed bugs
I'm going to check if there is way to implement pfnNameForFunction for emscripten using GOT/wasmImports metadata then will reopen/create new PR

PR #2238, opened 2025-07-25
Fix CL_DrawCenterPrint aligment for cyrillic text

pwd491:
fix https://github.com/Velaron/cs16-client/issues/40
fix https://github.com/Velaron/cs16-client/issues/26

### DIFF
--- a/engine/client/cl_game.c
+++ b/engine/client/cl_game.c
@@ -436,12 +436,15 @@ void CL_DrawCenterPrint( void )
 
 		while( *pText && *pText != '\n' && lineLength < MAX_LINELENGTH )
 		{
-			byte c = *pText;
-			line[lineLength] = c;
-			CL_DrawCharacterLen( font, c, &charWidth, NULL );
+			int number = Con_UtfProcessChar(( byte ) * pText );
+			pText++;
+			if( number == 0 )
+				continue;
+
+			line[lineLength] = number;
+			CL_DrawCharacterLen( font, number, &charWidth, NULL );
 			width += charWidth;
 			lineLength++;
-			pText++;
 		}
 
 		if( lineLength == MAX_LINELENGTH )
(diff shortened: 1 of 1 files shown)


engine/client/cl_game.c
```diff
@@ -436,12 +436,15 @@ void CL_DrawCenterPrint( void )
 
 		while( *pText && *pText != '\n' && lineLength < MAX_LINELENGTH )
 		{
-			byte c = *pText;
-			line[lineLength] = c;
-			CL_DrawCharacterLen( font, c, &charWidth, NULL );
+			int number = Con_UtfProcessChar(( byte ) * pText );
```
a1batross, line 439: It's a legacy function that have a nasty global state. Use utflib.h functions, there is plenty of examples on how to use them correctly.

rootcreep: > It's a legacy function that have a nasty global state. Use utflib.h functions, there is plenty of examples on how to use them correctly.

There are some difficulties when using Q_DecodeUTF8. The function returns a code point of type uint32_t, which is the correct code point in Unicode. However, the font system and CL_DrawCharacter expect a single-byte value (usually in CP1251 or CP1252 encoding), rather than a full Unicode code point. If you pass the code point directly, it will either be truncated or incorrectly displayed, resulting in unreadable characters for Cyrillic.

To display the text correctly, you must explicitly convert the Unicode code point returned by Q_DecodeUTF8 to the corresponding one-byte code page value (using Q_UnicodeToCP1251, Q_UnicodeToCP1252, etc.) before passing it to the display functions. This adds an extra step over the legacy function, which (despite the global state issues) implicitly handled this mapping to match font resources.

In addition, to determine which transformation to use, we need to query the current con_charset value at runtime.

Is it really necessary to use utflib.h now?

a1batross: Yes, I forgot that there is a charset involved, thanks for correcting me.

PR #2253, opened 2025-08-02
Make Android follow library naming scheme

a1batross:
cc @Velaron


Velaron:
I was going to merge the branch myself, locally I have almost fully rewritten the app...

a1batross:
@Velaron I know, I only merged it to get the ninja support. Also, you're really better to send things early here, so they don't get stalled and rewritten before they even have the chance to get cleaned up.

Velaron:
LGTM then

PR #2257, opened 2025-08-04
engine: common: add a host_writeconfig console command

Velaron:


engine/common/host.c
```diff
@@ -914,6 +914,13 @@ static void Host_Userconfigd_f( void )
 	Mem_Free( t );
 }
 
+#if !XASH_DEDICATED
+static void Host_WriteConfig_f( void )
```
a1batross: I don't think you have to add _f function for it, as Host_WriteConfig function matches needed signature to be a command.

Other than that, I'm fine with the changes.

Velaron: Initially I wanted to add a parameter to specify config name, but I'm not sure if that's a good idea and what's the best way to do it

PR #2283, opened 2025-08-17
update strings.xml pt-BR android client

davidmacalister:


a1batross:
Please follow commit message style in the future.

Instead of `update strings.xml pt-BR android client` it's much better to say `android: update Brazilian Portuguese translation`

PR #2303, opened 2025-09-06
Update gl_backend.c

dassarthak18:
Fixes the waf build error:

task in 'ref_gl' failed with exit status 1:
{task 139875678916496: c gl_backend.c -> gl_backend.c.1.o,gl_backend.c.1.d}

error in some older gcc versions due to type mismatch.


a1batross:
Thanks for contribution but while it fixes 64-bit builds, it breaks 32-bit instead.

Superseded by https://github.com/FWGS/xash3d-fwgs/commit/c6790370856896d22cb47509a2456a9cef967306

PR #2305, opened 2025-09-08
engine: voice: fix voice loopback bit

pwd491:


engine/client/cl_main.c
```diff
@@ -810,7 +810,7 @@ static void CL_WritePacket( void )
 		MSG_WriteByte( &buf, 0 );
 
 		if( proto == PROTO_GOLDSRC && voice_loopback.value )
-			SetBits( packet_loss, 7 ); // set 7-th bit to tell server that we want voice loopback
```
SNMetamorph, line 813: The better would be:
`SetBits( packet_loss, BIT( 7 ));`

a1batross: Oh I feel so damn stupid...

Yes, it better be what Metamorph has said but much thanks for the fix.

PR #2307, opened 2025-09-09
ref: soft: implement r_fullbright

LekKit:
This is an old fix of mine to make r_fullbright fork on ref_soft renderer. It seems it is actually still needed upstream, as r_fullbright only affects ent models and not the world.

All other places in the code invalidate light values correctly but without this PR it eventually just reloads lightdata immediately instead.


LekKit:
D'oh, why github would mark it as mergeable and only then tell me it has merge conflict

a1batross:
Thanks!

PR #2310, opened 2025-09-12
android: add Chinese translation

BaiBlowSnow:


a1batross:
No way, Hatsune Miku?? Just kidding

Thank you. LGTM :)

PR #2312, opened 2025-09-15
Android: Add French translation.

malortie:
This adds French translation to Android strings.

While I haven't tested it, I'm sure it'll be ok.

Each sentence has been checked with:

- Google translate (French -> English and English -> French)
- [Reverso](https://www.reverso.net/orthographe/correcteur-francais/) (Grammar check)

**Note:** Some sentences don't match the ones you'd get when translating from English -> French using Google Translate. This is because some words in English and French are written the same but have different meanings.

I attest that each sentence's meaning is accurate and free of grammar errors.


a1batross:
Thank you very much!

PR #2326, opened 2025-10-03
Improved compatibility with GoldSrc-based clients behavior

SNMetamorph:


a1batross:
I actually wonder where these commands are coming from.

If they are generated by GameUI (which we obviously don't have), then it makes sense to send them in GoldSrc protocol compatibility mode.

SNMetamorph:
To be honest, I have no idea xD

a1batross:
@SNMetamorph might make sense to mention original author of this patch if they don't mind it

SNMetamorph:
Yup, I already did this is commit message

Garey27:
> To be honest, I have no idea xD

Me either, i just track commands (server-side) that default cs1.6 client sends during connection state, patched, tested and "dproto" module stops kicking me for "Fake Client".

a1batross:
If it's just CS1.6, then `specmode` might come out from it's client.dll. `hlsdk-portable` and Valve's HL client.dll sends `specmode 1` on connection, in `CHudSpectator::InitHUDData` which is missing in `cs16-client`. 

`unpause` me thinks comes from GameUI.dll.

SNMetamorph:
So yeah, these commands sent from GameUI
https://github.com/CS-NextClient/NextClient/blob/a7f2c4823fa93902831d667bb4b2bd402552c274/nextclient/gameui/src/GameUi/GameUi.cpp

image

a1batross:
Let's merge it the way it is for now, so that everybody could enjoy playing on dproto servers, and figure out the rest later.

a1batross:
Thanks @Garey27 for fix! :3

PR #2330, opened 2025-10-06
Support multiple demo protocols.

Garey27:
### Changes:
* Refactored demo related code.
* Moved Xash3D/Quake (cl_demo.c) demo related code into another dir.
* Implemented playback/record for GoldSRC demos.
* "Private" interface for working with demos.
* Added GoldSRC-related call funcs in multiple functions. If there no function demo interface vtable - does nothing.

### New cvar:
dem_format - Default demo format for writing demos: [xash3d, goldsource] default: "xash3d".

To-Do:
- [ ]  Test
- [ ] Move Quake demo record/playback related code into separate file.

### Comment:
I was played on GoldSRC servers (ReHLDS), and sometimes there mismatch between GoldSRC and Xash3D engine with sound coordinates and i think handling network packets (sometimes i hear enemy player not like GoldSRC in 3d-space so i think "panning and volume" a little different).

So i hope this PR at first will help when debugging some GoldSRC/Xash3D multiplayer-mismatches.


Garey27:
Sorry i messed up wscript, seems like my editor was made auto-format.

engine/client/cl_events.c
```diff
@@ -555,4 +555,7 @@ void GAME_EXPORT CL_PlaybackEvent( int flags, const edict_t *pInvoker, word even
 	args.bparam2 = bparam2;
 
 	CL_QueueEvent( flags, eventindex, delay, &args );
+
+	if (cls.demorecording)
```
a1batross, line 559: Please follow our code style.

There is no space between if and (, but there are spaces inside ( and ), except when they are consecutive, i.e. instead of ) ) ), it's ))).

Your new files can be formatted using Uncrustify tool.

engine/client/cl_main.c
```diff
@@ -606,12 +607,17 @@ static void CL_UpdateClientData( void )
 	cdat.iWeaponBits = cl.local.weapons;
 	cdat.fov = cl.local.scr_fov;
 
+	if (cls.demorecording)
+		memcpy(&oldcdat, &cdat, sizeof(oldcdat));
```
a1batross: In C you can do oldcdat = cdat.

engine/client/demofile/demofile.c
```diff
@@ -0,0 +1,299 @@
...
+static demo_handler_t* demo_handler = NULL;
+
+
+static CVAR_DEFINE_AUTO(dem_format, "xash3d", FCVAR_ARCHIVE | FCVAR_PROTECTED, "Default demo format for writting demos: [xash3d, goldsource]");
+
+
+static demo_handler_t* DEM_GetHandler(const char* format_name) {
+	demo_handler_t* current = demo_handlers;
+	while (current != NULL) {
+		if (strcmp(current->name, format_name) == 0) {
```
a1batross: We use Q_strcmp. 

Most of the engine is caseinsensitivite, so even Q_stricmp might be better option here.

a1batross: In fact, do demo handlers really need to be a linked list?

engine/client/demofile/demofile.c
```diff
@@ -0,0 +1,299 @@
...
+	{
+		demo_handler = DEM_GetHandler(dem_format.string);
+	}
+
+	// Second pass (if invalid handler in cvar string)
+	if (!demo_handler)
+	{
+		Con_Printf("dem_format \"%s\" is not valid, resetting to \"xash3d\"\n", dem_format.string);
+		Con_Printf("Available formats: [xash3d, goldsource]\n");
+		Cvar_Set("dem_format", "xash3d");
```
a1batross: Because the cvar object is visible here, you can use Cvar_DirectSet.

engine/client/cl_events.c
Garey27: Sorry, i forgot read [CONTRIBUTING.md](https://github.com/FWGS/xash3d-fwgs/blob/master/CONTRIBUTING.md).

engine/client/demofile/demofile_gs.c
```diff
@@ -0,0 +1,892 @@
...
+	// output the buffer.
+	FS_Write(cls.demofile, buffer, size);
+}
+
+static double DEM_GS_GetHostFPS(void)
+{
+	return demo.fps;
+}
+
+static demo_handler_t GS_DemoHandler = {
```
a1batross, line 859: I think it's even possible to make it `static const`

engine/client/demofile/demofile_legacy.c
a1batross, line 1: An empty file?

engine/client/demofile/demofile_quake.c
```diff
@@ -0,0 +1,16 @@
...
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+*/
+
+// TODO: Move original quake related code from demofile_xash.c to this file.
```
a1batross: Let's not keep it as an emlty file, just move it the TODO into `demofile_xash.c`.

engine/wscript
```diff
@@ -7,16 +7,16 @@ from waflib.extras import pthread
 import os
 from copy import copy
 
-top = '.'
+top = "."
```
a1batross, line 11: Is there a reason to reformat the wscript? If not, please revert this change to save some time on review process.

engine/client/cl_events.c
a1batross: Yeah, everything in demofile folder can be reformatted automatically. There is a config file in the root of repository, and there is the tool: https://github.com/uncrustify/uncrustify

a1batross:
I kinda like how every demo handler just provide it's own interface, completely private to the file where it's implemented.

Also, never thought we could write GoldSrc demos, only read, so good job on that.

engine/client/demofile/demofile.c
Garey27: I think yes, or in specific order because Quake demo is last one (as i fast inspect quake1 source - it don't have unique header in the file). So it should be "processed" last when all other "readers" rejected to process this file?

Garey27: Also if you wonder about optimization, strcmp (will be Q_strcmp after fix) and generally this function will be called once only (in CL_Playdemo_f) when we don't have handler.

a1batross: I mean, there is a similar problem in imagelib and soundlib and they just use an array. Anyway, yeah, that's more about the preference, I'm just asking here.

Garey27:
Seems like calling CL_WriteDemoStringCmd from Cmd_ExecuteStringWithPrivilegeCheck (as in original GoldSRC) https://github.com/Garey27/xash3d-fwgs/blob/cl_demo_gs/engine/common/cmd.c#L1008 is wrong.

I assume cmd.c also using in server-code? Any clue where to properly record string-command into demo?

Garey27:
Maybe we need some func table interface for working with demo-file by function pointers (Public Demo API)? Also this will help to make headless "HLTV" server-side.

a1batross:
>I assume cmd.c also using in server-code? Any clue where to properly record string-command into demo?

Do you want to write commands that are coming from the server or that's locally executed by the client or both?

In case of server-only commands, put your function call in `svc_stufftext` handler.

a1batross:
>Maybe we need some func table interface for working with demo-file by function pointers (Public Demo API)? Also this will help to make headless "HLTV" server-side.

Let's put this out of scope of this PR for now, as there is no HLTV yet.

Garey27:
> Do you want to write commands that are coming from the server or that's locally executed by the client or both?

Original goldsrc record's only cliendll cmds like +jump/+attack etc (FBitSet(cmd->flags, CMD_CLIENTDLL)), mainly used for demochecking by speedrunners communities like "Kreedz". So only client one.

Garey27:
> `svc_stufftext`

stufftext already recorded in "net message" cmd.

a1batross:
[compile-fix.patch](https://github.com/user-attachments/files/22719316/compile-fix.patch)

@Garey27 made some changes to fix compiling on Linux. Please apply.

a1batross:
> > Do you want to write commands that are coming from the server or that's locally executed by the client or both?
> 
> Original goldsrc record's only cliendll cmds like +jump/+attack etc (FBitSet(cmd->flags, CMD_CLIENTDLL)), mainly used for demochecking by speedrunners communities like "Kreedz". So only client one.

Then put the branch under `#ifndef XASH_DEDICATED`. It should make it compilable at least.

a1batross:
While testing, I noticed that bringing up console messes up timings during goldsrc demo playback. Playing back xash demos doesn't do this, so the problem is somewhere in `demofile_gs.c`.

Garey27:
> While testing, I noticed that bringing up console messes up timings during goldsrc demo playback. Playing back xash demos doesn't do this, so the problem is somewhere in `demofile_gs.c`.

https://github.com/Garey27/xash3d-fwgs/blob/cl_demo_gs/engine/client/demofile/demofile_gs.c#L278

GoldSRC doesn't record demo fps as Xash3D format, i think something wrong with this logic.

a1batross:
Heh now we need GoldSrc 47 protocol support to playback speedruns. I'm joking of course. :)

engine/client/cl_demo.c
```diff
@@ -16,135 +16,18 @@ GNU General Public License for more details.
...
-	float		realstarttime;
-	float		timestamp;
-	float		lasttime;
-	int		entryIndex;
-
-	// interpolation stuff
-	demoangle_t	cmds[ANGLE_BACKUP];
-	int		angle_position;
-} demo;
+#define FDEMO_TITLE         0x01        // Show title
```
a1batross, line 22: I think these flags can be removed from both here and demofile_xash.c, they were never used for anything meaningful.

a1batross:
Now that it builds everywhere, let's get this merged. @Garey27 is there anything else planned?

Garey27:
I need to properly test this PR on GoldSrc demos. I'm concerned about an old issue where "the console messes up timings during playback." Does this issue still exist, and is there a CVar (console variable) that displays messages related to these timing problems?

Additionally, I'm planning to move the Quake-related demo format to a separate file. To do this, I need to know how to install Quake assets in the Xash3D engine. Could someone provide instructions on this? Also, which game library would I need to use for testing the Quake demo playback in Xash3D?

a1batross:
>I'm concerned about an old issue where "the console messes up timings during playback."

I didn't tested it yet, but it was easy to reproduce:
1. record a goldsrc demo
2. play it back
3. open console with `~` key
4. wait a bit and close it

It's like the demo playback should pause while console is opened but that didn't happened for some reason.

>To do this, I need to know how to install Quake assets in the Xash3D engine

It's easy. You need `id1` folder from Quake (not rerelease IIRC) and Quake Wrapper itself: https://www.moddb.com/games/quake-wrapper/downloads/quake-wrapper-v07-revision-1-with-sources. You can also compile it from the source code: https://github.com/FWGS/quakewrapper

SNMetamorph:
@Garey27 what's up with this PR. Is it ready for merge?

Garey27:
Thanks for checking in, @SNMetamorph. I'll be running tests on this PR over the weekend (November 1-2). It should be ready for merge by the end of the weekend.

000exploit:
Many weekends have passed. It's still ready? Just curious.

WhiteMaterials:
Any news?🙄

PR #2339, opened 2025-10-11
Team Beef's VR port

a1batross:
Updated to upstream branch, but not tested and it doesn't link anywhere except Android.

cc @lvonasek


lvonasek:
You will need this commit to make it run in VR: https://github.com/lvonasek/xash3d-fwgs/commit/38b736642dd25b0b51ab6c2c8f5ce1ad2007014c

lvonasek:
@a1batross CSVR is now opensourced: https://github.com/Team-Beef-Studios/CSVR

I wanted to make this branch working but I didn't manage to compile it on Android.

The steps to get generic OpenXR support for many games are:
* Add or remove bHaphics support from CSVR project.
* Figure out how to render aim cursor without modifying in all games.
* Get rid of motion controls (throwing grenade, knife attack etc.), this would have to be done per game.
* Get rid of two hand weapon feature, this would have to be done per game and you would have to manually split the model to left and right hand part (like community did for us in CSVR case).
* Manage the weapon rendering. The engine takes care that it shots in the controler direction but to make the weapon aligned with the controlers, I modified the StudioModelRenderer in CS client. There are probably better ways to do that but I didn't figure that one out.

---

@velaron, thank you for the great work on cs16-client. I kind of fixed smoke grenade issue I reported here: https://github.com/Velaron/cs16-client/issues/232 but I do not know what exactly fixed it. In the VR version I modified the camera behavior and adjusted the particle geometry to face it. In that phase it started sometimes rendering but it was incorrect. I modifed the renderer to do blending as I needed and then it looked correct. In the client I made a hack that it doesn't remove the smoke immediatelly after spawn. It is probably incorrect but it works in our testing (only for spectator it is sometimes broken).

a1batross:
@lvonasek congrats with the release!

a1batross:
Not viable to support in the current state as it brings a bunch of legacy code shared from another projects.

PR #2342, opened 2025-10-12
Fixed blueish outline on sprays converted to WAD3 for GoldSrc servers

SNMetamorph:
Before:
image

After:
image

### DIFF
--- a/engine/client/cl_spray.c
+++ b/engine/client/cl_spray.c
@@ -151,9 +151,11 @@ qboolean CL_ConvertImageToWAD3( const char *filename )
 			}
 		}
 
+		// make last palette color black to avoid blueish texture filtering artifacts
 		quant->palette[255 * 3 + 0] = 0;
 		quant->palette[255 * 3 + 1] = 0;
-		quant->palette[255 * 3 + 2] = 255;
+		quant->palette[255 * 3 + 2] = 0;
+
 		memcpy( palette, quant->palette, SPRAY_PALETTE_BYTES );
 		indexed = quant->buffer;
 	}


rootcreep:
It is displayed incorrectly in GoldSrc, it is worth doing a revert

image

a1batross:
Yeah, that's unfortunate.

rootcreep:
The problem only occurs when the **gl_texture_nearest** parameter is **disabled** (texture filtering activated). One solution to avoid messing up the display in GoldSrc is to use the average color of opaque pixels, rather than pure blue, when converting indexed wad images to RGBA for a 255 index. It's not a clean solution, but it should work.

```gl_texture_nearest 1```

image

```gl_texture_nearest 0```

image

**Comment by SNMetamorph** [comment]
Of course, disabling texture filtering will fix it. But on GoldSrc spray decals correctly rendered even with enabled texture filtering, so there is some another way to make it working.

**Comment by rootcreep** [comment]
> Of course, disabling texture filtering will fix it. But on GoldSrc spray decals correctly rendered even with enabled texture filtering, so there is some another way to make it working.

Since I don’t fully understand the rendering code, I’m suggesting a simple "hack".

The blue halo appears because the transparent color in WAD textures (index 255) corresponds to pure blue (0, 0, 255). When bilinear filtering is applied, the GPU interpolates between visible pixels and these fully transparent blue pixels, producing unwanted blue edges.

To fix this, instead of using the blue color for transparent pixels (index 255), we calculate the average color of all opaque pixels in the image and use it as the RGB value for transparent pixels (with alpha = 0). This way, texture filtering blends neighboring pixels with a color that matches the spray itself, instead of bright blue.

The palette in WAD files is not modified it still contains (0, 0, 255) at index 255, preserving compatibility with GoldSrc.
The change only affects how indexed images are converted to RGBA for rendering in Xash3D


```diff
diff --git a/engine/common/imagelib/img_wad.c b/engine/common/imagelib/img_wad.c
index 3f0d5740..94c1905a 100644
--- a/engine/common/imagelib/img_wad.c
+++ b/engine/common/imagelib/img_wad.c
@@ -581,6 +581,8 @@ qboolean Image_LoadWAD( const char *name, const byte *buffer, fs_offset_t filesi
 		uint32_t      m0size, m1size, m2size, m3size;
 		qboolean      alpha_mode = false;
 		unsigned char frontR = 0, frontG = 0, frontB = 0;
+		unsigned char avgR = 0, avgG = 0, avgB = 0;
+		int opaque_count = 0;
 		float t;
 		byte  idx;
 
@@ -641,17 +643,57 @@ qboolean Image_LoadWAD( const char *name, const byte *buffer, fs_offset_t filesi
 		image.rgba = Mem_Malloc( host.imagepool, image.size );
 		image.palette = NULL;
 
+		// this is hack
+		// see: https://github.com/FWGS/xash3d-fwgs/pull/2342
+		if( !alpha_mode )
+		{
+			for( j = 0; j < (int)( width * height ); ++j )
+			{
+				idx = pixels[j];
+				if( idx != 255 )
+				{
+					avgR += use_palette[idx * 3 + 0];
+					avgG += use_palette[idx * 3 + 1];
+					avgB += use_palette[idx * 3 + 2];
+					opaque_count++;
+				}
+			}
+			if( opaque_count > 0 )
+			{
+				avgR /= opaque_count;
+				avgG /= opaque_count;
+				avgB /= opaque_count;
+			}
+		}
+
 		// convert indexed pixels to RGBA
 		for( j = 0; j < (int)( width * height ); ++j )
 		{
 			idx = pixels[j];
-			image.rgba[j * 4 + 0] = use_palette[idx * 3 + 0];
-			image.rgba[j * 4 + 1] = use_palette[idx * 3 + 1];
-			image.rgba[j * 4 + 2] = use_palette[idx * 3 + 2];
 			if( alpha_mode )
+			{
+				image.rgba[j * 4 + 0] = use_palette[idx * 3 + 0];
+				image.rgba[j * 4 + 1] = use_palette[idx * 3 + 1];
+				image.rgba[j * 4 + 2] = use_palette[idx * 3 + 2];
 				image.rgba[j * 4 + 3] = idx; // soft transparency
+			}
 			else
-				image.rgba[j * 4 + 3] = ( idx == 255 ) ? 0 : 255; // classic
+			{
+				if( idx == 255 )
+				{
+					image.rgba[j * 4 + 0] = avgR;
+					image.rgba[j * 4 + 1] = avgG;
+					image.rgba[j * 4 + 2] = avgB;
+					image.rgba[j * 4 + 3] = 0; // transparent
+				}
+				else
+				{
+					image.rgba[j * 4 + 0] = use_palette[idx * 3 + 0];
+					image.rgba[j * 4 + 1] = use_palette[idx * 3 + 1];
+					image.rgba[j * 4 + 2] = use_palette[idx * 3 + 2];
+					image.rgba[j * 4 + 3] = 255; // opaque
+				}
+			}
 		}
 		return true;
 	}
```

SNMetamorph:
> To fix this, instead of using the blue color for transparent pixels (index 255), we calculate the average color of all opaque pixels in the image and use it as the RGB value for transparent pixels (with alpha = 0). This way, texture filtering blends neighboring pixels with a color that matches the spray itself, instead of bright blue.

This won't work good enough in case if image has some outline or borders. Because average pixels color will be pretty different with borders color. So, the actual fixes are disabling texture filtering for spray texture, or rendering these spray textures like GoldSrc does.

PR #2343, opened 2025-10-15
android: try to automatically create .nomedia file in directory where games are located

a1batross:
cc @Velaron 

~~Didn't even tested if it builds yet, as I don't have SDK installed right now.~~

Tested, works on my device.


Velaron:
I think I would add removing the .nomedia if the user created it as a folder as this doesn't really work on all devices, maybe doesn't work at all.

a1batross:
> I think I would add removing the .nomedia if the user created it as a folder as this doesn't really work on all devices, maybe doesn't work at all.

Done. Though I think deleting .nomedia recursively, especially silently, is in general bad idea, so I didn't implement that.

Velaron:
Then it seems ok for now.

PR #2349, opened 2025-10-23
engine: client: deprecate and remove all code associated with linear/cubic upsampling

a1batross:
To put upsampling filters out of scope in upcoming mixer rewrite, I want to deprecate and remove s_lerping out of the mixer.

It doesn't mean the feature won't come back, it just really helps to get rid of around 200 lines of code of dubious origin.


a1batross:
If somebody has any objections, please feel free to put your comments.

SNMetamorph:
LGTM

Velaron:
I don't think I've seen this used outside nooby cs tutorials that improve recoil

a1batross:
I've asked Aynekko and KREO3OT in DMs, and they said they never used the feature. I think it can be safely removed.

I will remove it from the menu, update submodule hash and merge this then.

PR #2352, opened 2025-10-28
Restore iOS build support

ksagameng2:
This is based on the old engine port code, feedback would be appreciated.

### DIFF
--- a/engine/platform/ios/launchdialog.m
+++ b/engine/platform/ios/launchdialog.m
@@ -13,6 +13,7 @@
  GNU General Public License for more details.
  */
 
+#include "SDL_syswm.h"
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
 #import <AVFoundation/AVFoundation.h>
@@ -45,6 +46,7 @@ int szArgc;
 char **szArgv;
 char *g_szLibrarySuffix;
 float g_iOSVer;
+bool isdark;
 
 const char *IOS_GetDocsDir(void)
 {
@@ -138,6 +140,7 @@ void IOS_PrepareView(void)
 	[[controller view] setBackgroundColor:[UIColor grayColor]];
 	[window setRootViewController:controller];
 	[window makeKeyAndVisible];
+	if([[controller traitCollection] userInterfaceStyle] == UIUserInterfaceStyleDark && g_iOSVer >= 13.0) isdark = true; else isdark = false;
 #if 0
 	UIButton *button = [[UIButton alloc] initWithFrame:CGRectMake(10, 10, 100, 20)];
 	int button1 = -1;
@@ -195,6 +198,8 @@ void IOS_LaunchDialog( void )
 
 	UITextField *args = [[UITextField alloc] initWithFrame:CGRectMake(0, 30, 300, 30)];
 	[args setBackgroundColor:[[UIColor alloc] initWithRed:1 green:1 blue:1 alpha:1]];
+	if(isdark) [args setBackgroundColor:[[UIColor alloc] initWithRed:0 green:0 blue:0 alpha:1]];
+	
 
 	UILabel *ftptitle = [[UILabel alloc] initWithFrame:CGRectMake(0, 60, 200, 30)];
 	[ftptitle setText:@"FTP Server"];
@@ -204,6 +209,7 @@ void IOS_LaunchDialog( void )
 	
 	UITextField *suffix = [[UITextField alloc] initWithFrame:CGRectMake(140, 90, 160, 30 )];
 	[suffix setBackgroundColor:[[UIColor alloc] initWithRed:1 green:1 blue:1 alpha:1]];
+	if(isdark) [suffix setBackgroundColor:[[UIColor alloc] initWithRed:0 green:0 blue:0 alpha:1]];
 
 	UILabel *suffixtitle = [[UILabel alloc] initWithFrame:CGRectMake(0, 90, 140, 30)];
 	[suffixtitle setText:@"Library suffix"];


SNMetamorph:
At first, follow CONTRIBUTING.md rules for naming commits.

nekonomicon:
Please do not upload `*.dylib.`
Also, in the old port there was a script to build all available libraries for mods:
https://github.com/mittorn/xash3d-ios/blob/master/scripts/build-all-mods.sh
And actual list of mods here:
https://github.com/FWGS/hlsdk-portable/wiki/Mods
And tfc here:
https://github.com/Velaron/tf15-client
Also, for some simple mods like `Blue Shift` or `Absolute Redemption` there currently exists `mobile_hacks` branch in `hlsdk-portable`.

ksagameng2:
Sorry for that, I was running into git related issues and they somehow were uploaded, will fix it and try to implement a build mods script tomorrow

a1batross:
Wait, that wasn't a dream.

Good job on the port!

engine/platform/ios/launchdialog.m
```diff
@@ -0,0 +1,367 @@
...
+	[argstitle release];
+	[port release];
+	[suffix release];
+	[suffixtitle release];
+
+	[alert release];
+	}
+	else // do not know what wrong with 6.x, just skip launch dialog
+	{
+		static char *args[32] = { "xash", "-dev", "5", "-log"};
```
a1batross: Xash3D doesn't use so many dev levels anymore.

It's `-console`, `-dev`, `-dev 2`.

The better way is to use `-console`, or completely read this from `cmdline.txt`, if user can modify this file.

engine/platform/ios/lib_ios.c
```diff
@@ -12,13 +12,15 @@ but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
 */
+#include <string.h>
```
a1batross, line 15: I think it can be moved under `#if`.

But also because you moved this file into it's own `platform/ios` directory, we can get rid of `#if` here and add these files for build only when we target iOS.

engine/platform/ios/lib_ios.c
```diff
@@ -12,13 +12,15 @@ but WITHOUT ANY WARRANTY; without even the implied warranty of
...
 #if TARGET_OS_IPHONE
 #include <SDL.h>
 #include "common.h"
 #include "library.h"
 #include "filesystem.h"
 #include "server.h"
-#include "platform/apple/ios_lib.h"
+#include "platform/ios/lib_ios.h"
+#include "dlfcn.h"
```
a1batross: it's `<dlfcn.h>`

engine/platform/ios/lib_ios.c
```diff
@@ -52,17 +54,17 @@ static void *IOS_LoadLibraryInternal( const char *dllname )
...
 
 	string name;
 	char *postfix = g_szLibrarySuffix;
 	char *pHandle;
-
-	if( !postfix ) postfix = GI->gamefolder;
-
-	Q_snprintf( name, MAX_STRING, "%s_%s", dllname, postfix );
+	
+	if ( !postfix ) postfix = "";
```
a1batross: We don't put spaces after `if`

engine/platform/sdl2/vid_sdl2.c
```diff
@@ -496,10 +496,8 @@ static qboolean VID_SetScreenResolution( int width, int height, window_mode_t wi
 {
 	SDL_DisplayMode got;
 	Uint32 wndFlags = 0;
-
 	if( vid_highdpi.value )
```
a1batross, line 500: Unnecessary change.

ios/dlls/cs_arm64.dylib
a1batross, line 1: Don't commit binaries here.

engine/wscript
```diff
@@ -205,7 +208,11 @@ def build(bld):
 	# include common POSIX conformant code
 	if bld.env.DEST_OS not in ['win32', 'dos']:
 		source += bld.path.ant_glob('platform/posix/*.c')
-
+	if bld.env.DEST_OS2 == 'ios':
```
a1batross: Please use `env.DEST_OS` for this.

scripts/ios/createipa.sh
a1batross, line 1: Put platform-specific scripts in `scripts` directory.

engine/platform/ios/bundle/ftp_commands.plist
a1batross, line 1: Aren't `plist` files are supposed to be text files?

a1batross: Anyway, if it has a source file that can be compiled into this binary format, it's probably better to keep it in `engine/platform/ios`.

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,49 @@
...
+class iOS:
+	ctx = None
+	simulator = None
+	sdkpath = None
+	target = None
+	
+	def __init__(self, ctx, simulator):
+		self.ctx = ctx
+
+		if simulator == 'true': sdk = "iphonesimulator"
```
a1batross:
```py
sdk = 'iphonesimulator' if simulator else 'iphoneos'
```

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,49 @@
...
+		else: sdk = "iphoneos"
+
+		path = subprocess.run(["xcrun", "--show-sdk-path", "--sdk", sdk], stdout=subprocess.PIPE)
```
a1batross: There is waf context method for this, it should be something like this:

```python
self.sdkpath = ctx.cmd_and_log('xcrun --show-sdk-path --sdk %s' % sdk).strip()
```

Also, `simulator` could be a boolean.

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,49 @@
...
+		self.sdkpath = str.rstrip(self.sdkpath)
+
+		self.target = '--target=aarch64-apple-ios'
+		if simulator == 'true': self.target += '-simulator'
+	
+	def cflags(self, cxx = False):
+	
+		cflags = []
+	
+		cflags += [ '-isysroot' + self.sdkpath, self.target, '-mios-version-min=12.0', '-Wno-shorten-64-to-32', '-fpic' ]
```
a1batross: Please avoid adding any kind of `-W` flag here.

Same with `-fpic`, it must be enabled already through `enforce_pic` waf tool.

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,49 @@
...
+		return cflags
+		
+	def linkflags(self):
+		
+		linkflags = []
+		
+		linkflags += [ '-isysroot' + self.sdkpath, self.target, '-mios-version-min=12.0', '-dead_strip', '-fpic' ]
```
a1batross: What's the reason of `-dead_strip`?

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,49 @@
...
+		
+		return linkflags
+		
+	def cc(self):
+
```
a1batross, line 69: Remove unnecessary newlines.

scripts/waifulib/xcompile.py
```diff
@@ -754,6 +799,28 @@ def configure(conf):
 
 		conf.env.PSP_RENDER_TYPE = psp.render_type
 		conf.env.PSP_BUILD_PRX = psp.build_prx
+	elif conf.options.IOS_OPTS:
+		values = conf.options.IOS_OPTS.split(',')
```
a1batross: You don't need to split it if you don't accept multiple arguments here.

scripts/waifulib/xcompile.py
```diff
@@ -754,6 +799,28 @@ def configure(conf):
...
+		if len(values) != 1:
+			conf.fatal('Invalid --ios paramater value!')
+
+		simulatorvalues = ['true', 'false']
```
a1batross: ... Just cast it into a boolean.

scripts/waifulib/xcompile.py
```diff
@@ -754,6 +799,28 @@ def configure(conf):
...
+		if values[0] not in simulatorvalues:
+			conf.fatal("Invalid build for simulator answer, you must answer with true or false!")
+		conf.ios = ios = iOS(conf, values[0])
+		conf.env['cshlib_PATTERN'] = 'lib%s.dylib'
```
a1batross: It's something `waf` already should do by detecting target system as Apple.

What's the reason of setting up `c{,xx}shlib_PATTERN`, `FRAMEWORK_ST`, `FRAMEWORKPATH_ST`?

As for `FRAMEWORK` flag, it will just link everything you listed to _every_ executable and shared library. Please, use Waf's uselib for that. There is probably a way to check existence of these frameworks which will generate a uselib. It goes something (very loosely) like this:

```python
conf.check(framework_name='Foundation', uselib_store='everything_apple')
conf.check(framework_name='UIKit', uselib_store='everything_apple')
conf.check(framework_name='QuartzCore', uselib_store='everything_apple')
...
```

And then in engine build task:
```python
bld.shlib(..., use='some_other_libs everything_apple')
```

scripts/waifulib/xcompile.py
```diff
@@ -754,6 +799,28 @@ def configure(conf):
...
+		conf.env['FRAMEWORKPATH_ST'] = '-F%s'
+		conf.env['FRAMEWORK'] = [ 'Foundation', 'UIKit', 'QuartzCore', 'GameController', 'CoreMotion', 'SystemConfiguration', 'CoreFoundation', 'CFNetwork', 'AVFoundation', 'CoreGraphics' ]
+
+		conf.environ['CC'] = ios.cc()
+		conf.environ['CXX'] = ios.cxx()
+		
+		conf.env.CFLAGS += ios.cflags()
+		conf.env.CXXFLAGS += ios.cflags()
+		conf.env.LINKFLAGS += ios.linkflags()
+		conf.env.DEST_OS2 = 'ios'
```
a1batross: Again, check if you can set `DEST_OS`. Maybe similarly to what we do on Android, which by default gets detected as `linux`

README.md
```diff
@@ -98,6 +98,16 @@ This repository contains our fork of HLSDK and restored source code for Half-Lif
 
 * Clone this repostory: `$ git clone --recursive https://github.com/FWGS/xash3d-fwgs`.
 
+#### iOS/iPadOS
```
a1batross, line 101: It would be good if you set up a CI build and deploy.

GitHub Actions provide macOS machines for free.

scripts/waifulib/xcompile.py
```diff
@@ -610,6 +653,8 @@ def options(opt):
 		help='enable building for Emscripten')
 	xc.add_option('--psp', action='store', dest='PSP_OPTS', default=None,
 		help='enable building for PlayStation Portable, format: --psp=<module type>,<fw version>,<render type>, example: --psp=prx,660,HW')
+	xc.add_option('--ios', action='store', dest='IOS_OPTS', default=None,
```
a1batross: I have a better suggestion: `--ios` flag to target real device, `--ios-simulator` to target the simulator?

wscript
```diff
@@ -247,6 +247,12 @@ def configure(conf):
 		conf.options.GL4ES            = True
 		conf.options.GLES3COMPAT      = True
 		conf.options.GL               = False
+	if conf.env.DEST_OS2 == 'ios':
+		conf.options.NANOGL           = True
```
a1batross, line 251: Do we still have GLES1 on iOS? Oh wow, I'm surprised.

engine/platform/ios/launchdialog.m
ksagameng2: This file only gets generated if the OS is above or equal to ios 7.0, It's from the old port so I guess I could remove it since compiling for an OS this old on a modern mac is unsupported now

scripts/waifulib/xcompile.py
ksagameng2: Unfortunately if I force DEST_OS to ios then waf will end up passing incompatible linker flags, of course I could modify the problematic scripts but I assume that's a big no no

engine/platform/ios/launchdialog.m
a1batross: You're correct, we can remove that.

scripts/waifulib/xcompile.py
a1batross: Ah, you're right, `waf` internally depends on DEST_OS being equal to `darwin` on Apple systems.

IMHO similarly to MotoMAGX and Sailish ports, that are basically GNU/Linux, I think we don't need `DEST_OS2`, but maybe separate variable just for iOS targets.

So instead of: `ctx.env.DEST_OS2 == 'ios'` which misleadingly suggests there might be other DEST_OS2 values (when there aren't any), let's create a boolean `ctx.env.IOS` variable and use it instead.

ios/Info.plist
```diff
@@ -0,0 +1,78 @@
...
+<plist version="1.0">
+<dict>
+	<key>BuildMachineOSBuild</key>
+	<string>25A362</string>
+	<key>CFBundleDevelopmentRegion</key>
+	<string>en</string>
+	<key>CFBundleExecutable</key>
+	<string>xash</string>
+	<key>CFBundleIdentifier</key>
+	<string>com.ksagameng.xash3d-fwgs-ios</string>
```
a1batross: Keep it the same with Android port.

It's `su.xash.engine`.

a1batross: Also, `plist` files can be moved to `engine/platform/ios`.

scripts/ios/buildhlsdk.sh
```diff
@@ -0,0 +1,34 @@
...
+    git clone --recursive https://github.com/FWGS/hlsdk-portable -b $1 $MODPATH
+fi
+
+cd $MODPATH
+
+#compiling hlsdk for ios release is broken, so just build for debug
+cmake -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_BUILD_TYPE=Debug -B build -S .
+cmake --build build
+
+if [ -z $2 ]; then
```
a1batross: There is `install` target implemented.

scripts/ios/createipa.sh
```diff
@@ -0,0 +1,40 @@
...
+    cp ./*.plist ../build/xash3d.app
+    if [ ! -d SDL2.framework ]; then
+        echo "Couldn't find SDL2.framework, place it in the 'ios' directory"
+        exit 1
+    fi
+    cp -r SDL2.framework ../build/xash3d.app
+
+    cd ../build
+
+    if [ ! -e engine/xash ]; then
```
a1batross: Again, we also have `install` here that will copy all the binaries.

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,37 @@
...
+	def cflags(self, cxx = False):
+	
+		cflags = []
+	
+		cflags += [ '-isysroot' + self.sdkpath, self.target, '-mios-version-min=12.0' ]
+		return cflags
+		
+	def linkflags(self):
+		
+		linkflags = []
```
a1batross: That's unnecessary.

Just do:
```python
linkflags = [blahblah]
return linkflags
```

scripts/waifulib/xcompile.py
```diff
@@ -45,6 +45,37 @@
...
+	def __init__(self, ctx, issim):
+		self.ctx = ctx
+		sdk = 'iphonesimulator' if issim else 'iphoneos'
+		self.sdkpath = ctx.cmd_and_log('xcrun --show-sdk-path --sdk %s' % sdk).strip()
+		self.target = '--target=aarch64-apple-ios'
+		if issim: self.target += '-simulator'
+	
+	def cflags(self, cxx = False):
+	
+		cflags = []
```
a1batross: That's unnecessary.

Just do:
```python
cflags = [blahblah]
return cflags
```

ios/Info.plist
ksagameng2: By the way they don't actually need to be compiled, I just originally copied them from the Xcode project that I had which makes them in binary format by default

ios/Info.plist
```diff
@@ -0,0 +1,78 @@
...
+	<key>CFBundleInfoDictionaryVersion</key>
+	<string>6.0</string>
+	<key>CFBundleName</key>
+	<string>xash3d-fwgs-ios</string>
```
a1batross: It's also just `xash3d-fwgs`.

ios/Info.plist
a1batross: Can you explain then how they get compiled during build?

ksagameng2: They get converted by sideloading software automatically

a1batross: Oh, ok. Still, I would like to not pollute root directory with new folders.

scripts/ios/buildhlsdk.sh
```diff
@@ -0,0 +1,35 @@
...
+mkdir ../../build/ios
+IOSDIR=$(realpath ../../build/ios)
+cd $MODPATH
+
+#compiling hlsdk for ios release is broken, so just build for debug
+cmake -DCMAKE_SYSTEM_NAME=iOS -DGAMEDIR=$IOSDIR -DCMAKE_BUILD_TYPE=Debug -B build -S .
+cmake --build build --target install
+
+
+#if [ -z $2 ]; then
```
a1batross: Don't keep code in the comments.

README.md
```diff
@@ -98,6 +98,16 @@ This repository contains our fork of HLSDK and restored source code for Half-Lif
...
+* Install Xcode from the appstore.
+* Install [Homebrew package manager](https://brew.sh).
+
+* Install build dependencies by running: `brew install python sdl2 freetype opus opusfile libvorbis libogg`.
+
+* Clone the SDL2 repo `$ git clone --recursive https://github.com/libsdl-org/SDL.git -b SDL2` and compile the iOS framework by navigating to SDL/Xcode/SDL and opening the Xcode project.
+
+* Clone this repostory: `$ git clone --recursive https://github.com/ksagameng/xash3d-fwgs-ios`.
```
a1batross: Change link to our repository please. :)

a1batross:
Is it done?

ksagameng2:
Should be ready after that last fix

SNMetamorph:
By the way, did you get it working on actual iOS device? Or this is just makes engine buildable for iOS?

ksagameng2:
It actually works on real hardware
IMG_1962

PR #2356, opened 2025-10-31
A little refactored voice handling

pwd491:


a1batross:
Just by quick looking, doesn't look suspicious. I will review it normally in the morning.

By the way, me and @SNMetamorph discussed slowly phasing out custom modes for voice chat in Xash protocol and instead using the same parameters as GoldSrc. In this case, we could probably reuse the same code some time in the future.

rootcreep:
> By the way, me and @SNMetamorph discussed slowly phasing out custom modes for voice chat in Xash protocol and instead using the same parameters as GoldSrc. In this case, we could probably reuse the same code some time in the future.

Yes, I also thought about it, it would greatly simplify the code of the processing and packaging logic, but it would break compatibility with older clients, which would cause a wave of false error messages in voice chat.

a1batross:
Works for me, merging. Good job!

PR #2363, opened 2025-11-05
Swap bandwidthtest and challenge

a1batross:
The problem with bandwidthtest packet is that it doesn't require any confirmation, thus can be easily exploited in attacks and, by requesting largest packet size, exhaust all bandwidth.

To resolve this issue, bandwidthtest now requires challenge first. Client is expected to first request the challenge and then complete bandwidthtest, and at any point client can send connect packet.

For compatibility with older clients, server interprets bandwidthtest packet as challenge request, making client connect with fragmentation enabled. Same with new clients on old servers: they completely ignore bandwidth test and connect with `FRAGMENT_DEFAULT_SIZE` packet size.

In future, it might make sense to deprecate bandwidth test as it in my experience often doesn't help to figure out MTU. Old engine and GoldSrc used 1400 and 1200 bytes limit respectively and it seems to work almost everywhere an where it didn't in old engine there was connection issues dialog in the UI that suggested some useful presets. 

Thanks to troll338cz for report and an idea to swap them.

### DIFF
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -2391,7 +2391,7 @@ static void CL_Challenge( const char *c, netadr_t from )
 	{
 		// when connecting to old server or server that has bandwidth test disabled
 		// it might be more preferrable to have some sane fragment size
-		if( !Q_atoi( Cmd_Argv( 2 ))
+		if( !Q_atoi( Cmd_Argv( 2 )))
 		{
 			Cvar_SetValue( "cl_dlmax", FRAGMENT_DEFAULT_SIZE );
 			CL_SendConnectPacket( cls.legacymode, cls.bandwidth_test.challenge );


SNMetamorph:
> In future, it might make sense to deprecate bandwidth test as it in my experience often doesn't help to figure out MTU.

Absolutely agree, because:
1) We cannot rely on IP fragmentation, because there is no guarantee that fragmented packets wouldn't be dropped on some of network nodes due to high load ([RFC4963](https://www.rfc-editor.org/rfc/rfc4963))
2) We cannot rely on PMTUD, because some network nodes drops ICMP traffic (usually this called *ICMP black hole*) which is required for proper working of PMTUD ([RFC2923](https://www.rfc-editor.org/rfc/rfc2923))

Therefore, the only reliable option is to lock on some fixed value that will working everywhere. Like 1200 bytes for IPv4 and 1280 bytes for IPv6.

SNMetamorph (approved):

engine/client/cl_main.c
```diff
@@ -2352,8 +2385,27 @@ static void CL_Challenge( const char *c, netadr_t from )
...
 
-	// challenge from the server we are connecting to
-	CL_SendConnectPacket( cls.legacymode, Q_atoi( Cmd_Argv( 1 )));
+	cls.bandwidth_test.challenge = Q_atoi( Cmd_Argv( 1 ));
+
+	if( cls.legacymode == PROTO_CURRENT && cl_test_bandwidth.value && !cls.bandwidth_test.passed )
+	{
+		// when connecting to old server or server that has bandwidth test disabled
+		// it might be more preferrable to have some sane fragment size
+		if( !Q_atoi( Cmd_Argv( 2 ))
```
SNMetamorph: Missing closing bracket here

a1batross:
Yeah, I made it last minute and didn't checked. Thanks

1

Issue #2383, opened 2025-11-27
Screen resolution changing works weird


SNMetamorph:
1. At first, game running at fullscreen 1920x1080
2. Switching screen resolution to windowed 1600x900
3. Video mode changed, shutting down game
4. Running game, it starts as windowed but in 1920x1080 resolution

SNMetamorph:
Logs (though I don't see any useful info):
```
================================================================================
Xash3D FWGS 0.21 (3893, 2736d1fd, master, win32-i386)
Game started at Nov27 2025 [22:49.11]
================================================================================
Console initialized.
filesystem_stdio: custom pool allocation functions found
filesystem_stdio: custom memory allocation functions found
filesystem_stdio: custom platform-specific functions found
FS_LoadProgs: filesystem_stdio successfully loaded
D:/xashstuff/xash3d-fwgs-x86 is working directory now
Adding directory: ./
FS_InitStdio: done
Program args: xash3d.exe -game valve -log -dev 2 
Developer level: 2
couldn't exec valve/vfs.cfg
FS_LoadGameInfo( valve, 0x0, english )
FS_Rescan( Half-Life )
FS_AddGameHierarchy( valve )
Adding WAD: valve/downloaded/cs_assault.wad (22 files)
Adding WAD: valve/downloaded/fpsb.wad (1 files)
Adding directory: valve/downloaded/
Adding PAK: valve/hg_industrialzone.pak (35 files)
Adding WAD: valve/hg_industrialzone.pak/chateau.wad (136 files)
Adding WAD: valve/hg_industrialzone.pak/cs_747.wad (143 files)
Adding WAD: valve/hg_industrialzone.pak/cs_bdog.wad (132 files)
Adding WAD: valve/hg_industrialzone.pak/cs_havana.wad (122 files)
Adding WAD: valve/hg_industrialzone.pak/de_airstrip.wad (69 files)
Adding WAD: valve/hg_industrialzone.pak/de_piranesi.wad (160 files)
Adding WAD: valve/hg_industrialzone.pak/itsitaly.wad (99 files)
Adding WAD: valve/hg_industrialzone.pak/torntextures.wad (162 files)
Adding PAK: valve/scp_foundation_bw38.pak (1 files)
Adding ZIP: valve/extras.pk3 (118 files)
Adding WAD: valve/cached.wad (2 files)
Adding WAD: valve/decals.wad (222 files)
Adding WAD: valve/fonts.wad (3 files)
[...]

PR #2385, opened 2025-11-30
engine: fix vid_maximized always set to true when switching to window…

stevenlafl:
…ed mode

Resolves: #2383


SNMetamorph:
Thanks! Looks good to me.

a1batross:
Good job, thanks!

PR #2387, opened 2025-11-30
engine: drop legacy protocol support

a1batross:
Now that old protocol servers were removed from our master servers few months ago, it's time to drop it from the engine as well. Cleans up ~1kLoC.

Thanks @mittorn for implementing it many years ago and making rough upgrade path slightly easier.

### DIFF
--- a/engine/common/netchan.h
+++ b/engine/common/netchan.h
@@ -167,10 +167,10 @@ typedef enum fragsize_e
 
 typedef enum netchan_flags_e
 {
-	NETCHAN_USE_MUNGE = BIT( 1 ),
-	NETCHAN_USE_BZIP2 = BIT( 2 ),
-	NETCHAN_GOLDSRC = BIT( 3 ),
-	NETCHAN_USE_LZSS = BIT( 4 ), // mutually exclusive with bzip2
+	NETCHAN_USE_MUNGE = BIT( 0 ),
+	NETCHAN_USE_BZIP2 = BIT( 1 ),
+	NETCHAN_GOLDSRC   = BIT( 2 ),
+	NETCHAN_USE_LZSS  = BIT( 3 ), // mutually exclusive with bzip2
 } netchan_flags_t;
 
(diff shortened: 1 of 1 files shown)


common/net_api.h
```diff
@@ -27,7 +27,6 @@
 // Set this flag for things like broadcast requests, etc. where the engine should not
 // kill the request hook after receiving the first response
 #define FNETAPI_MULTIPLE_RESPONSE		( 1<<0 )
-#define FNETAPI_LEGACY_PROTOCOL			( 1<<1 ) // xash3d-fwgs extension
```
SNMetamorph, line 30: Won't it break something? Maybe keep this bit as reserved?

engine/common/netchan.h
```diff
@@ -207,7 +167,6 @@ typedef enum fragsize_e
 
 typedef enum netchan_flags_e
 {
-	NETCHAN_USE_LEGACY_SPLIT = BIT( 0 ),
```
SNMetamorph, line 210: I think it's worth keep notice that this bit already reserved by deprecated flag.

SNMetamorph (approved):

engine/common/netchan.h
a1batross: It's an internal flag. If nothing uses it, it's fine if it will get reused for something in the future. 

But leaving the comment might make sense.

common/net_api.h
a1batross: Yes, in this case, keeping it reserved might makse sense, plus we have a plenty of space here.

engine/common/netchan.h
a1batross: Instead I shifted all the bit flags to the right.

common/net_api.h
a1batross: Left a comment for this.

Velaron (approved):

a1batross:
Merged.

PR #2412, opened 2025-12-30
engine: fix black fullscreen flash on startup in windowed mode

stevenlafl:
... and fix saved window position drift.

Changes:

- Set SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR to "0" to prevent a brief fullscreen black flash when launching in windowed mode on X11 with compositing enabled.

- Fix saved window position drifting downward on each launch. SDL reports client area position via SDL_GetWindowPosition and SDL_WINDOWEVENT_MOVED, but SDL_CreateWindow positions the window frame. This mismatch caused the position to drift by the title bar height on each launch. **Fixed by subtracting the window decoration offset (via SDL_GetWindowBordersSize) before saving, so coordinates represent the frame position that SDL_CreateWindow expects.**

### DIFF
--- a/engine/platform/sdl2/host_sdl2.c
+++ b/engine/platform/sdl2/host_sdl2.c
@@ -387,11 +387,22 @@ static void SDLash_EventHandler( SDL_Event *event )
 		case SDL_WINDOWEVENT_MOVED:
 		{
 			char val[32];
+			int top, left;
+			int x = event->window.data1;
+			int y = event->window.data2;
 
-			Q_snprintf( val, sizeof( val ), "%d", event->window.data1 );
+			// adjust for window decorations - SDL reports client area position,
+			// but SDL_CreateWindow positions the frame
+			if( SDL_GetWindowBordersSize( host.hWnd, &top, &left, NULL, NULL ) == 0 )
+			{
+				x -= left;
+				y -= top;
+			}
+
+			Q_snprintf( val, sizeof( val ), "%d", x );
 			Cvar_DirectSet( &window_xpos, val );
 
-			Q_snprintf( val, sizeof( val ), "%d", event->window.data2 );
+			Q_snprintf( val, sizeof( val ), "%d", y );
 			Cvar_DirectSet( &window_ypos, val );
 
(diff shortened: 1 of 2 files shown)


a1batross:
To be honest, I don't really think we need to save window position at all. If we don't try to force window position, it will be much easier to handle all of this.

The only reason I tried to do this is to make it easier during development and select screen in multi-monitor configurations.

Can I ask you? Did you use an AI to create these PRs? If you did, how exactly, was it only for API reference or code generation too?

stevenlafl:
Not to that extent, just to help with some things I didn't understand. Although I've experimented before, I'm not as familiar with SDL or X programming.

Did you want to just remove it and center it every time?

a1batross (approved):

a1batross:
>Set SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR to "0" to prevent a brief fullscreen black flash when launching in windowed mode on X11 with compositing enabled.

I'm going to revert this part as apparently this hint is used to _disable_ compositing in X11.

Disabled compositing helps with performance on very low end GPUs and it's not really our job if WM is like that and renders whole screen black while toggling compositor.

If that's an issue, it's probably should be configured in the WM itself. `kwin_x11` have such option in per-application rules for example.

stevenlafl:
What's the adverse affects? I'm on a modern RX 9070XT, and I get this issue. I don't in other games. Why do you think it is? Maybe there's some other way to fix this. I just wouldn't expect the entire full screen to black before displaying the window

a1batross:
On X11 it's usually possible to do software and hardware compositing. Software only allows basic window placement, while hardware enable special effects like shadows, wobbly windows and so on.

The switch from one to another requires the whole screen to be redrawn.

And that's probably where your window manager makes a mistake, instead of simply disabling effects and redrawing all the windows it for some reason flashes the whole screen.

I've seen this bug, but can't reproduce it on AMD right now (also on 9070XT) and don't remember if that was an issue on NVIDIA as well. So that might be window manager bug. If that helps, I'm on `kwin_x11 6.5.4`.

PR #2413, opened 2025-12-30
engine: client: block player input during intermission

### LINKED ISSUE #2409: End-map scoreboard (SVC_INTERMISSION) allows players to move and shoot
Self explanatory. When changing levels (with e.g. amx_map that sends SVC_INTERMISSION) I get the end round scorecard, but I can still shoot and move. 

Image

stevenlafl:
During SVC_INTERMISSION, the server blocks movement and actions, but the client was still allowing players to move and shoot visually.

Changes:
- cl_main.c: Add !cl.intermission to the 'active' flag passed to CL_CreateMove(). This tells the client DLL not to process movement input and handles keys that were already held before intermission. Then we clear cmd->buttons after all input processing completes to prevent all inputs regardless of where it was initiated from.

Closes #2409

### DIFF
--- a/engine/client/keys.c
+++ b/engine/client/keys.c
@@ -766,7 +766,7 @@ void GAME_EXPORT Key_Event( int key, int down )
 	}
 
 	// distribute the key down event to the apropriate handler
-	if( cls.key_dest == key_game && !cl.intermission )
+	if( cls.key_dest == key_game )
 	{
 		Key_AddKeyCommands( key, kb, down );
 	}
--- a/engine/client/cl_main.c
+++ b/engine/client/cl_main.c
@@ -686,6 +686,10 @@ static void CL_CreateCmd( void )
 	clgame.dllFuncs.CL_CreateMove( host.frametime, cmd, active );
 	IN_EngineAppendMove( host.frametime, cmd, active );
 
+	// clear all input during intermission regardless of how it was set
+	if( cl.intermission )
+		cmd->buttons = 0;
+
 	CL_PopPMStates();
 
 	if( !cls.demoplayback )


SNMetamorph:
LGTM

engine/client/keys.c
stevenlafl, line 1: Noticed a problem, this can be bypassed with aliases. We should clear buttons in intermission at cl_main.c

stevenlafl (commented):
Initially, we added an intermission check in Key_Event() before calling Key_AddKeyCommands():
```
if( cls.key_dest == key_game && !cl.intermission )
    Key_AddKeyCommands( key, kb, down );
```
This only blocks direct key bindings. Aliases bypass it entirely:
```
  alias attack "+attack"
  bind mouse1 attack
```

When mouse1 is pressed:
  1. Key_AddKeyCommands() sees "attack" (no + prefix, passes check)
  2. Command buffer queues "attack\n"
  3. Alias expands to "+attack" during command execution
  4. +attack runs, IN_ATTACK flag is set
  
The fix: clear cmd->buttons after all input processing completes:
```
  clgame.dllFuncs.CL_CreateMove( host.frametime, cmd, active );
  IN_EngineAppendMove( host.frametime, cmd, active );
```

As of commenting, this fix has been applied and the PR is ready for re-review.

SNMetamorph:
But what if shouldn't send usercmd's during intermission at all? Is it correct approach?

stevenlafl:
Buttons was the main problem in the client - we aren't actually sending buttons by blocking it here - it's only for the client vis. To your credit, it's possible a mod implements a command they don't actually intend to be allowed to run that does not set `buttons`. According to @a1batross in #2409 however, no server implementation allows commands during SVC_INTERMISSION. But maybe he can confirm that's actually the case, as it's possible I misunderstood.

a1batross:
It's a fine fix, but IIRC GoldSrc doesn't do this.

This is what has been puzzling me, as a mod might want to accept player's input in theory and if GoldSrc allows that, we should too.

Or, if it doesn't, it's safe to assume anybody will ever do that.

a1batross:
Superseded by bdc205b994cfc71719429d4c37e525247f26f7bb

PR #2414, opened 2025-12-30
engine: client: allow console toggle from menu via keybind

stevenlafl:
Changes:
 - Allows opening/closing the console from the main menu using the console keybind (default `)
 - Console toggle now returns to the **same menu screen** instead of resetting to main menu
 - Works consistently whether console is opened via keybind or via menu's Console button
 
Previously, pressing the console key while in the menu was blocked by the condition:
```c++
  if( cls.state != ca_active || cls.key_dest == key_menu )
      return;
```

It appears this was intentional, but it's unclear **why** it was intentional. I use the console constantly, so having that shortcut immediately upon game launch is extremely valuable for executing saved history (e.g. uparrow, `connect 127.0.0.1`, enter)

### DIFF
--- a/engine/client/console.c
+++ b/engine/client/console.c
@@ -263,8 +263,8 @@ void Con_ToggleConsole_f( void )
 
 	SCR_EndLoadingPlaque();
 
-	// show console only in game or by special call from menu
-	if( cls.state != ca_active || cls.key_dest == key_menu )
+	// show console only in game, from menu, or to close console
+	if( cls.state != ca_active && cls.key_dest != key_menu && cls.key_dest != key_console )
 		return;
 
 	Con_ClearTyping();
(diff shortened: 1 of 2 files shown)


engine/client/console.c
```diff
@@ -263,22 +263,32 @@ void Con_ToggleConsole_f( void )
...
+		{
+			// in game - return to game
+			UI_SetActiveMenu( false );
+		}
 	}
 	else
 	{
-		UI_SetActiveMenu( false );
+		// opening console - just switch key_dest, don't call UI_SetActiveMenu
+		// which would reset menu state via on_menu_hide()
```
a1batross, line 291: What on_menu_hide?

stevenlafl: That's a from maintui. I was having trouble at one point and had left a comment. My build uses: `configure --enable-tui` and I tested with `-menulib ./libmenu_tui.so`

a1batross (approved):

a1batross:
I will have to revert this one for now due to a bug when we have active menu while engine is technically in `key_game` key_dest.

As @Aynekko told me, the easy way to reproduce the bug is to:
* Enable ui_renderworld 1 so that you can see what's happening in the game.
* Start a new game.
* Press ESC to open the menu.
* Press `~` to open the console.
* Press ESC to go back to menu.
* Instead, we see the menu rendering but the game receives the keyboard input.

stevenlafl:
So to be clear this is a bigger bug that needs addressing later but for now we're going to fix it by temporarily reverting so this use case is covered?

I've been rummaging around this code almost every day in some capacity, so if the real problem is deeper I might want to poke around there.

a1batross:
Yeah, I would like to keep this feature. I tried to fix it myself, but don't really get your intention either, i.e. why you want to keep the menu state.

I thought of making SetKeyDest save the previous state and restoring it through a separate function. That would make the logic much cleaner but that's also much more work.

a1batross:
>I thought of making SetKeyDest save the previous state and restoring it through a separate function. That would make the logic much cleaner but that's also much more work.

Well, now that I said it, it feels much easier than I thought. Maybe I will implement it right now then.

a1batross:
Nevermind, sorry, I fixed it here: a26c56629e39b5e5adb729a13c45279ecf2d55da

stevenlafl:
> > I thought of making SetKeyDest save the previous state and restoring it through a separate function. That would make the logic much cleaner but that's also much more work.
> 
> Well, now that I said it, it feels much easier than I thought. Maybe I will implement it right now then.

Perhaps conciseness wasn't the answer here. It seemed redundant to manually set a lot that that Menu function does, iirc.

stevenlafl:
> Nevermind, sorry, I fixed it here: [a26c566](https://github.com/FWGS/xash3d-fwgs/commit/a26c56629e39b5e5adb729a13c45279ecf2d55da)

Nice job!!

PR #2431, opened 2026-01-15
engine: client: implement steam broker client side

a1batross:
To be used with https://github.com/FWGS/steam-broker

Early proof of concept. Able to join servers online in Half-Life. 

TODOs/ideas for polishing:
- [ ] Only start Steam API with sb_connect and stop on sb_terminate, so that we don't keep ourselves online too much
- [ ] Do rich presence like HL25?
- [ ] Obvious code cleanups and error reporting

### DIFF
--- a/engine/client/cl_steam.c
+++ b/engine/client/cl_steam.c
@@ -80,11 +80,12 @@ void SteamBroker_TerminateGameConnection( void )
 	if( cls.legacymode != PROTO_GOLDSRC )
 		return;
 
-	// sb_terminate <ip:port> <challenge>quitйгше
+	// sb_terminate <ip:port> <challenge>
 	char buf[512];
 	int len = Q_snprintf( buf, sizeof( buf ), "sb_terminate %s %d", NET_AdrToString( cls.serveradr ), broker.challenge );
 
 	NET_SendPacket( NS_CLIENT, len, buf, broker.adr );
+	NET_NetadrSetType( &broker.adr, NA_UNDEFINED );
 }
 
 void SteamBroker_HandlePacket( netadr_t from, sizebuf_t *msg )


SNMetamorph:
Maybe it is better to use TCP for broker interaction? Yes, it will be more complicated from code perspective, but might be helpful in case if broker will running in LAN, and other clients such as Switch/PS Vita/smartphones will trying to reach it. Wireless connection aren't so reliable, so UDP packets could be dropped or corrupted which is not what we really want. TCP on the other hand will provide enough reliability for us.

zackgood991:
I would suggest TCP over UDP. It's actually quite easy to implement in C, not sure about Rust.

a1batross:
I don't think stream sockets matter that much, considering engine in general never uses them and there is no infrastructure to send or receive them, except in HTTP client which I still plan to drop and replace with something more sane.

>UDP packets could be dropped

With working CheckForResend the packet drop wouldn't be an issue.

>corrupted

To receive a corrupted packet on application level, it should pass the checksum check, which seems unlikely to me, unless there are some bad intentions by a third party but for that you need to run broker in untrusted network and that's a bad idea by the definition.

SNMetamorph:
> With working CheckForResend the packet drop wouldn't be an issue.

Fine then.

engine/client/cl_main.c
```diff
@@ -1075,6 +1068,40 @@ static void CL_WriteSteamTicket( sizebuf_t *send )
...
+
+	protinfo[0] = 0;
+
+	Info_SetValueForKey( protinfo, "prot", "3", sizeof( protinfo )); // steam auth type
+	Info_SetValueForKeyf( protinfo, "unique", sizeof( protinfo ), "%i", 0xffffffff );
+	Info_SetValueForKey( protinfo, "raw", "steam", sizeof( protinfo ));
+	CL_GetCDKey( protinfo, sizeof( protinfo ));
+	name = Info_ValueForKey( cls.userinfo, "name" );
+	if( cl_advertise_engine_in_name.value && Q_strnicmp( name, "[Xash3D]", 8 ))
+		Info_SetValueForKeyf( cls.userinfo, "name", sizeof( cls.userinfo ), "[Xash3D]%s", name );
```
zackgood991, line 1086: Should probably respect `cl_advertise_engine_in_name`.

EDIT: oops, FF didn't render that line properly

a1batross:
Attended last GoldSrc Saturday game night in Cold Ice Remastered v1, with steam-broker running at my local machine.

Didn't notice any hiccups. It just works.

SNMetamorph:
Is it ready to merge? Or maybe something are still in progress?

ilovesodaa:
so uhhh how do we use this?

PR #2448, opened 2026-01-31
Add files for cmake builds

hazardousfirmware:
This PR introduces cmake as a build alternative.

Currently only linux x64 has been tested, there is future scope for introducing other platforms.

Existing waf remains as is.

### DIFF
--- a/engine/CMakeLists.txt
+++ b/engine/CMakeLists.txt
@@ -8,7 +8,7 @@ pkg_check_modules(OPUS REQUIRED opus)
 pkg_check_modules(OPUSFILE REQUIRED opusfile)
 pkg_check_modules(VORBIS REQUIRED vorbis)
 pkg_check_modules(VORBISFILE REQUIRED vorbisfile)
-pkg_check_modules(MPG123 REQUIRED libmpg123)
+#pkg_check_modules(MPG123 REQUIRED libmpg123)
 
 if(ENABLE_FFMPEG)
 	pkg_check_modules(FFMPEG REQUIRED IMPORTED_TARGET
@@ -58,7 +58,7 @@ file(GLOB ENGINE_FILES
 	platform/sdl2/*.c
 	client/*.c
 	client/soundlib/*.c
-	# client/soundlib/libmpg/*.c # use system libmpg123 instead
+	client/soundlib/libmpg/*.c # TODO: use system libmpg123 instead
 	server/*.c
 )
 
@@ -89,7 +89,7 @@ target_include_directories(xash PUBLIC
 	${OPUS_INCLUDE_DIRS}
 	${OPUSFILE_INCLUDE_DIRS}
 	${VORBIS_INCLUDE_FILES}
-	${MPG123_INCLUDE_DIRS}
+#	${MPG123_INCLUDE_DIRS}
 )
 
 target_link_libraries(xash
@@ -106,7 +106,7 @@ target_link_libraries(xash
 	${VORBISFILE_LIBRARIES}
 	${OPUSFILE_LIBRARIES}
 	${FFMPEG_LIBRARIES}
-	${MPG123_LIBRARIES}
+#	${MPG123_LIBRARIES}
 )
 
 file(GLOB SERVER_FILES


a1batross:
What's the reason of reintroducing cmake?

I would rather not support something like that.

engine/client/avi/avi_ffmpeg.h
```diff
@@ -37,6 +37,26 @@ GNU General Public License for more details.
 #define SUPPORTED_SWR_VERSION_MAJOR 5
 #define SUPPORTED_SWS_VERSION_MAJOR 8
 
+#ifdef ALLOW_UNSUPPORTED_FFMPEG
```
a1batross, line 40: This is a bad idea, and why it's bad is explained at lines 29-33 in this file.

If you don't want to trigger this checks, don't build with `XASH_FFMPEG_DLOPEN`.

engine/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+		${BACKTRACE_INCLUDE_DIRS}
+	)
+endif()
+
+# Game client library
+file(GLOB ENGINE_FILES
+	common/*.c
+	common/imagelib/*.c
+	common/soundlib/*.c
+	platform/linux/*.c
```
a1batross, line 56: If you specifically making it for linux, I would recommend keeping it out from the main tree to not create confusion. A folder called `contrib` is often used for something contributed, and not directly supported or polished enough.

engine/CMakeLists.txt
```diff
@@ -0,0 +1,153 @@
...
+file(GLOB SERVER_FILES
+	server/*.c
+	common/*.c
+	common/imagelib/*.c
+	common/soundlib/*.c
+	platform/linux/*.c
+	platform/posix/*.c
+)
+
+add_executable(xashded
```
a1batross, line 121: It's called just `xash`.

ref/gl/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+include_directories(
+	./
+)
+
+file(GLOB REF_GL_FILES
+	*.c
+	gl2_shim/*.c
+)
+
+# Find OpenGL
```
a1batross, line 10: You don't need to do that. OpenGL is linked dynamically by default and we don't use system headers as all definitions already exist in `gl_export.h`.

CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
+# Cmake project for xash3d. Currently only Linux amd64 is supported, other systems might work but haven't been tested
```
a1batross, line 1: Considering you directly `linux` platform folder, no they won't.

CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+
+cmake_minimum_required(VERSION 3.30)
+
+project(xash3d-fwgs
+	LANGUAGES CXX C
+)
+
+set(CMAKE_CXX_STANDARD 23)
```
a1batross, line 9: There is no requirement for C++23 yet.

`mainui_cpp` targets C++11/14.

CMakeLists.txt
```diff
@@ -0,0 +1,138 @@
...
+		else()
+			message(STATUS "Not found: ${flag}")
+		endif()
+
+		# Clean up the test file
+		file(REMOVE("${CMAKE_BINARY_DIR}/test.cpp"))
+	endfunction()
+
+	# List of CPU optimizations to check
+	set(OPTIMIZATION_FLAGS
```
a1batross, line 56: Not a great idea to do this by default, you're only making the result binary depend on your build machine.

Not only that, it's not checking for your machine either, it just checks whether the simple C++ program above builds with it, and the compiler wouldn't complain if your CPU doesn't support the extension.

engine/client/avi/avi_ffmpeg.h
hazardousfirmware: Its enabled by default when enabling avi ffmpeg in waf.

engine/CMakeLists.txt
hazardousfirmware: yes, same target not allowed in cmake

ref/gl/CMakeLists.txt
hazardousfirmware: will fix

CMakeLists.txt
hazardousfirmware: will fix

CMakeLists.txt
hazardousfirmware: will remove

hazardousfirmware:
> What's the reason of reintroducing cmake?
> 
> I would rather not support something like that.

For now will cancel this PR. Thanks for the feedback, will incorporate into the fork.

PR #2451, opened 2026-02-06
avi: add audio context check before sending packet

Razcoina:
Fix a crash in AVI_Think when playing videos without an audio stream (e.g. logo.avi).


Buggem:
LGTM!

Razcoina:
Required for [mainui_cpp PR #144](https://github.com/FWGS/mainui_cpp/pull/144)

Issue #9, opened 2018-11-09
Haiku port


a1batross:

Vladislav4KZ:
[#478](https://github.com/FWGS/xash3d-fwgs/pull/478)

a1batross:
Can be closed I guess. 

Hey @exstrim401 wouldn't you mind maintaining Haiku port? See Documentation/ports.md.

oooo-ps:
Guys, is there any installation guide or just official build for this OS?

a1batross:
There isn't, you have to build it from the source code yourself. Follow build instructions for other platforms from README.md, they should be similar.

If you figure out the exact steps, we can add them to README.md.

oooo-ps:
Okay, just found [this video](https://www.reddit.com/r/opensourcegames/comments/ra4lan/playing_halflife_on_haiku_using_a_foss_halflife/) and it looks like package is available in repository; maybe unofficial so.

a1batross:
It's definitely unofficial, we don't publish engine in any repository for now. The only real official place is GitHub Releases page, if it's not there, then it's not ours.

But engine is available publicly, you're free to do whatever you want.

oooo-ps:
Well, I made following steps:

``` bash
pkgman install libsdl2_devel freetype_devel opus_devel bzip2_devel libvorbis_devel opusfile_devel libogg_devel
./waf configure
./waf build
```
and got this error:
``` bash
[388/457] Processing engine/common/whereami.c
../engine/common/whereami.c:860:2: error: #error unsupported platform
  860 | #error unsupported platform
      |  ^~~~~
```

Currently, I'm running Haiku 64 bit, so changed config to `-8`:
``` bash
./waf configure -8
```

I believe that's unexpected condition in `whereami.c:860:2` - ~could somebody please check this file?~ untested yet but should work by PR #2459

Now I have next issue:
``` bash
[451/457] Processing engine/client/soundlib/snd_ogg_vorbis.c
../engine/client/soundlib/snd_ogg_vorbis.c:50:9: error: initialization of 'int (*)(void *, ogg_int64_t,  int)' {aka 'int (*)(void *, long long int,  int)'} from incompatible pointer type 'int (*)(void *, int64_t,  int)' {aka 'int (*)(void *, long int,  int)'} [-Werror=incompatible-pointer-types]
   50 |         OggFilestream_Seek,
      |         ^~~~~~~~~~~~~~~~~~
[...]

PR #2459, opened 2026-02-16
add missed haiku declaration

oooo-ps:
#9

### DIFF
--- a/engine/common/whereami.c
+++ b/engine/common/whereami.c
@@ -12,7 +12,7 @@
 extern "C" {
 #endif
 
-#if defined(__linux__) || defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


a1batross:
Does Haiku provide /proc/self/exe?

oooo-ps:
Seems that nope. Is this critical? Because I've successfully built and launched the game (including network) but it lags, maybe the issue with video drivers..

p.s. there is yet another problem found, with vorbis/ogg data types ([temporarily solved with cast](https://github.com/FWGS/xash3d-fwgs/issues/9#issuecomment-3910416388)) - just not ready to include in this PR context.

a1batross:
This feature is required for an ability to restart the engine into another game.

There seem to be a forum post discussing it: https://discuss.haiku-os.org/t/equivalent-to-proc-self-exe/14343

PR #2482, opened 2026-03-09
Broker messages changes and additions

SNMetamorph:


engine/client/cl_steam.c
```diff
@@ -41,15 +41,22 @@ static struct
...
-		}
+	}
+	return true;
+}
+
+qboolean SteamBroker_InitiateGameConnection( netadr_t serveradr, int challenge )
+{
+	if( !SteamBroker_UpdateBrokerAddress( ))
+	{
+		Con_Printf( "%s: failed to parse broker address \"%s\"\n", __func__, cl_steam_broker_addr.string );
```
a1batross: It can be parse error, sure, but it also can be DNS resolving error.

`SteamBroker_InitiateGameConnection: failed to parse broker address "ame.lan"` sounds a bit misleading, if the only thing I did is disconnected from my home network.

engine/client/cl_steam.c
```diff
@@ -80,14 +87,50 @@ void SteamBroker_TerminateGameConnection( void )
 	if( cls.legacymode != PROTO_GOLDSRC )
 		return;
 
-	// sb_terminate <ip:port> <challenge>
+	// sb_disconnect <ip:port> <challenge>
 	char buf[512];
-	int len = Q_snprintf( buf, sizeof( buf ), "sb_terminate %s %d", NET_AdrToString( cls.serveradr ), broker.challenge );
+	int len = Q_snprintf( buf, sizeof( buf ), "sb_disconnect %s %d", NET_AdrToString( cls.serveradr ), broker.challenge );
```
a1batross, line 84: Initially I called it `sb_terminate` because under the hood it calls legacy `TerminateGameConnection` SteamAPI call.

It's fine by me if you change it, just FYI.

engine/client/cl_steam.c
```diff
@@ -80,14 +87,50 @@ void SteamBroker_TerminateGameConnection( void )
...
+		return;
+
+	if( !SteamBroker_UpdateBrokerAddress( ))
+	{
+		Con_Printf( "%s: failed to parse broker address \"%s\"\n", __func__, cl_steam_broker_addr.string );
+		return;
+	}
+
+	char buf[512];
+	int len = Q_snprintf( buf, sizeof( buf ), "sb_terminate" );
```
a1batross: Could be:
`NET_SendPacket( NS_CLIENT, sizeof( "sb_terminate" ) - 1, "sb_terminate", broker.adr );`

engine/client/cl_steam.c
```diff
@@ -80,14 +87,50 @@ void SteamBroker_TerminateGameConnection( void )
...
+	if( !SteamBroker_UpdateBrokerAddress( ))
+	{
+		Con_Printf( "%s: failed to parse broker address \"%s\"\n", __func__, cl_steam_broker_addr.string );
+		return;
+	}
+
+	NET_Config( true, true ); // initialize sockets to be able to send packets to broker
+
+	char buf[512];
+	int len = Q_snprintf( buf, sizeof( buf ), "sb_gamedir %s", gamedir );
```
a1batross, line 97: Do you think we should handle different AppIDs on steam-broker level?

The thing is, there is client function `pfnGetAppID` which does return current game AppID. But I don't think anybody except maybe original CS client.dll uses it, so it's very minor.

SNMetamorph: It is not used only for this, but also for initializing Steam API exactly when game was started, not when broken was started. What about AppID handling, it's worth to try to auto-generate steam_appid.txt depending on gamedir name, but I didn't tested it yet.

engine/client/cl_steam.c
SNMetamorph: Replace it to "failed to resolve broker address" then?

engine/client/cl_steam.c
SNMetamorph: `sb_connect <ip:port> <server's steam id> <secure> <challenge>` - used for obtaining auth ticket during connection to GoldSrc server
`sb_disconnect <ip:port> <challenge>` - used for signaling broker about disconnecting from GoldSrc server
`sb_gamedir <gamedir>` - used for signaling broker about game startup and for announcing started mod, so broker could choose proper AppID for Steam API initialization
`sb_terminate` - used for signaling broker about game shutdown, ideally broker should be terminated too (and later restarted automatically and waiting for `sb_gamedir` message)

engine/client/cl_steam.c
a1batross: Yes, that would be fine.

engine/client/cl_steam.c
a1batross: Can you put them somewhere? In `Documentation` or in `cl_steam.c` file at least?

engine/common/host.c
```diff
@@ -1181,6 +1181,10 @@ int EXPORT Host_Main( int argc, char **argv, const char *progname, int bChangeGa
 	Cbuf_ExecStuffCmds();	// execute stuffcmds (commandline)
 	SCR_CheckStartupVids();	// must be last
 
+#if !XASH_DEDICATED
+	SteamBroker_AnnounceGameStart( GI->gamefolder );
```
a1batross: This could also be moved to `CL_Init`, after client.dll successfully loads.

SNMetamorph: As far as I remember CL_Init is too early for that, because at that point config.cfg are not loaded and therefore we can't get proper values for `cl_ticket_generator` and `cl_steam_broker_addr`

a1batross: Maybe then call it in the main loop, so that you can also catch these cvars changes?

SNMetamorph: > Maybe then call it in the main loop, so that you can also catch these cvars changes?

Maybe, but how to do this better? I don't have ideas. We need to send this message only once.

a1batross: something like this:

```c
void SteamBroker_Frame( void )
{
	qboolean restart = FBitSet( cl_steam_broker_addr.flags|cl_ticket_generator.flags, FCVAR_CHANGED );

	if( !restart || !Q_stricmp( cl_ticket_generator.string, "steam" ))
		return;

	// terminate old steam broker if instance address has changed
	...

	SteamBroker_AnnounceGameStart( GI->gamefolder );

	ClearBits( cl_ticket_generator.flags, FCVAR_CHANGED );
	ClearBits( cl_steam_broker_addr.flags, FCVAR_CHANGED );
}
```

And then put `SteamBroker_Frame` somewhere in client mainloop, like Host_ClientBegin maybe.

This way we can notify steam broker instantly once the address or ticket generator cvars has been changed, without the need to restart the engine just to notify steam broker.

SNMetamorph: Ok, I did something like this example.

a1batross:
Oh no, I merged it and it will break current steam broker. :(

I wouldn't be able to work on `steam-broker` side until Tuesday next week.

SNMetamorph:
Also I updated documentation, but this will be in another PR

PR #2495, opened 2026-03-19
More iOS support improvements

ksagameng2:
Most significant changes are that createipa.sh can now sign the binaries with entitlements specified in entitlements.plist (the ones that are currently set are for satisfying ios when the user directly installs the ipa), and the engine will now look for dylibs that end in _(gamedir).dylib and try to load them first, .dsym generation is disabled because some signers still have trouble with them.

### DIFF
--- a/engine/platform/ios/lib_ios.c
+++ b/engine/platform/ios/lib_ios.c
@@ -62,7 +62,8 @@ void *IOS_LoadLibrary( const char *dllname )
 		else postfix = "";
 	}
 
-	Q_strncpy( strippedname, dllname, strlen(dllname) - EXT_LENGTH );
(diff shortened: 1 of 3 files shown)


engine/platform/ios/lib_ios.c
```diff
@@ -14,58 +14,61 @@ GNU General Public License for more details.
...
+	if( !postfix )
+	{
+		if ( Q_strcmp(FS_Gamedir(), "valve" ) )
+		{
+			postfix = FS_Gamedir( );
+		}
+		else postfix = "";
+	}
+
+	Q_strncpy( strippedname, dllname, strlen(dllname) - EXT_LENGTH );
```
a1batross: There is `COM_StripExtension` function for that.

a1batross:
Good job, thanks!

PR #2514, opened 2026-03-30
engine: client: avi: remove XASH_FFMPEG_DLOPEN

RevengeRip:
Is this a mistake? This define effectively forces `--enable-ffmpeg-dlopen` at all times ignoring configuration (which in turn makes build to fail with ffmpeg 8.1).


a1batross:
Yeah, definitely a mistake. Thanks, merging it.

PR #2527, opened 2026-04-07
Add gyroscope aiming support & standardize input functions to use (pitch, yaw) parameter order

Vladislav4KZ:
### New Gyroscope CVars:
**joy_gyro_pitch** (default ``1.0``): Vertical sensitivity (tilting controller up/down).
**joy_gyro_yaw** (default ``1.0``): Horizontal sensitivity for "laser pointer" style rotation (turning controller left/right).
**joy_gyro_roll** (default ``0.0``): Horizontal sensitivity for "steering wheel" style rotation (tilting controller sideways).

Also brought the axes order to a unified standard format.
In some functions it was ``yaw, pitch``, and in others ``pitch, yaw``.
There was an issue in ``input.h`` with this function:
```
void Joy_FinalizeMove(float *fw, float *side, float *dpitch, float *dyaw);
```

But in ``input.c`` the axes arguments were swapped:
```
Joy_FinalizeMove(forward, side, yaw, pitch);
```


SNMetamorph:
That's cool feature, I'll test it on PS Vita when it will be done.

engine/client/in_joy.c
```diff
@@ -317,8 +323,19 @@ void Joy_FinalizeMove( float *fw, float *side, float *dpitch, float *dyaw )
...
+	if( joy_have_gyro.value )
+	{
+		float pitch_speed = joy_gyro_speed[0] * ( 180.0f / M_PI );
+		float yaw_speed   = joy_gyro_speed[1] * ( 180.0f / M_PI );
+		float roll_speed  = joy_gyro_speed[2] * ( 180.0f / M_PI );
+
+		*dpitch -= joy_gyro_pitch.value * pitch_speed * host.realframetime;
+		*dyaw   += joy_gyro_yaw.value   * yaw_speed   * host.realframetime;
+		*dyaw   += joy_gyro_roll.value  * roll_speed  * host.realframetime;
+	}
```
a1batross, line 343: `VectorClear( joy_gyro_speed )`

engine/client/in_joy.c
```diff
@@ -317,8 +323,19 @@ void Joy_FinalizeMove( float *fw, float *side, float *dpitch, float *dyaw )
 
 	*fw     -= joy_forward.value * (float)joyaxis[JOY_AXIS_FWD ].val/(float)SHRT_MAX;  // must be form -1.0 to 1.0
 	*side   += joy_side.value    * (float)joyaxis[JOY_AXIS_SIDE].val/(float)SHRT_MAX;
-	*dpitch -= joy_pitch.value * (float)joyaxis[JOY_AXIS_PITCH].val/(float)SHRT_MAX * host.realframetime;
-	*dyaw   += joy_yaw.value   * (float)joyaxis[JOY_AXIS_YAW  ].val/(float)SHRT_MAX * host.realframetime;
+	*dpitch += joy_pitch.value * (float)joyaxis[JOY_AXIS_PITCH].val/(float)SHRT_MAX * host.realframetime;
+	*dyaw   -= joy_yaw.value   * (float)joyaxis[JOY_AXIS_YAW  ].val/(float)SHRT_MAX * host.realframetime;
+
+	if( joy_have_gyro.value )
```
a1batross: Let's guard it with `joy_calibrated.value` too.

a1batross (changes requested):

engine/client/in_joy.c
a1batross: We have to do this because gyro events are sent asynchronously from platform code and if for some reason they are skipped, this will cause a small drift every frame.

a1batross:
Thanks, out of excitement, I made all the changes myself. :)

Works on my PS5 controller, I will also check it with Steam Controller and Steam Deck, and if it's working OK, will merge it.

a1batross:
It does indeed work with Steam Controller, apparently something broke in SDL2, but it does work with SDL3 (over sdl2-compat).

Merging. Also I wrote some debugging tool that will become an aid in the future.

PR #2531, opened 2026-04-08
engine: client: in_joy: implement per-axis gyroscope deadzone support

Vladislav4KZ:
### New Gyroscope Deadzone CVars
**joy_gyro_pitch_deadzone** (default ``0.5``): Deadzone for the pitch axis, in deg/s.
**joy_gyro_yaw_deadzone** (default ``0.5``): Deadzone for the yaw axis, in deg/s.
**joy_gyro_roll_deadzone** (default ``0.5``): Deadzone for the roll axis, in deg/s.

By the way, I'm not sure where the default gyroscope deadzone should be defined -- as a constant in ``defaults.h`` (like the stick deadzone), or directly in the cvar definitions?


a1batross:
>directly in the cvar definitions

Yeah, that's fine. `default.h` is usually for what could be platform specific or for build tailored to specific needs.

Can you add deadzone bars to the `Joy_DrawDebug`? The code should be similar to what it already does for non-trigger axes.

Vladislav4KZ:
> Can you add deadzone bars to the Joy_DrawDebug? The code should be similar to what it already does for non-trigger axes.

Done.

a1batross (approved):

PR #2532, opened 2026-04-08
fix eliminate blue halo around sprites

rootcreep:
also applies to #2342

### DIFF
--- a/ref/gl/gl_image.c
+++ b/ref/gl/gl_image.c
@@ -1238,6 +1238,8 @@ static void GL_ProcessImage( gl_texture_t *tex, rgbdata_t *pic )
 	// force upload texture as RGB or RGBA (detail textures requires this)
 	if( tex->flags & TF_FORCE_COLOR ) pic->flags |= IMAGE_HAS_COLOR;
 	if( pic->flags & IMAGE_HAS_ALPHA ) tex->flags |= TF_HAS_ALPHA;
+	if( FBitSet( pic->flags, IMAGE_PREMULTIPLIED ))
+		SetBits( tex->flags, TF_PREMULTIPLIED );
 
 	tex->encode = pic->encode; // share encode method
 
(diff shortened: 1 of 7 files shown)


ref/gl/gl_context.c
```diff
@@ -159,7 +159,7 @@ static qboolean Mod_ProcessRenderData( model_t *mod, qboolean create, const byte
 		loaded = true;
 		break;
 	case mod_sprite:
-		Mod_LoadSpriteModel( mod, buf, &loaded, mod->numtexinfo );
+		Mod_LoadSpriteModel( mod, buf, &loaded, TF_NOMIPMAP | TF_CLAMP | TF_ALLOW_NEAREST );
```
a1batross: This will break sprites loaded through `pfnSPR_LoadExt`.

ref/gl/gl_draw.c
```diff
@@ -64,6 +64,16 @@ R_DrawStretchPic
 */
 void R_DrawStretchPic( float x, float y, float w, float h, float s1, float t1, float s2, float t2, int texnum )
 {
+	gl_texture_t	*glt = R_GetTexture( texnum );
+
+	if( FBitSet( glt->flags, TF_PREMULTIPLIED ) && pglIsEnabled( GL_BLEND ))
+	{
+		GLint	dst = 0;
```
a1batross: Why does this change blending mode? What's the reason of reading dst blending?

ref/gl/gl_image.c
```diff
@@ -1275,6 +1277,23 @@ static void GL_ProcessImage( gl_texture_t *tex, rgbdata_t *pic )
 		// processing image before uploading (force to rgba, make luma etc)
 		if( pic->buffer ) gEngfuncs.Image_Process( &pic, 0, 0, img_flags, 0 );
 
+		if( pic->buffer && pic->type == PF_RGBA_32 && FBitSet( pic->flags, IMAGE_HAS_ALPHA )
+		    && !FBitSet( pic->flags, IMAGE_PREMULTIPLIED ) && tex->name[0] && Q_stristr( tex->name, "logo" ))
```
a1batross: If that's only for custom logo, I think Q_stristr could be replaced with Q_strncmp, it should be something like `Q_strncmp( tex->name, "#logo.", 6 )`

a1batross (commented):
Does this only applies to logos or just textures in general? If so, can you show any examples? I don't really remember if there is any sprites that have this issue...

a1batross:
> Does this only applies to logos or just textures in general? If so, can you show any examples? I don't really remember if there is any sprites that have this issue...

Yep, doesn't seem to be sprite related with the latest commit.

a1batross (approved):
I will test it later today, hopefully, and will merge it if everything is fine.

PR #2540, opened 2026-04-14
Implement built-in gyroscope support for SDL2 platforms

Vladislav4KZ:
Introduce native support for mobile gyroscope sensors using the **SDL2 Sensor API**, enabling player view and aiming control with the built-in gyroscope on Android, iOS, and other SDL2-compatible platforms.

The implementation is split into platform-agnostic logic in ``engine/client/in_gyro.c`` and a platform-specific sensor layer in ``engine/platform/sdl2/sensor_sdl2.c``.
Axis mapping is pre-configured for landscape orientation, typical for mobile gaming.

### New CVars
**gyro_enable**: Master toggle for the built-in gyroscope.
**gyro_available** (Read-only): Hardware status indicator.
**gyro_pitch, gyro_yaw, gyro_roll**: Individual axis sensitivity.
**gyro_pitch_deadzone, gyro_yaw_deadzone, gyro_roll_deadzone**: Configurable thresholds to eliminate sensor jitter and noise.

### DIFF
--- a/engine/client/in_gyro.c
+++ b/engine/client/in_gyro.c
@@ -1,0 +1,120 @@
+/*
+in_gyro.c - System gyroscope input code
+Copyright (C) 2026 Xash3D FWGS contributors
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+*/
+
+#include "common.h"
+#include "input.h"
+#include "client.h"
+
+static CVAR_DEFINE_AUTO( gyro_enable, "0", FCVAR_ARCHIVE | FCVAR_FILTERABLE, "enables aiming with built-in device gyroscope" );
+static CVAR_DEFINE_AUTO( gyro_available, "0", FCVAR_READ_ONLY, "tells whether system gyroscope hardware is available or not" );
+static CVAR_DEFINE_AUTO( gyro_pitch, "1.0", FCVAR_ARCHIVE | FCVAR_FILTERABLE, "built-in gyroscope sensitivity for looking up and down" );
+static CVAR_DEFINE_AUTO( gyro_yaw, "1.0", FCVAR_ARCHIVE | FCVAR_FILTERABLE, "built-in gyroscope sensitivity for turning left and right" );
+static CVAR_DEFINE_AUTO( gyro_roll, "0.0", FCVAR_ARCHIVE | FCVAR_FILTERABLE, "built-in gyroscope sensitivity when tilting the device sideways" );
+static CVAR_DEFINE_AUTO( gyro_pitch_deadzone, "0.5", FCVAR_ARCHIVE | FCVAR_FILTERABLE, "built-in gyroscope pitch axis deadzone (deg/s)" );
+static CVAR_DEFINE_AUTO( gyro_yaw_deadzone, "0.5", FCVAR_ARCHIVE | FCVAR_FILTERABLE, "built-in gyroscope yaw axis deadzone (deg/s)" );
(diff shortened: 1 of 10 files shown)


engine/client/in_gyro.c
```diff
@@ -0,0 +1,113 @@
...
+
+Apply gyro movement to view angles
+=============
+*/
+void IN_GyroFinalizeMove( float *fw, float *side, float *dpitch, float *dyaw )
+{
+	if( !cl_gyro.value || !cl_gyro_available.value )
+		return;
+
+	// In Landscape mode axes are swapped relative to natural (Portrait) orientation
```
a1batross, line 101: Does it handle or needs to handle both left and right landscape modes?

engine/platform/sdl2/sensor_sdl2.c
```diff
@@ -0,0 +1,103 @@
...
+==============
+SDLash_SensorUpdate
+
+==============
+*/
+void SDLash_SensorUpdate( SDL_SensorEvent sensor )
+{
+	if( sensor.which == g_system_gyro_id )
+	{
+		IN_GyroEvent( sensor.data );
```
a1batross, line 96: Doesn't system gyroscope require calibration to get rid of noise or sudden angle changes?

a1batross (commented):
Seems mostly based on already existing gyro code but SDL2 handles system and joystick gyros separately, so maybe it's fine for now.

engine/client/in_gyro.c
Vladislav4KZ: I overlooked this part. The axes are currently mapped for the default landscape orientation, so when the device is rotated, the camera control becomes inverted.

SDL can detect the display orientation, so it’s possible to map the axes for the inverted landscape as well.

I’ll look into this later.

a1batross: I think it can handle all possible rotations, to not make any assumptions whether reported portrait mode is actually portait mode.

engine/platform/sdl2/sensor_sdl2.c
Vladislav4KZ: On Android, additional calibration is generally not required because the OS performs internal sensor fusion. The [`SDL_SENSOR_GYRO`](https://github.com/libsdl-org/SDL/blob/725db83c2392083c7e92301e9d855aba0785d507/src/sensor/android/SDL_androidsensor.c#L105-L106) maps to [`TYPE_GYROSCOPE`](https://developer.android.com/reference/android/hardware/Sensor#TYPE_GYROSCOPE), which uses accelerometer and magnetometer data in the background (Sensor HAL) to apply drift compensation.

On iOS, however, SDL uses raw [`CMGyroData`](https://github.com/libsdl-org/SDL/blob/725db83c2392083c7e92301e9d855aba0785d507/src/sensor/coremotion/SDL_coremotionsensor.m#L159-L173) [(CMGyroData - Apple Developer Documentation)](https://developer.apple.com/documentation/coremotion/cmgyrodata) instead of the fused [`CMDeviceMotion.rotationRate`](https://developer.apple.com/documentation/coremotion/cmdevicemotion/rotationrate) API.

Maybe it's worth adding a manual calibration command (e.g., `gyro_calibrate`) to calculate and compensate for static bias (offset) in real-time? Fully automatic calibration might be unreliable here, as handheld micro-movements tend to interfere with bias estimation. A manual command would allow the user to simply place the device on a stable surface for a moment to establish a perfect "zero."

**Sources:**
- [Android Sensor Types](https://source.android.com/docs/core/interaction/sensors/sensor-types)
- [Android Motion Sensors](https://developer.android.com/develop/sensors-and-location/sensors/sensors_motion)
- [CMMotionManager - Apple Developer Documentation](https://developer.apple.com/documentation/coremotion/cmmotionmanager)

a1batross: Let's keep it as is for now, if anybody will complain, we will add calibration.

I think calibration code can be shared with joystick code, it's pretty much platform-independent, and also doesn't care where the sensor is attached to.

engine/client/in_gyro.c
Vladislav4KZ: Done.

Vladislav4KZ:
Renamed the **cl_gyro** cvar to **gyro_enable**, and updated other cvars starting with **cl_gyro_** to **gyro_**, following the naming convention of **touch_** and **joy_** cvars

Vladislav4KZ:
I think everything is ready to be merged now.

aomarov-c2 (approved):

a1batross (approved):

PR #1466, opened 2023-10-22
Big endian support for BSP loader

Xav101:
This set of patches seems like it's enough to get the game loading BSPs on big endian platforms. Tested on big endian mips-irix using half life 1 maps. This does not have support for extra lumps and doesn't support the QBSP2 files.

I probably missed something in here because there's quite a few fields in a BSP so please let me know if you notice anything.

PR #2562, opened 2026-04-23
Big endian support (dedicated server only)

a1batross:
The idea is to pre-swap as much as possible before it hits game code, to make porting much easier. For now it only focuses to flesh out dedicated server path, and with this I'm able to successfully connect to a server running on PPC32 machine, hosting Half-Life Deathmatch built from master branch of hlsdk-portable.

It also does Xash extensions, but those aren't tested.

Supersedes #1466 

cc @Xav101 @IntriguingTiles


IntriguingTiles:
Very nice. Looks much better than my approach. FYI, Valve made two mistakes in `delta.lst` that they get away with on little endian.

- `skin` in `custom_entity_state_t` claims to be `DT_INTEGER` but should really be `DT_SHORT`. This causes certain `env_beam` configurations, like some in the SDK example map `env_beam.rmf`, to point to the world origin instead of the end point entity (although I just tested it now on an LE client connecting to a BE server and these beams were invisible instead).
- `eflags` in `entity_state_t` also claims to be `DT_INTEGER` but should really be `DT_BYTE`. Not sure what problems this causes, if any.

[My approach](https://github.com/IntriguingTiles/xash3d-fwgs/commit/36f9366ffe4e8324891e3f11bec06823652e9432) to fixing this was to detect incorrect size flags and fix them up, which works, but it would probably be better to completely ignore size flags from `delta.lst` and instead use the size info already present in `delta_field_t`.

a1batross:
>skin in custom_entity_state_t claims to be DT_INTEGER but should really be DT_SHORT. This causes certain env_beam configurations, like some in the SDK example map env_beam.rmf, to point to the world origin instead of the end point entity (although I just tested it now on an LE client connecting to a BE server and these beams were invisible instead).

I think that's what I already noticed in mass spectrometer chamber, there is a beam that goes to 0 0 0.

>eflags in entity_state_t also claims to be DT_INTEGER but should really be DT_BYTE. Not sure what problems this causes, if any.

eflags used very rarely, so it's less of a deal.

>[My approach](https://github.com/IntriguingTiles/xash3d-fwgs/commit/36f9366ffe4e8324891e3f11bec06823652e9432) to fixing this was to detect incorrect size flags and fix them up, which works, but it would probably be better to completely ignore size flags from delta.lst and instead use the size info already present in delta_field_t.

I had an idea to validate and potentially let modders find such mistakes in delta.lst. But I think letting engine read less of data in delta.lst is generally has it's use, though I'm not sure about network compatibility between big and little endian in this case. But reading more field can provide would just lead to a memory corruption.

a1batross:
Check out https://github.com/FWGS/xash3d-fwgs/pull/2566

PR #2566, opened 2026-04-24
Big endian support (now with client)

a1batross:
Does a lot of stuff here and there, as while porting the client I found out about broken libbacktrace that was only caused by incorrect configure script running sequentially. (heh, reverse race condition)

Almost everything works as intended, except for a few known issues:
- sometimes animations get corrupted, but I wasn't been able to consistently reproduce that, so it's a known bug
- already pushed patch to hlsdk-portable that reverts byte swapping in studio model renderer previously added by @IntriguingTiles, as it's now completely handled by the engine.
- voice chat is probably broken

### DIFF
--- a/engine/common/model.c
+++ b/engine/common/model.c
@@ -614,7 +614,7 @@ void GAME_EXPORT Mod_LoadCacheFile( const char *filename, cache_user_t *cu )
 #if XASH_BIG_ENDIAN
 	if( size >= sizeof( int ) && LittleLong( IDSEQGRPHEADER ) == *(uint *)cu->data )
 	{
-		studiohdr_t *phdr = (studiohdr_t *)(intptr_t)REF_GET_PARM( PARM_GET_STUDIO_HDR, 0 );
+		studiohdr_t *phdr = (studiohdr_t *)REF_GET_PARM( PARM_GET_STUDIO_HDR, 0 );
 		if( !phdr )
 			return;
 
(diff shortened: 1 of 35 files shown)


engine/common/model.c
```diff
@@ -608,6 +608,31 @@ void GAME_EXPORT Mod_LoadCacheFile( const char *filename, cache_user_t *cu )
 	cu->data = Mem_Malloc( com_studiocache, size );
 	memcpy( cu->data, buf, size );
 	Mem_Free( buf );
+
+	// this handles when studio model renderer tries to load sequence files on it's own
+	// which is what they always do in HLSDK
+#if XASH_BIG_ENDIAN
+	if( size >= sizeof( int ) && LittleLong( IDSEQGRPHEADER ) == *(uint *)cu->data )
+	{
+		studiohdr_t *phdr = (studiohdr_t *)(intptr_t)REF_GET_PARM( PARM_GET_STUDIO_HDR, 0 );
```
IntriguingTiles: `REF_GET_PARM` returns a 32-bit integer, so this won't work on 64-bit archs, and indeed, I see a crash here on s390x.

a1batross: Good point.

I changed it to intptr_t a while ago for EngineGetParm as engine passes pointers to renderers that way, and @SNMetamorph asked to do this for RenderGetParm. As it's a common pattern, I thought I already did it :D

a1batross: Fixed.

PR #2572, opened 2026-04-29
New implementation of Steam broker interaction code

SNMetamorph:


engine/common/net_ws.c
```diff
@@ -150,6 +150,19 @@ static inline qboolean NET_IsSocketValid( int socket )
 #endif
 }
 
+qboolean NET_MakeSocketNonBlocking( int socket_fd )
+{
+	u_long flag = 1;
+#if XASH_WIN32 || XASH_PSVITA
+	if( ioctlsocket( socket_fd, FIONBIO, &flag ) == SOCKET_ERROR )
```
a1batross: NET_IsSocketError

a1batross: Look at what also wants non-blocking sockets and use that function there.

`u_long flag` must be declared in `XASH_WIN32 || XASH_PSVITA` block.

Also where `u_long` did come from?

engine/client/cl_steam.c
```diff
@@ -31,16 +33,52 @@ GNU General Public License for more details.
...
 	int challenge;
 	netadr_t serveradr;
-	qboolean announced;
-	qboolean addr_initialized;
-} broker;
+	double connection_timeout;
+	double idle_cycle_timeout;
+	uint8_t rx_buffer[SBRK_MAX_FRAME_SIZE + 64];
+	uint8_t tx_buffer[SBRK_MAX_FRAME_SIZE + 64];
+	uint32_t rx_buffer_pos;
```
a1batross, line 69: Why not use `sizebuf_t` to track the position? It does also checks for overflows.

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+{
+	SteamBroker_CloseSocket();
+	SteamBroker_SetState( SBRK_STATE_IDLE );
+}
+
+static qboolean SteamBroker_ConnectImpl( void )
+{
+	int result;
+	int addr_family;
+	u_long nonblocking = 1;
```
a1batross: Seems an unused variable.

a1batross: Also, declare variable where they used. Variable declaration at the start of the block is not enforced anymore.

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	{
+		addr_family = AF_INET;
+	}
+	else if( NET_NetadrType( &broker.adr ) == NA_IP6 )
+	{
+		addr_family = AF_INET6;
+	}
+	else
+	{
+		Con_Printf( S_ERROR "%s: unsupported broker address type\n", __func__ );
```
a1batross: Would be nice to print `cl_steam_broker_addr` cvar string

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+		return false;
+	}
 
-	// sb_connect <ip:port> <server's steam id> <secure> <challenge>
-	// the message calls
-	char buf[512];
-	int len = Q_snprintf( buf, sizeof( buf ), "sb_connect %s %"PRIu64" %s %d", NET_AdrToString( serveradr ), cls.server_steamid, cls.vac2_secure ? "true" : "false", broker.challenge );
+	if( !NET_MakeSocketNonBlocking( broker.socket ))
+	{
+		Con_Printf( S_ERROR "%s: failed to set non-blocking mode\n", __func__ );
```
a1batross: There is function that converts errno to string, would be nice to print the actual error.

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
-	NET_SendPacket( NS_CLIENT, len, buf, broker.adr );
+	NET_NetadrToSockadr( &broker.adr, &addr );
 
+	result = connect( broker.socket, (struct sockaddr *)&addr, NET_SockAddrLen( &addr ));
+	if( NET_IsSocketError( result ))
+	{
+		int err = WSAGetLastError();
+		if( err != WSAEWOULDBLOCK && err != WSAEALREADY && err != WSAEINPROGRESS )
+		{
+			Con_Printf( S_ERROR "%s: failed to connect to broker at %s\n", __func__, cl_steam_broker_addr.string );
```
a1batross: There is function that converts errno to string, would be nice to print the actual error.

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+			closesocket( broker.socket );
+			broker.socket = INVALID_SOCKET;
+			return false;
+		}
+	}
+
+	broker.connection_timeout = Sys_DoubleTime() + SBRK_CONNECT_TIMEOUT;
```
a1batross: In non-API code Platform_DoubleTime should be used.

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	SteamBroker_SetState( SBRK_STATE_CONNECTING );
 	return true;
 }
 
-void SteamBroker_TerminateGameConnection( void )
+static qboolean SteamBroker_SendFrame( const char *payload, uint32_t payload_size )
```
a1batross: There is no reason for `payload_size` to not be `size_t`

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
 {
-	if( NET_NetadrType( &cls.serveradr ) != NA_IP )
+	if( payload_size > SBRK_MAX_FRAME_SIZE )
```
a1batross, line 165: Isn't it should be `payload_size >= SBRK_MAX_FRAME_SIZE`?

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	{
+		Con_Printf( S_WARN "%s: payload too large (%u > %u)\n", __func__, payload_size, SBRK_MAX_FRAME_SIZE );
```
a1batross: If you make it size_t, there is %zu for it. :)

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	uint16_t length = *p | (*(p + 1) << 8);
+	p += SBRK_FRAME_LENGTH_SIZE;
+
+	uint32_t frame_size = SBRK_FRAME_HEADER_SIZE + SBRK_FRAME_LENGTH_SIZE + length;
+
+	if( broker.rx_buffer_pos < frame_size )
+		return false; // need more data
+
+	// process response if this is a sb_connect response
+	if( length >= SBRK_RESPONSE_HEADER_SIZE + 4 + 8 + 4 )
```
a1batross: What is 4 + 8 + 4 is supposed to be?

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	{
+		if( memcmp( p, SBRK_RESPONSE_HEADER, SBRK_RESPONSE_HEADER_SIZE ) == 0 )
+		{
+			uint8_t *response_data = p + SBRK_RESPONSE_HEADER_SIZE;
+			uint32_t response_size = length - SBRK_RESPONSE_HEADER_SIZE;
+
+			if( response_size >= 4 + 8 + 4 )
+			{
+				int32_t challenge = *(int32_t *)response_data;
```
a1batross: Reading data like this is prone to alignment and big endian issues. As protocol is binary now, I think sizebuf_t utils can be used for it, which do handle both endianness and alignment. Bitop stuff can be just ignored.

engine/client/cl_steam.c
```diff
@@ -52,140 +90,455 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	struct timeval tv;
+	tv.tv_sec = 0;
+	tv.tv_usec = 0;
+
+#if XASH_WIN32 == 1
+	int select_result = select( 0, NULL, &writefds, NULL, &tv );
+#else
+	int select_result = select( broker.socket + 1, NULL, &writefds, NULL, &tv );
+#endif
+	if( select_result == SOCKET_ERROR )
```
a1batross, line 407: NET_IsSocketError

a1batross (changes requested):
Is there an implementation to test it on?

engine/common/net_ws.c
SNMetamorph: > Look at what also wants non-blocking sockets and use that function there.

I guess it's better to do in separate PR. There are couple places where ioctksocket are used.

> Also where `u_long` did come from?

From ioctlsocket function definition from Windows. But I guess it's not mandatory to use this exact type.

a1batross: >I guess it's better to do in separate PR

It is going to be forgotten unless somebody notices it could be refactored.

>From ioctlsocket function definition from Windows

Let's keep it `u_long` then but it must be defined under ifdef for win32 || psvita. Not sure if PSVita build will complain though, it's not buildable right now anyway.

engine/client/cl_steam.c
SNMetamorph: Check is correct, soft limit for framze size is exactly 4096 bytes. `payload_size` with 4096 value would fit there.

engine/client/cl_steam.c
SNMetamorph: | Field | Size | Type | Description |
|-------|------|------|-------------|
| Header | 11 bytes | ASCII | String `"sb_connect\n"` |
| Challenge | 4 bytes | int32_t (LE) | Echo of challenge from request |
| SteamID | 8 bytes | uint64_t (LE) | Client SteamID from configuration |
| Size | 4 bytes | uint32_t (LE) | Length of ticket in bytes |
| Ticket | N bytes | Binary | Authentication ticket (opaque to client) |

So this is size of challenge + Steam ID + payload length. I need to refactor parsing so these checks will become unnecessary.

engine/client/cl_steam.c
SNMetamorph: Yeah, I totally agree.

SNMetamorph:
> Is there an implementation to test it on?

I have my own broker implementation specifically for Windows, and it were done in parallel with this PR. So it's well-tested and everything working perfectly.

engine/client/cl_steam.c
SNMetamorph: There is a problem with `sizebuf_t` - it cannot be used as FIFO-like data container (socket writes to RX data, other code reads data from there). Instead, it acts like LIFO-like container, and this does not suits our needs. But there is still room to apply it in profitable way.

a1batross:
> > Is there an implementation to test it on?
> 
> I have my own broker implementation specifically for Windows, and it were done in parallel with this PR. So it's well-tested and everything working perfectly.

Yeah, I know, but I wanted to test it myself. :D

engine/client/cl_steam.c
a1batross: It would be nice to document it somewhere and replace magic numbers with it, that's what I mean.

I can figure out what 4 + 8 + 4 means from the code by reading it carefully. But the point of writing good code is to minimize the effort reader has to make to understand it.

engine/common/net_ws.c
a1batross: Why you marked it as resolved?

SNMetamorph: Recheck it, now I used NET_MakeSocketNonBlocking everywhere it's possible.

a1batross: Oh, ok, F5 fixed it, good job.

Buggem:
Is this a seperate thing to a1ba's Steam broker, or do you just have two clients working with the same protocol

SNMetamorph:
> Is this a seperate thing to a1ba's Steam broker

Yes, it is written with C++ and designed to work specifically as Windows service.

Documentation/extensions/steam-broker.md
```diff
@@ -1,8 +1,139 @@
...
+└──────────────┘
+       │ 
+       ↓
+┌──────────────┐
+│    ACTIVE    │ ← Client shutting down game, terminating broker to reset Steam client state
+└──────────────┘
+       │ sb_terminate
+       ↓
+┌──────────────┐
+│  RESTARTING  │
```
a1batross, line 69: There is no such state `RESTARTING`.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	if( received == 0 )
+	{
+		Con_Printf( S_NOTE "%s: connection closed by broker\n", __func__ );
+		SteamBroker_Disconnect( );
+		return;
+	}
+
+	broker.rx_buffer_pos += received;
+
+	while( SteamBroker_ProcessFrame( ));
```
a1batross, line 339: This logic doesn't seem to be correct to me. I think the idea was to look for valid `sb_connect` in a multiple frames.

But imagine the situation where two frames arrive in one `recv` call, the first frame doesn't contain `sb_connect` for some reason and the second contains `sb_connect`. This will make `sb_connect` lost forever.

Documentation/extensions/steam-broker.md
```diff
@@ -1,8 +1,139 @@
...
+└──────────────┘
+       │ sb_gamedir
+       ↓
+┌──────────────┐
+│    ACTIVE    │ ← Session established, rejecting all of other TCP connections
+└──────────────┘
+       │ sb_connect
+       ↓
+┌──────────────┐
+│    ACTIVE    │ ← Client getting auth ticket to connect to game server
```
a1batross, line 54: There is two `ACTIVE` states?

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
-	if( restart )
+	fd_set writefds;
+	FD_ZERO( &writefds );
+	FD_SET( broker.socket, &writefds );
+
+	struct timeval tv;
+	tv.tv_sec = 0;
+	tv.tv_usec = 0;
+
+#if XASH_WIN32 == 1
```
a1batross: It's just `#if XASH_WIN32`.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	}
 
-	// sb_connect <ip:port> <server's steam id> <secure> <challenge>
-	// the message calls
-	char buf[512];
-	int len = Q_snprintf( buf, sizeof( buf ), "sb_connect %s %"PRIu64" %s %d", NET_AdrToString( serveradr ), cls.server_steamid, cls.vac2_secure ? "true" : "false", broker.challenge );
+	if( !NET_MakeSocketNonBlocking( broker.socket ))
+	{
+		Con_Printf( S_ERROR "%s: failed to set non-blocking mode, error %s\n", __func__, NET_ErrorString( ));
+		closesocket( broker.socket );
```
a1batross: You can just call `CloseSocket` here.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+	NET_NetadrToSockadr( &broker.adr, &addr );
 
+	int result = connect( broker.socket, (struct sockaddr *)&addr, NET_SockAddrLen( &addr ));
+	if( NET_IsSocketError( result ))
+	{
+		int err = WSAGetLastError();
+		if( err != WSAEWOULDBLOCK && err != WSAEALREADY && err != WSAEINPROGRESS )
+		{
+			Con_Printf( S_ERROR "%s: failed to connect to broker at %s with error %s\n", __func__, cl_steam_broker_addr.string, NET_ErrorString( ));
+			closesocket( broker.socket );
```
a1batross: You can just call `CloseSocket` here.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
-	NET_Config( true, true ); // initialize sockets to be able to send packets to broker
+	if( broker.state != SBRK_STATE_CONNECTED )
+		return;
 
+	// sb_gamedir <gamedir>
 	char buf[512];
 	int len = Q_snprintf( buf, sizeof( buf ), "sb_gamedir %s", gamedir );
 
-	NET_SendPacket( NS_CLIENT, len, buf, broker.adr );
+	SteamBroker_SendFrame( buf, len );
```
a1batross: Q_snprintf can technically return -1

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
-	int challenge;
-	uint32_t len;
-	uint8_t ticket[2048]; // 2048 bytes according to SDK docs
+	SteamBroker_HandleDataTx( );
+	SteamBroker_HandleDataRx( );
+}
 
-	if( !NET_CompareAdr( from, broker.adr ))
-		return;
+int SteamBroker_InitiateGameConnection( netadr_t serveradr, int challenge )
```
a1batross: I think it can return `qboolean`.

engine/client/cl_steam.c
a1batross: IMHO return value should always return true when frame was correctly parsed, regardless if it got sb_connect or not.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+			{
+				Con_Printf( S_ERROR "%s: challenge mismatch\n", __func__ );
+			}
+			else
+			{
+				uint64_t steam_id;
+				MSG_ReadBytes( &sb, &steam_id, sizeof( steam_id ));
+				uint32_t ticket_size = MSG_ReadDword( &sb );
+
+				if( ticket_size > SBRK_TICKET_SIZE_MAX || ticket_size < MSG_GetNumBytesLeft( &sb ))
```
a1batross: Could it be `ticket_size != MSG_GetNumBytesLeft( &sb )`?

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+#else
+	int select_result = select( broker.socket + 1, NULL, &writefds, NULL, &tv );
+#endif
+	if( select_result == SOCKET_ERROR )
 	{
-		broker.announced = false;
-		ClearBits( cl_ticket_generator.flags, FCVAR_CHANGED );
-		ClearBits( cl_steam_broker_addr.flags, FCVAR_CHANGED );
+		Con_Printf( S_ERROR "%s: select() failed\n", __func__ );
+		return;
```
a1batross, line 411: Shouldn't `SteamBroker_Disconnect` be called in all three failure branches?

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
-	if( !broker.announced )
+	if( FD_ISSET( broker.socket, &writefds ))
 	{
-		netadr_t previous_addr = broker.adr;
-		if( !SteamBroker_UpdateBrokerAddress( ))
-		{ 
-			Con_Printf( "%s: failed to resolve broker address \"%s\"\n", __func__, cl_steam_broker_addr.string );
+		// socket is writable - connection established or failed
+		int err = 0;
+		int err_len = sizeof( err );
```
a1batross: `socklen_t`.

a1batross (changes requested):

engine/client/cl_steam.c
SNMetamorph: Nope, because we could have parts of next frame remaining in RX buffer

Documentation/extensions/steam-broker.md
SNMetamorph: This diagram does not reflect current FSM state, it's just resume of "what happening now" on broker side.

a1batross:
Bug:
1. Set cl_steam_broker_addr to some valid address but where steambroker isn't being run
2. Change cl_ticket_generator to steam. Without SteamBroker_Disconnect in `connection failed with error` case, it will crash. With it will try to connect with some timeout, spamming in the console.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,442 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
+		if( NET_IsSocketError( getsockopt( broker.socket, SOL_SOCKET, SO_ERROR, (char *)&err, &err_len )))
+		{
+			Con_Printf( S_ERROR "%s: getsockopt() failed\n", __func__ );
+			return;
+		}
+		else if( err != 0 )
+		{
+			Con_Printf( S_ERROR "%s: connection failed with error %d\n", __func__, err );
```
a1batross, line 427: String error.

a1batross (changes requested):

a1batross:
Bug? Not sure if that's intended behavior.

Immediately after setting correct address in `cl_steam_broker_addr`, Steam shows that I'm in Half-Life, even though I'm not on the server.

a1batross:
Tested with your C++ implementation, seems to be working fine. With a simulated crash and even SIGKILL it also notified Steam about disconnect. Good job.

engine/client/cl_steam.c
SNMetamorph: > This will make `sb_connect` lost forever.

But why? Function removes frame from RX buffer regardless of frame content. But yeah, it won't be processed in same game frame, only on next data arrival, so it's a flaw.

Documentation/extensions/steam-broker.md
SNMetamorph: This diagram does not represents FSM state, it's just description of "what happening now" on broker side.

engine/client/cl_steam.c
SNMetamorph: You are right, good catch

engine/client/cl_steam.c
SNMetamorph: What is exact error? `err` are type of `int`, so `%d` fits there.

SNMetamorph:
> Bug? Not sure if that's intended behavior.
> 
> Immediately after setting correct address in `cl_steam_broker_addr`, Steam shows that I'm in Half-Life, even though I'm not on the server.

This is totally ok, Steam API being initialized upon connecting to broker, session lifetime isn't tied to connection to game server, this is matches GoldSrc behavior.

engine/client/cl_steam.c
SNMetamorph: Ok, I added check just in case

engine/client/cl_steam.c
SNMetamorph: Fixed

a1batross:
> > Bug? Not sure if that's intended behavior.
> > Immediately after setting correct address in `cl_steam_broker_addr`, Steam shows that I'm in Half-Life, even though I'm not on the server.
> 
> This is totally ok, Steam API being initialized upon connecting to broker, session lifetime isn't tied to connection to game server, this is matches GoldSrc behavior.

Not sure, I would prefer to not announce Steam when it's not required. This is a matter of preference, maybe we could handle this in engine under a cvar later.

Documentation/extensions/steam-broker.md
a1batross: Ok.

engine/client/cl_steam.c
a1batross: For me it says 110, it matches ETIMEDOUT value on my system. Can't find any credible source that SO_ERROR matches errno, so sure let's leave it that way.

engine/client/cl_steam.c
a1batross: Then there is alternative solution: check for MSG_ReadBytes return value. It does return false on overflow.

a1batross: ^ not sure about that, but it does return something for overflow :)

engine/client/cl_steam.c
a1batross: Wrong variable, not resolved.

engine/client/cl_steam.c
```diff
@@ -52,140 +91,441 @@ static qboolean SteamBroker_UpdateBrokerAddress( void )
...
 	char buf[512];
 	int len = Q_snprintf( buf, sizeof( buf ), "sb_gamedir %s", gamedir );
 
-	NET_SendPacket( NS_CLIENT, len, buf, broker.adr );
-}
-
-void SteamBroker_AnnounceGameShutdown( netadr_t broker_addr )
-{
-	NET_SendPacket( NS_CLIENT, sizeof( "sb_terminate" ) - 1, "sb_terminate", broker_addr );
+	if( len > 0 ) {
```
a1batross: Either remove braces or put opening brace on new line.

SNMetamorph: Fixed

SNMetamorph:
So I think I finished this PR

engine/client/cl_steam.c
SNMetamorph: Ok I'll check tomorrow

SNMetamorph: Fixed

engine/client/cl_steam.c
SNMetamorph: Fixed now?

a1batross (approved):

PR #2574, opened 2026-05-01
Fix hlsdk-xash3d references where applicable.

malortie:
This changes references of hlsdk-xash3d (previous name) to hlsdk-portable (current name) where applicable.

The other occurrences were left as is:

https://github.com/FWGS/xash3d-fwgs/blob/7d87d65a3fc5c4b5b6241c352d82102032d8a2e7/Documentation/opensource-mods.md?plain=1#L307

https://github.com/FWGS/xash3d-fwgs/blob/7d87d65a3fc5c4b5b6241c352d82102032d8a2e7/scripts/gha/deps_nswitch.sh#L16

https://github.com/FWGS/xash3d-fwgs/blob/7d87d65a3fc5c4b5b6241c352d82102032d8a2e7/scripts/gha/deps_psvita.sh#L50

### DIFF
--- a/Documentation/development/engine-porting-guide.md
+++ b/Documentation/development/engine-porting-guide.md
@@ -22,7 +22,7 @@ The other yet unsupported configuration is the big endian.
 
 2) Open `public/build.h` file. Add appropriate checks for your operating system and/or CPU architecture. It shouldn't be hard.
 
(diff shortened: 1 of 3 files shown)


a1batross (approved):
Good job.

Yeah, `hlsdk-xash3d` is the older name of the project. While GitHub should handle URL redirection, it's better to change it to the current name.

PR #2580, opened 2026-05-05
 engine: move {gamedir}/downloaded directory to {gamedir}_downloads for compatibility

a1batross:
Required to connect to Sven-Coop 4.8 servers as it's searches for soundcache files downloaded from the server and expects placement of download directory like in GoldSrc.

I don't see any reason to make it an option, as downloads directory are ephemeral, though I think it might be fair to rename the folder on first run to make migration somewhat seamless but that's not implemented right now.

Not tested at all at the moment.


a1batross:
If there is no objections, I'm merging it.

PR #2581, opened 2026-05-06
Mute non-local sounds and background music in multiplayer when menu is open

Vladislav4KZ:
Fixes a recent regression where sounds are paused instead of muted when opening a menu in multiplayer, preventing a sudden burst of accumulated sounds.

### DIFF
--- a/engine/client/sound/s_mix.c
+++ b/engine/client/sound/s_mix.c
@@ -261,6 +261,11 @@ static int S_MixNormalChannels( portable_samplepair_t *dst, int end, int rate )
 			{
 				// play, playvol
 			}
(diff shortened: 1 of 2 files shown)


a1batross:
Maybe we can revert the check instead?

I did change IsSinglePlayerGame to IsLocalClient because I thought it's supposed to be that.

Vladislav4KZ:
Previously, the menu didn't mute or pause the sounds in multiplayer at all.

Vladislav4KZ:
I think I reported a bug a long time ago where the menu stopped muting sounds in multiplayer. And now, in the latest commits, background in-game sounds no longer play, but instead of being muted, they are being paused.

Issue #10, opened 2018-11-09
iOS port


a1batross:

a1batross:
https://github.com/FWGS/xash3d-fwgs/issues/61

PR #2582, opened 2026-05-07
Implement downloading mod libraries on Android from hlsdk-mega-build

a1batross:


Vladislav4KZ:
It might be better to copy the mod libraries compiled for Android (from **cl_dlls** and **dlls**) into the engine's directory (e.g. /data/data/su.xash.engine.test/files/ or /data/user/0/su.xash.engine.test/files/) and load them from there.

Vladislav4KZ:
Also need to account for mod library updates: if the user installs a newer version of the mod (with the same gamedir), the libraries stored in the engine's directory should be updated as well, not reused from the previous version.

yalnie:
wOW

yalnie:
@a1batross, btw, i'm thinking of adding a documents provider so we can view the libraries downloaded to xash3d without root

yalnie:
we can view /data/data/su.xash.engine/files folder using the documents provider

yalnie:
<details>
Screenshot_20260509_151429_Files
Screenshot_20260509_151543_Files
</details>

a1batross:
I don't think it really makes sense. Nothing except barebones AOSP file manager supports documents providers.

If you think you're a power user, just stay away from Android.

It's also out of scope for this PR.

nekonomicon:
Need to have separate .json-file with lists of mod directories and/or mod titles(from liblist.gam).
Without title's check it looks useless.
Because we already have: 
 - mods with one mod directory(Poke646: Vendetta and Bloody Pizza: Vendetta);
 - mods without constant directory name(Residual Point and Residual Life);
 - mappacks for mods(Opposing Force, Spirit of Half-Life).

Plus we need to start somehow some mods from valve directory(Botman's bot #10, bubblemod, miniag).
Also, need option to download mod extras(cstrike, tfc).
Plus need option to download external bots for different mods.

PR #2584, opened 2026-05-09
engine: net_ws: round-robin IPv4/IPv6 reads

vs49688:
Previously, it would always return on the first one that had activity.
This meant that if IPv4 starts getting slammed, IPv6 reads wouldn't happen.

First commit implements the logic, second applies the indentation changes for easy diff'ing.

I've been running this on my server with both v4/v6 for a week, and have thrown all sorts of 
`socat -u /dev/urandom UDP6:` and `socat  -u /dev/urandom UDP:` variants at it with no issue.


a1batross:
Nice catch!

PR #2596, opened 2026-05-17
Implementation of remote console protocol

### LINKED ISSUE #2406: Remote console access protocol (as alternative for legacy RCON protocol)
We definitely need to have two-way remote connection with dedicated servers. My suggestion is to use VConsole protocol from Source 2, this is [simple binary protocol, based on TCP](https://github.com/theokyr/CS2RemoteConsole/tree/master/libvconsole). Though it does not have any encryption, but I think this should not be a problem as usually you don't need to pass remote connection directly over the Internet, usually VPN or SSH forwarding are involved in such contexts. Also, there is already some clients for this protocol, which will be helpful during testing.

Why we shouldn't use RCON anymore:
- Works on same port as game protocol, therefore it's pretty problematic to filter RCON packets on firewall
- Based on raw UDP, not even Netchan. Vulnerable to packet loss and data corruption
- Does not have three-way handshake check (aka "challenge") which makes it vulnerable to source IP spoofing.

SNMetamorph:
Resolves #2406

### DIFF
--- a/engine/common/system.c
+++ b/engine/common/system.c
@@ -517,6 +517,7 @@ void Sys_Print( const char *pMsg )
 	Sys_PrintLog( pMsg );
 
 	Rcon_Print( &host.rd, pMsg );
+	XRcon_Print( pMsg );
 }
 
 /*
--- a/engine/common/xrcon.c
+++ b/engine/common/xrcon.c
@@ -1,0 +1,566 @@
+/*
+xrcon.c - implementation of XRCON remote console access server
+Copyright (C) 2026 Xash3D FWGS contributors
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+*/
+
+#include "common.h"
+#include "net_ws.h"
+#include "net_ws_private.h"
+#include "net_buffer.h"
+
+#define XRCON_HEADER_SIZE		12
+#define XRCON_MAX_FRAME_SIZE	4096
+#define XRCON_PRINT_BUFFER_SIZE	4096
+#define XRCON_TX_BUFFER_SIZE	(XRCON_MAX_FRAME_SIZE * 4)
+#define XRCON_CMND_VERSION		0x00D40000
+#define XRCON_CHAN_RECORD_SIZE	58
+#define XRCON_MAX_PACKET_SIZE	(XRCON_HEADER_SIZE + 4 + 24 + XRCON_PRINT_BUFFER_SIZE)
+#define XRCON_FLUSH_INTERVAL	0.05
+#define XRCON_RETRY_DELAY		5.0
+
+static CVAR_DEFINE_AUTO( xrcon_enable, "0", FCVAR_PRIVILEGED, "enable XRCON server" );
+static CVAR_DEFINE_AUTO( xrcon_address, "127.0.0.1:27000", FCVAR_ARCHIVE | FCVAR_PRIVILEGED, "XRCON server bind address and port" );
+
+typedef enum
+{
+	XRCON_STATE_IDLE,
+	XRCON_STATE_LISTENING,
+	XRCON_STATE_CONNECTED,
+	XRCON_STATE_DISCONNECTING
+} xrcon_state_t;
+
+typedef struct
+{
+	netadr_t bindadr;
+	int listen_socket;
+	int client_socket;
+	xrcon_state_t state;
+
+	uint32_t tx_pos;
+	uint32_t rx_pos;
+	uint8_t tx_buffer[XRCON_TX_BUFFER_SIZE];
+	uint8_t rx_buffer[XRCON_MAX_FRAME_SIZE + 64];
+	
+	char print_buffer[XRCON_PRINT_BUFFER_SIZE];
+	uint32_t print_pos;
+	double print_flush_time;
+	double retry_timeout;
+} xrcon_t;
+
+static xrcon_t xrcon;
+
+static void XRcon_SetState( xrcon_state_t new_state )
+{
+	xrcon.state = new_state;
+}
+
+static void XRcon_CloseListenSocket( void )
+{
+	if( NET_IsSocketValid( xrcon.listen_socket ))
+	{
+		closesocket( xrcon.listen_socket );
+		xrcon.listen_socket = INVALID_SOCKET;
+	}
+}
+
+static void XRcon_CloseClientSocket( void )
+{
+	if( NET_IsSocketValid( xrcon.client_socket ))
+	{
+		closesocket( xrcon.client_socket );
+		xrcon.client_socket = INVALID_SOCKET;
+	}
+}
+
+static void XRcon_DisconnectClient( void )
+{
+    if( xrcon.state == XRCON_STATE_CONNECTED )
+        Con_Printf( S_NOTE "%s: client disconnected\n", __func__ );
+
+	XRcon_CloseClientSocket();
+	xrcon.print_pos = 0;
+	xrcon.tx_pos = 0;
+	xrcon.rx_pos = 0;
+	xrcon.print_buffer[0] = '\0';
+}
+
+static qboolean XRcon_SendPacket( const char *type, const void *body, size_t body_len )
+{
+	uint16_t total_len = XRCON_HEADER_SIZE + body_len;
+	uint8_t frame[XRCON_MAX_PACKET_SIZE];
+	sizebuf_t sb;
+
+	if( total_len > sizeof( frame ))
+	{
+		Con_Printf( S_WARN "%s: packet too large (%u > %zu)\n", __func__, total_len, sizeof( frame ));
+		return false;
+	}
+
+	MSG_Init( &sb, "XRcon_SendPacket", frame, sizeof( frame ));
+	MSG_WriteBytes( &sb, type, 4 );
+	MSG_WriteDword( &sb, htonl( XRCON_CMND_VERSION ));
+	MSG_WriteWord( &sb, htons( total_len ));
+	MSG_WriteWord( &sb, 0 );
+
+	if( body && body_len > 0 )
+		MSG_WriteBytes( &sb, body, body_len );
+
+	size_t frame_size = MSG_GetRealBytesWritten( &sb );
+	int sent = send( xrcon.client_socket, (const char *)frame, frame_size, 0 );
+	if( NET_IsSocketError( sent ))
+	{
+		int err = WSAGetLastError();
+		if( err != WSAEWOULDBLOCK && err != WSAEALREADY )
+		{
+			Con_Printf( S_ERROR "%s: send error %s\n", __func__, NET_ErrorString( ));
+			XRcon_DisconnectClient();
+			return false;
+		}
+		sent = 0;
+	}
+
+	size_t unsent = frame_size - sent;
+	if( unsent > 0 )
+	{
+		size_t available = sizeof( xrcon.tx_buffer ) - xrcon.tx_pos;
+		if( available < unsent )
+		{
+			Con_Printf( S_ERROR "%s: transmit buffer overflow\n", __func__ );
+			XRcon_DisconnectClient();
+			return false;
+		}
+
+		memcpy( xrcon.tx_buffer + xrcon.tx_pos, frame + sent, unsent );
+		xrcon.tx_pos += unsent;
+	}
+
+	return true;
+}
+
+static void XRcon_SendChan( void )
+{
+	uint8_t body[2 + XRCON_CHAN_RECORD_SIZE];
+	sizebuf_t sb;
+	char chan_name[34];
+
+	MSG_Init( &sb, "XRcon_SendChan", body, sizeof( body ));
+
+	MSG_WriteWord( &sb, htons( 1 ));
+
+	MSG_WriteDword( &sb, 0 ); // id
+	MSG_WriteDword( &sb, 0 ); // unknown1
+	MSG_WriteDword( &sb, 0 ); // unknown2
+	MSG_WriteDword( &sb, htonl( 5 )); // verbosity_default
+	MSG_WriteDword( &sb, htonl( 5 )); // verbosity_current
+
+	// RGBA = white
+	MSG_WriteByte( &sb, 255 );
+	MSG_WriteByte( &sb, 255 );
+	MSG_WriteByte( &sb, 255 );
+	MSG_WriteByte( &sb, 255 );
+
+	Q_strncpy( chan_name, "Console", sizeof( chan_name ));
+	MSG_WriteBytes( &sb, chan_name, sizeof( chan_name ));
+
+	XRcon_SendPacket( "CHAN", body, MSG_GetRealBytesWritten( &sb ));
+}
+
+static qboolean XRcon_UpdateBindAddress( void )
+{
+	if( NET_NetadrType( &xrcon.bindadr ) == NA_UNDEFINED )
+	{
+		if( !NET_StringToAdr( xrcon_address.string, &xrcon.bindadr ))
+			return false;
+	}
+	return true;
+}
+
+static void XRcon_StartListening( void )
+{
+	int addr_family;
+	struct sockaddr_storage addr = { 0 };
+
+	if( !XRcon_UpdateBindAddress( ))
+	{
+		Con_Printf( S_ERROR "%s: invalid address \"%s\"\n", __func__, xrcon_address.string );
+		return;
+	}
+
+	if( NET_NetadrType( &xrcon.bindadr ) == NA_IP )
+		addr_family = AF_INET;
+	else if( NET_NetadrType( &xrcon.bindadr ) == NA_IP6 )
+		addr_family = AF_INET6;
+	else
+	{
+		Con_Printf( S_ERROR "%s: unsupported address type for %s\n", __func__, xrcon_address.string );
+		return;
+	}
+
+	xrcon.listen_socket = socket( addr_family, SOCK_STREAM, IPPROTO_TCP );
+	if( !NET_IsSocketValid( xrcon.listen_socket ))
+	{
+		Con_Printf( S_ERROR "%s: failed to create listen socket\n", __func__ );
+		return;
+	}
+
+	if( !NET_MakeSocketNonBlocking( xrcon.listen_socket ))
+	{
+		Con_Printf( S_ERROR "%s: failed to set non-blocking mode, error %s\n", __func__, NET_ErrorString( ));
+		XRcon_CloseListenSocket();
+		return;
+	}
+
+	NET_MakeSocketReuseAddr( xrcon.listen_socket );
+	NET_NetadrToSockadr( &xrcon.bindadr, &addr );
+
+	if( bind( xrcon.listen_socket, (struct sockaddr *)&addr, NET_SockAddrLen( &addr )) == SOCKET_ERROR )
+	{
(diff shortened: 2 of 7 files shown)


SNMetamorph:
Client https://github.com/theokyr/CS2RemoteConsole is compatible with it.

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+void XRcon_Print( const char *msg )
+{
+	if( xrcon.state != XRCON_STATE_CONNECTED )
+		return;
+
+	if( !msg || !*msg )
+		return;
+
+	size_t len = Q_strlen( msg );
+	for( size_t i = 0; i < len; i++ )
```
a1batross: Can you rewrite it using `strchr` and `memcpy`?

There is code in system.c that does exactly that.

a1batross:
Sorry, can you remind me, why exactly we need the new protocol?

And also tell how much of VConsole2 protocol is implemented.

engine/client/cl_steam.c
```diff
@@ -45,7 +45,7 @@ GNU General Public License for more details.
 #define SBRK_CONNECT_RETRY_DELAY	5.0
 #define SBRK_TICKET_SIZE_MAX 		2048
 
-static CVAR_DEFINE_AUTO( cl_steam_broker_addr, "127.0.0.1:27420", FCVAR_ARCHIVE, "address of steam broker instance" );
+static CVAR_DEFINE_AUTO( cl_steam_broker_addr, "127.0.0.1:27420", FCVAR_PRIVILEGED | FCVAR_ARCHIVE, "address of steam broker instance" );
```
a1batross: Good catch. Can you send it separately?

engine/common/system.c
```diff
@@ -521,6 +521,7 @@ void Sys_Print( const char *pMsg )
 	Sys_PrintLog( pMsg );
 
 	Rcon_Print( &host.rd, pMsg );
+	XRcon_Print( pMsg );
```
a1batross, line 520: Can it use the same redirect facility from usual rcon, or it's not applicable for some reason?

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+		memcpy( xrcon.tx_buffer + xrcon.tx_pos, frame + sent, unsent );
+		xrcon.tx_pos += unsent;
+	}
+
+	return true;
+}
+
+static void XRcon_SendAinf( void )
+{
+	uint8_t body[76];
```
a1batross: `uint8_t body[76] = { 0 };`

What is 76? Is it the protocol requirement?

a1batross: What is AINF message?

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+	MSG_WriteDword( &sb, htonl( 5 )); // verbosity_current
+
+	// RGBA = white
+	MSG_WriteByte( &sb, 255 );
+	MSG_WriteByte( &sb, 255 );
+	MSG_WriteByte( &sb, 255 );
+	MSG_WriteByte( &sb, 255 );
+
+	Q_strncpy( chan_name, "Console", sizeof( chan_name ));
+	MSG_WriteBytes( &sb, chan_name, sizeof( chan_name ));
```
a1batross: I don't think you need chan_name variable here.

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+
+	sizebuf_t sb;
+	char type[5];
+	uint16_t total_len, handle;
+	uint32_t version;
+
+	MSG_Init( &sb, "XRcon_ProcessFrame", xrcon.rx_buffer, xrcon.rx_pos );
+	MSG_ReadBytes( &sb, type, 4 );
+	type[4] = '\0';
+	version = ntohl( MSG_ReadDword( &sb ));
```
a1batross: Declare variables where they used.

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+	if( Q_strcmp( type, "CMND" ) == 0 && body_len > 0 )
+	{
+		char cmd[XRCON_MAX_FRAME_SIZE];
+		size_t cmd_len = ( body_len < sizeof( cmd ) - 1 ) ? body_len : sizeof( cmd ) - 1;
+
+		MSG_ReadBytes( &sb, cmd, cmd_len );
+		cmd[cmd_len] = '\0';
+
+		// strip embedded NUL bytes from command string
+		{
```
a1batross, line 329: Get rid of this LLM-generated code block. I know they like to do this because that's what they trained to do, but we can... just not add them.

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+{
+	fd_set readfds;
+	struct timeval tv;
+	struct sockaddr_storage client_addr;
+	socklen_t addr_len = sizeof( client_addr );
+
+	FD_ZERO( &readfds );
+	FD_SET( xrcon.listen_socket, &readfds );
+
+	tv.tv_sec = 0;
```
a1batross: `struct timeval tv = { 0 }`.

engine/common/xrcon.c
```diff
@@ -0,0 +1,590 @@
...
+
+	xrcon.print_pos = 0;
+	xrcon.print_buffer[0] = '\0';
+}
+
+static void XRcon_UpdateListening( void )
+{
+	fd_set readfds;
+	struct timeval tv;
+	struct sockaddr_storage client_addr;
```
a1batross: `client_addr` scope could be reduced.

a1batross (changes requested):

SNMetamorph:
> Sorry, can you remind me, why exactly we need the new protocol?

Why we shouldn't use standard RCON anymore:
- Works on same port as game protocol, therefore it's pretty problematic to segregate from general traffic and filter RCON packets on firewall (to permit access for RCON for specific IPs, as an example)
- Based on raw UDP (not even Netchan). Vulnerable to packet loss, does not handles packets fragmentation and congestion control properly (these things are built-in for TCP)
- Does not have three-way handshake check (aka "challenge") which makes it vulnerable to source IP spoofing.
- New protocol could work both on server and client. And possibility to run it on client are useful for testing/development purposes
> And also tell how much of VConsole2 protocol is implemented.

Only very limited subset, like channel registration and printing command.

engine/common/xrcon.c
SNMetamorph: Won't this be even more complicated than current version?

engine/common/system.c
SNMetamorph: Seems that it isn't needed for this protocol implementation, because default RCON sends response data only upon sending command to server, but XRCON sends full console output to client.

a1batross:
>Only very limited subset, like channel registration and printing command.

Can you document it? Actually, yes, it would be nice to have an md file in Documentation/extensions about XRCon, what it is, what it's based on (the VConsole2 protocol) and what issues of usual Quake rcon it addresses.

I still don't like the fact that it allows connections from everything, especially when it's set to listen on non-loopback address. Can we do something about that?

Other than that, I'm fine with the general idea.

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+	if( NET_IsSocketValid( xrcon.client_socket ))
+	{
+		closesocket( xrcon.client_socket );
+		xrcon.client_socket = INVALID_SOCKET;
+	}
+}
+
+static void XRcon_DisconnectClient( void )
+{
+    if( xrcon.state == XRCON_STATE_CONNECTED )
```
a1batross: Spaces -> tabs.

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+
+	uint8_t body[4 + 24 + XRCON_PRINT_BUFFER_SIZE];
+	sizebuf_t sb;
+
+	MSG_Init( &sb, "XRcon_FlushPrintBuffer", body, sizeof( body ));
+	MSG_WriteDword( &sb, 0 ); // channel_id = 0 (Console)
+
+	// padding 24 bytes
+	for( int pad = 0; pad < 24; pad++ )
+		MSG_WriteByte( &sb, 0 );
```
a1batross: MSG_WriteDword would be more efficient, writing 4 bytes at once.

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+	XRcon_CloseClientSocket();
+	xrcon.print_pos = 0;
+	xrcon.tx_pos = 0;
+	xrcon.rx_pos = 0;
+	xrcon.print_buffer[0] = '\0';
+}
+
+static qboolean XRcon_SendPacket( const char *type, const void *body, size_t body_len )
+{
+	uint16_t total_len = XRCON_HEADER_SIZE + body_len;
```
a1batross: Use size_t.

body_len would easily overflow 16-byte total_len.

a1batross: Or check that body_len is at reasonable size.

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+		return false;
+
+	sizebuf_t sb;
+	char type[5] = { 0 };
+
+	MSG_Init( &sb, "XRcon_ProcessFrame", xrcon.rx_buffer, xrcon.rx_pos );
+	MSG_ReadBytes( &sb, type, 4 );
+	uint32_t version = ntohl( MSG_ReadDword( &sb ));
+	uint16_t total_len = ntohs( MSG_ReadWord( &sb ));
+	uint16_t handle = ntohs( MSG_ReadWord( &sb )); // how to make sure that compiler does not prune this?
```
a1batross: >how to make sure that compiler does not prune this?

why it would?

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+
+	if( version != XRCON_CMND_VERSION )
+		Con_Printf( S_WARN "%s: unexpected version 0x%08X\n", __func__, version );
+
+	if( total_len < XRCON_HEADER_SIZE || xrcon.rx_pos < total_len )
+		return false;
```
a1batross: Early return wouldn't consume a frame, making it called every frame and early return again.

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+static void XRcon_UpdateIdle( void )
+{
+	if( xrcon.retry_timeout > Platform_DoubleTime( ))
+		return;
+
+	XRcon_StartListening();
+	xrcon.retry_timeout = Platform_DoubleTime() + XRCON_RETRY_DELAY;
+}
+
+void XRcon_Print( const char *msg )
```
a1batross, line 524: Make sure that anything called inside XRcon_Print doesn't call XRcon_Print itself.

What I mean is that XRcon_SendPacket called from XRcon_Print might call Con_Printf, which in turn will call XRcon_Print, causing a recursion.

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+		}
+
+		if( !NET_MakeSocketNonBlocking( client ))
+		{
+			Con_Printf( S_ERROR "%s: failed to set client non-blocking\n", __func__ );
+			closesocket( client );
+			return;
+		}
+
+		xrcon.client_socket = (int)client;
```
a1batross: Can we avoid this cast and just use SOCKET type?

engine/common/xrcon.c
```diff
@@ -0,0 +1,555 @@
...
+	{
+		if( msg[i] == '\n' || msg[i] == '\r' )
+		{
+			if( xrcon.print_pos > 0 )
+			{
+				XRcon_FlushPrintBuffer();
+				xrcon.print_flush_time = Platform_DoubleTime() + XRCON_FLUSH_INTERVAL;
+			}
+		}
+		else if( msg[i] == '^' && i + 1 < len )
```
a1batross: BTW there are macros for them.

engine/common/xrcon.c
a1batross: Doesn't look to me very complicated. We already do this.

a1batross (changes requested):

engine/client/cl_steam.c
a1batross: Still, send it separately :)

engine/common/xrcon.c
SNMetamorph: Code in `system.c` does exactly per-byte traversing same as current code. Since we need to check several kinds of tokens, we need to call `strchr` several times for each step, which is even more code than simple loop. Or maybe I misunderstood you?

a1batross: My bad, it's `Sys_PrintLogfile`. Though instead of Q_strchr it might make sense to use Q_strpbrk as it matching multiple characters.

SNMetamorph: Q_strpbrk does not helps here, because we need to match not a string, but several different tokens (not necessarily consequent)

a1batross: It will work but `msg` has to be null-terminated.

engine/common/xrcon.c
SNMetamorph: Ok

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+		{
+			size_t payload_length = xrcon.parser.frame.length - XRCON_HEADER_SIZE;
+			if( xrcon.rx_pos < payload_length )
+				return;
+
+			if( Q_strcmp( xrcon.parser.frame.type, "CMND" ) == 0 )
+			{
+				char cmd[XRCON_MAX_FRAME_SIZE];
+				size_t cmd_len = Q_min( payload_length, sizeof( cmd ) - 1 );
+				memcpy( cmd, xrcon.rx_buffer, payload_length );
```
a1batross, line 338:
```suggestion
				memcpy( cmd, xrcon.rx_buffer, payload_length );
				cmd[cmd_len] = 0;
```

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+			{
+				memcpy( xrcon.print_buffer + xrcon.print_pos, msg, length );
+				xrcon.print_pos += length;
+			}
+			break;
+		}
+		else if( *p == '\n' || *p == '\r' )
+		{
+			size_t length = p - msg;
+			memcpy( xrcon.print_buffer + xrcon.print_pos, msg, length );
```
a1batross: Overflow check missing here. You did one in `p == NULL` branch, just copy it here.

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+				xrcon.print_flush_time = Platform_DoubleTime() + XRCON_FLUSH_INTERVAL;
+			}
+			msg = p + 1;
+		}
+		else if( IsColorString( p ))
+		{
+			if( p != msg )
+			{
+				size_t length = p - msg;
+				memcpy( xrcon.print_buffer + xrcon.print_pos, msg, length );
```
a1batross: Same here with the overflow check.

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+				xrcon.print_pos += length;
+				
+			}
+			msg = p + 2;
+		}
+		else
+		{
+			size_t length = p - msg + 1;
+			memcpy( xrcon.print_buffer + xrcon.print_pos, msg, length );
```
a1batross: Same here with the overflow check.

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+#define XRCON_CMND_VERSION		0x000000D4
+#define XRCON_CHAN_RECORD_SIZE	58
+#define XRCON_CHAN_NAME_SIZE	34
+#define XRCON_AINF_PACKET_SIZE	77
+#define XRCON_MAX_PACKET_SIZE	(XRCON_HEADER_SIZE + 4 + 24 + XRCON_PRINT_BUFFER_SIZE)
+#define XRCON_FLUSH_INTERVAL	0.05
+#define XRCON_RETRY_DELAY		5.0
+
+static CVAR_DEFINE_AUTO( xrcon_enable, "0", FCVAR_PRIVILEGED, "enable remote console access server" );
+static CVAR_DEFINE_AUTO( xrcon_address, "127.0.0.1:27000", FCVAR_ARCHIVE | FCVAR_PRIVILEGED, "XRCON server bind address and port" );
```
a1batross: Let's remove FCVAR_ARCHIVE here. It's easy to set it to 0.0.0.0 and then forget, and then expose the console to everybody on the whole Internet.

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+
+static void XRcon_SendADON( const char *name )
+{
+	sizebuf_t sb;
+	uint8_t data[512];
+	
+	MSG_Init( &sb, __func__, data, sizeof( data ));
+	MSG_WriteWord( &sb, htons( 0 ));
+	MSG_WriteWord( &sb, htons( strlen( name )));
+	MSG_WriteString( &sb, name );
```
a1batross: `MSG_WriteString` null terminates. Is this what supposed to be used here, considering the header already passes the string length?

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+{
+	xrcon.print_pos = 0;
+	xrcon.tx_pos = 0;
+	xrcon.rx_pos = 0;
+	xrcon.print_buffer[0] = '\0';
+	xrcon.parser.state = XRCON_STATE_WAIT_HEADER;
+
+	XRcon_CloseClientSocket();
+	XRcon_SetState( XRCON_STATE_LISTENING );
+    Con_Printf( S_NOTE "%s: client disconnected\n", __func__ );
```
a1batross:
```suggestion
	Con_Printf( S_NOTE "%s: client disconnected\n", __func__ );
```

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+	return true;
+}
+
+static void XRcon_SendCHAN( void )
+{
+	sizebuf_t sb;
+	uint8_t body[2 + XRCON_CHAN_RECORD_SIZE];
+	char channel_name[XRCON_CHAN_NAME_SIZE] = { 0 };
+
+	strncpy( channel_name, "Console", sizeof( channel_name ));
```
a1batross: Q_strncpy

engine/common/xrcon.c
```diff
@@ -0,0 +1,642 @@
...
+}
+
+static void XRcon_SendADON( const char *name )
+{
+	sizebuf_t sb;
+	uint8_t data[512];
+	
+	MSG_Init( &sb, __func__, data, sizeof( data ));
+	MSG_WriteWord( &sb, htons( 0 ));
+	MSG_WriteWord( &sb, htons( strlen( name )));
```
a1batross: Q_strlen

a1batross (changes requested):

engine/common/xrcon.c
SNMetamorph: Yes, I agree.

SNMetamorph:
Not sure about documentation though. The only thing we can document properly is framing format. Any messages types description will be nothing more but plenty of unknown/padding fields (but there are potential for us to explore things more).

a1batross:
>Not sure about documentation though. The only thing we can document properly is framing format. Any messages types description will be nothing more but plenty of unknown/padding fields (but there are potential for us to explore things more).

Even if it's unknown, unused, reserved, it's fine. It's a study on reverse engineering original protocol and not all things could be known at this moment, only those that are handled and implemented.

SNMetamorph:
> > Not sure about documentation though. The only thing we can document properly is framing format. Any messages types description will be nothing more but plenty of unknown/padding fields (but there are potential for us to explore things more).
> 
> Even if it's unknown, unused, reserved, it's fine. It's a study on reverse engineering original protocol and not all things could be known at this moment, only those that are handled and implemented.

Ok, I submitted documentation, it seems somehow good enough for now. So let's finalize this PR :)

SNMetamorph:
Closing since it is merged manually.

PR #2598, opened 2026-05-18
engine: update player logo color to RGB format

Vladislav4KZ:
See also: https://github.com/FWGS/mainui_cpp/pull/159


a1batross:
I would honestly keep logo creation on UI level, engine doesn't really care what kind of logo it's going to send. On UI level we can not only implement any kind of post-processing, but let player draw their own logo with a simple MS Paint-like drawing widget.

a1batross:
Therefore I don't really think engine needs the `cl_logocolor` cvar itself but a way to notify the engine that we need to send it to the server.

Vladislav4KZ:
Reverted all changes and converted the ``cl_logocolor`` cvar value to RGB format.

a1batross (approved):
Will merge after https://github.com/FWGS/mainui_cpp/pull/159.

Vladislav4KZ:
@a1batross check the new commits.

a1batross (approved):

PR #2601, opened 2026-05-20
Unaligned access handler for MIPS-IRIX

Xav101:
More of the unaligned access handlers needed on MIPS `mstudioanimvalue_t` data is commonly allocated at unaligned positions.

Because this is in C code we can't use the C++ template-based method we used in hlsdk. (You could do something similar with a number of various macros and compiler extensions, but I can't think of one that would be well supported across all compilers on all platforms. So it's probably best to just leave it at having to manually specify the type you're accessing for now.)

### DIFF
--- a/common/xash3d_types.h
+++ b/common/xash3d_types.h
@@ -273,4 +273,11 @@ typedef int qboolean;
 	#define HostFourCC( a, b, c, d ) LittleFourCC( a, b, c, d )
 #endif
 
+static inline short UnalignedShort( short *x )
+{
+	short y;
+	memcpy( &y, x, sizeof( y ) );
+	return y;
+}
+
 #endif // XASH_TYPES_H
--- a/public/xash3d_mathlib.c
+++ b/public/xash3d_mathlib.c
@@ -470,23 +470,23 @@ void R_StudioCalcBones( int frame, float s, const mstudiobone_t *pbone, const ms
 
 		if( panimvalue->num.valid > j )
 		{
-			v1[i] = panimvalue[j + 1].value;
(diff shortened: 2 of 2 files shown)


common/xash3d_types.h
```diff
@@ -273,4 +273,13 @@ typedef int qboolean;
 	#define HostFourCC( a, b, c, d ) LittleFourCC( a, b, c, d )
 #endif
 
+static inline short _unaligned_short( short *x )
```
a1batross: Identifiers starting with `_` are reserved. 

Do not wrap it into a macro, just use it as normal function.

common/xash3d_types.h
```diff
@@ -273,4 +273,13 @@ typedef int qboolean;
...
+{
+	short y;
+	memcpy( &y, x, sizeof(short) );
```
a1batross:
```suggestion
	memcpy( &y, x, sizeof( y ));
```

a1batross (changes requested):

common/xash3d_types.h
Xav101: The reason I wanted it as a macro was to avoid having to write `unaligned_short( &(x) )` everywhere in the code where we have to use it. If that's fine then we can just convert it to a normal function.

a1batross: Yeah, I get it. Not sure if we can do anything about it in C, so let's not introduce new macros.

Also, use PascalCase so that UnalignedShort reads the same as LittleShort, for example.

Xav101:
Requested changes completed

a1batross (approved):

PR #2603, opened 2026-05-20
fs_find console command

### LINKED ISSUE #2124: Console command "fs_lookupfile" for searching files in search pathes
We definitely need some method to investigate where exactly located file which we're trying to find, because we easily can have hundreds of different PAK/PK3 archives contained in game directory. 

Let's pretend, we know that file `models/tree01.mdl` located somewhere in game directory and we have willing to find it and modify in some way. But it could be located anywhere across a lot of different search pathes, and if we will search for it manually, it will be very non-optimal & time consuming.

But if we could have such console command, we will just pass `fs_lookupfile models/tree01.mdl` to console, and could see exact PAK/PK3 archive where desired file located. Or even better, see all search pathes where this file exists, in descending order by their priority (top-level search path overrides everything else, i.e. just like VFS works by itself).

SNMetamorph:
Resolves #2124
image

### DIFF
--- a/filesystem/filesystem.h
+++ b/filesystem/filesystem.h
@@ -241,6 +241,7 @@ typedef struct fs_api_t
 	qboolean (*GetRootDirectory)( char *path, size_t size );
 
 	void (*MakeGameInfo)( void );
+	void (*FindFile_f)( const char *filename );
 } fs_api_t;
 
 typedef struct fs_interface_t
--- a/filesystem/searchpath.c
+++ b/filesystem/searchpath.c
@@ -787,6 +787,35 @@ void FS_Path_f( void )
 
 /*
 ====================
+FS_FindFile_f
+
+Print all search paths where the file was found,
+ordered by priority (first one is used for FS operations)
+====================
+*/
+void FS_FindFile_f( const char *filename )
+{
+	int	count = 0;
+	Con_Printf( "File " S_YELLOW "%s" S_DEFAULT " occurences:\n", filename );
+
+	for( searchpath_t *s = fs_searchpaths; s; s = s->next )
+	{
+		string fixedname;
+		if( s->pfnFindFile( s, filename, fixedname, sizeof( fixedname )) >= 0 )
+		{
+			string info;
+			count++;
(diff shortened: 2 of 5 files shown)


filesystem/filesystem.h
```diff
@@ -167,6 +167,7 @@ typedef struct fs_api_t
 	int (*SetCurrentDirectory)( const char *path );
 	qboolean (*FindLibrary)( const char *dllname, qboolean directpath, fs_dllinfo_t *dllinfo );
 	void (*Path_f)( void );
+	void (*FindFile_f)( const char *filename );
```
a1batross: If you add it the middle of struct, bump version.

If you don't want to bump version, move new function at the end of the struct.

filesystem/searchpath.c
```diff
@@ -773,6 +785,35 @@ void FS_Path_f( void )
...
+Print all search paths where the file was found,
+ordered by priority (first one is used for FS operations)
+====================
+*/
+void FS_FindFile_f( const char *filename )
+{
+	int	count = 0;
+	string fixedname, info;
+
+	Con_Printf( "File ^3%s^7 occurences:\n", filename );
```
a1batross: Can you use `S_XXX` macros from `com_strings.h`?

filesystem/searchpath.c
```diff
@@ -773,6 +785,35 @@ void FS_Path_f( void )
...
+FS_FindFile_f
+
+Print all search paths where the file was found,
+ordered by priority (first one is used for FS operations)
+====================
+*/
+void FS_FindFile_f( const char *filename )
+{
+	int	count = 0;
+	string fixedname, info;
```
a1batross: Reduce visibility of these variables, `fixedname` could be moved into loop, `info` could be moved into pfnFindFile branch.

a1batross (changes requested):

filesystem/searchpath.c
SNMetamorph: Fixed

a1batross (approved):

PR #2622, opened 2026-05-24
android: implement steamkit integration for downloading game files

hasandramali:
Since downloading game files from third-party sources often causes broken behavior and unnecessary bug reports, I added a "Game Data Downloader" feature to the Xash3D launcher. It fetches files anonymously from Steam servers, similar to SteamCMD (referencing the JavaSteam project). Additionally, I fixed a bug where the app requested storage permissions three times consecutively upon first launch. Please test and review it.
Screenshot_2026-05-24-16-44-43-541_su xash engine test

https://github.com/user-attachments/assets/6196a01c-da58-400c-8be8-5d32cc6290a1

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -138,6 +138,7 @@ dependencies {
 	implementation(libs.swiperefreshlayout)
 
 	implementation(libs.acra.http)
+	implementation(libs.xz)
 }
 
 fun getBuildNum(): Int {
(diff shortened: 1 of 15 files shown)


hasandramali:
you're already working on this.

a1batross:
We do not condone piracy, therefore we won't use HLDS game data, as it's only provided for running servers.

And you're definitely not running a dedicated server on your phone.

PR #2630, opened 2026-05-30
engine: add support for legacy 'm' query responses

Vladislav4KZ:
This should resolve the issue with retrieving the server status (hostname, map, numcl/maxcl) for ``srv.tsmod.net:27015``.

I referred to the obsolete GoldSource response documentation:
https://developer.valvesoftware.com/wiki/Server_queries#Obsolete_GoldSource_Response


a1batross (commented):
Looks like WON message.

Not sure why 48 protocol servers even respond with it, but I guess it's due to usage of dproto/reunion, as current HLDS doesn't have any of these messages.

Seems ok to me, merging.

PR #2633, opened 2026-06-01
ref: gl: gl_triapi: fix TriFog

GLoOoccK:
https://github.com/Velaron/cs16-client/issues/369

When a `Fog` message was received with `density = 0.0f`, TriFog correctly called `pglDisable(GL_FOG)`, but did not reset `RI.fogSkybox`.


a1batross:
Not sure if this is a correct approach. Fog control in Xash3D is a complete mess and before doing something like that I want to be sure we're not breaking anything else.

PR #2637, opened 2026-06-05
public: wrap macro arguments in parentheses

901D3:


a1batross:
looks good to me

will merge once CI passes

PR #2649, opened 2026-06-14
add missed `libatomic.i686` dependency

oooo-ps:
resolves `Compiler can't create 32-bit code!`

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ This repository contains our fork of HLSDK and restored source code for Half-Lif
 
 ##### RedHat/Fedora
 * Only for 32-bit engine on 64-bit x86 operating system:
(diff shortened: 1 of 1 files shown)


a1batross:
Can you post `build/config.log` in error state?

oooo-ps:
I closed log file after fixing this issue. It's Fedora 44.

a1batross:
You can try deleting the package, reproduce the error and post the log.

oooo-ps:
> You can try deleting the package, reproduce the error and post the log.

sorry too lazy, maybe in future.

just recently I had similar issue, with another engine but on same Fedora version, where people wondering their engine does not use one dependency directly, not sure if it's related but for info:
https://github.com/diasurgical/DevilutionX/pull/8622

I created this PR just for myself, because time-to-time I build engine on different machines and tired to find these package names undefined in docs required to build.

nekonomicon:
We doesn't have C11 code at all, and if it's waf requirement it looks strange too.

PR #2671, opened 2026-07-14
platform: ios: Remake launch dialog using UIAlertController

ksagameng2:
This is how it looks like:

I've removed the suffix textbox too since it now does nothing


a1batross:
Nice. Is it ready for merge?

ksagameng2:
> Nice. Is it ready for merge?

Yes should be

PR #2677, opened 2026-07-21
ref: soft: Try to stop decal->pnext loops from recursing infinitely

Buggem:
Doesn't fix the root cause of the problem, which is the mysterious setting of decal->pnext to decal which doesn't happen in GL, but does act as a blockade and a temporary fix.

I tried to find the root cause, but it was quite the challenge and I came back empty handed. For now this will do the trick to stop it. 

Attached is a save that gets the SW renderer stuck in an infinite loop, at least on my system.
[quick.zip](https://github.com/user-attachments/files/30222288/quick.zip)

### DIFF
--- a/ref/soft/r_decals.c
+++ b/ref/soft/r_decals.c
@@ -470,6 +470,8 @@ static decal_t *R_DecalIntersect( decalinfo_t *decalinfo, msurface_t *surf, int 
 				}
 			}
 		}
+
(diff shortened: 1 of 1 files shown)


a1batross:
Yeah, technically we shouldn't do that, but it does work as a temporary solution. We're probably leaking memory this way too.

PR #2678, opened 2026-07-23
Fix SDL3

JetbladeDevsStuff:
Fixes SDL3 support by adding missing functions to the platform (`VID_Info_f`, `Platform_GetDisplayOrientation`, sensors) and adding resize support (change from windowed to fullscreen, change mode).

Some of the variable definitions look weird after being formatted (e.g. in `VID_SaveWindowSize`) but that's what uncrustify outputted.

Most of the missing functions are copied from the SDL2 platform and updated for SDL3.

Tested using 64-bit build on arm64 macOS `./waf configure -8 --sdl-use-pkgconfig --use-sdl3 -T debug`, including built hlsdk-portable and valve assets. Also tested using xash3d-maintui. Please let me know if something is wrong on other platforms as I didn't test on them.

### DIFF
--- a/engine/platform/sdl3/platform_sdl3.h
+++ b/engine/platform/sdl3/platform_sdl3.h
@@ -28,4 +28,12 @@ GNU General Public License for more details.
 void SDLash_InitCursors( void );
 void SDLash_FreeCursors( void );
 
+//
+// sensor_sdl3.c
+//
+void SDLash_InitSensors( void );
+void SDLash_ShutdownSensors( void );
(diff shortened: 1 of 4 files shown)


a1batross:
Thanks. I already did this just two days ago and didn't pushed because it ended up clumped with a gyro aim rework.

Though I like your changes more, especially that you ported fullscreen changes from SDL2 to SDL3 backend, so I will review them and merge.

a1batross (approved):

a1batross:
Thanks, merged!

PR #2688, opened 2026-07-28
engine: implement reading hardware serial ID on Mac, implement reading MAC addresses on *BSDs

a1batross:


engine/client/identification.c
```diff
@@ -539,6 +542,17 @@ static bloomfilter_t ID_GenerateRawId( void )
 
 	count += ID_ProcessCPUInfo( &value );
 	count += ID_ProcessFiles( &value, "/sys/block", "device/cid" );
+#endif
+#if XASH_APPLE && !XASH_IOS
```
ksagameng2: Is it a good idea to create XASH_OSX instead of this? I also think that XASH_APPLE implies that it applies to all apple platforms

a1batross: Yes, XASH_APPLE implies all Apple platforms. I will add XASH_OSX.

PR #2703, opened 2026-08-05
engine: client: deliver keys in Key_Event's handled-in-client case

### LINKED ISSUE #1923: [MOD: Cry of Fear] Can't type anything except TAB button on restaurant's computer.
This bug make nearly impossible to progress since this computer contain passcode for padlock. passcode is randomized.

Info:
Patch version: 1.55 (1.2 and 1.0 untested)
Xash3D-FWGS: https://github.com/FWGS/xash3d-fwgs/commit/e751092555975c7ec3a65e1da1144ca6a4e6a1b1
OS: Windows 10 x86_64bit
GPU: GTX 1650
Custom renderer: Enabled
Map: c_start2

MohiMad:
Resolves #1923 

Below is an summary of the bug and my findings (essentially repeating what has been brought up in the issue):
---

The bug: The very first puzzle in Cry of fear is a computer screen with two input fields: a username and a password. The bug is that in FWGS the only keys that work are the TAB (alternates fields) and the ENTER button (does the comparison). None of the other keys work. This is cruicial to the gameplay as the computer puzzle reveals a randomly generated code to a padlock needed for further progression through the game. 

The reason why TAB and ENTER they work is that the client calls win32's `GetAsyncKeyState` immediately rather than letting the engine handle it. For every other key, the viewport key input dispatcher refers to the input handler which is a one-liner: 

```
return binding && !strcmp( binding, "toggleconsole" );
```

This is only ever true for the key that's bound to the console, which is why opening the console works while the computer interface is open. In fact, running `bind x toggleconsole` and then hitting "x" does open the console AND input the letter x in the input field for the computer. Funny enough, the key input handler routes the engine's Key_Event to the "handled in client.dll" case, while STILL expecting the engine to handle it, as it does nothing otherwise. 

This can be resolved by calling `VGui_KeyEvent( key, down );` regardless in Key_Event's "handled in client.dll" segment. It does reproduce the GoldSrc behaviour for the computer puzzle in CoF. It *could* produce unwanted behaviour for newer mods built using fwgs that rely on suppressing the engine's input, although this is a special/rare case as the mods will need to be inside a VGUI focus state AND suppress the engine's input for this to take effect.

### DIFF
--- a/engine/client/input/in_keys.c
+++ b/engine/client/input/in_keys.c
@@ -722,6 +722,7 @@ void GAME_EXPORT Key_Event( int key, int down )
 				keys[key].gamedown = false;
 				keys[key].repeats = 0;
 			}
+			VGui_KeyEvent( key, down ); // resolves issue #1923
 			return; // handled in client.dll
 		}
 	}
--- a/engine/client/input/in_keys.c
+++ b/engine/client/input/in_keys.c
@@ -722,8 +722,12 @@ void GAME_EXPORT Key_Event( int key, int down )
 				keys[key].gamedown = false;
 				keys[key].repeats = 0;
 			}
-			VGui_KeyEvent( key, down ); // resolves issue #1923
-			return; // handled in client.dll
+
+			// deliver keys regardless of whether the client.dll handled them or not
+			// reproduces GoldSrc behavior, where keys are passed even if the client.dll claims to have handled them
+			// this is needed for Cry of fear's Computer interface input, for more context, see https://github.com/FWGS/xash3d-fwgs/issues/1923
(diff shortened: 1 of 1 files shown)


a1batross:
Do not write comments in the code which issue was resolved, at least not with just a number, it doesn't make sense outside of GitHub. You can add a brief comment, which mod does it affect and which problem does it resolve there. You can as well add (but do not replace proper brief comment) URL to the issue.

a1batross (approved):
I wouldn't trust "reproduces GoldSrc behavior" claim, but to make it makes sense to notify VGUI about key presses regardless, so I will merge it. Thanks for the fix!

PR #2714, opened 2026-08-16
fix cross-compile for win64 from Linux via mingw64; preserve double slash in COM_FixSlashes

danglingptr0x0:
adding a mingw compilation and linking fix since I wanted to compile for win64 from within Linux. compiled and tested on my system. there was also another issue, which I only noticed upon pulling (I initially cloned months ago), which was that COM_FixSlashes was collapsing double slashes (UNC), so fixed that as well

### DIFF
--- a/engine/client/input/input.c
+++ b/engine/client/input/input.c
@@ -144,6 +144,7 @@ void IN_MouseSavePos( void )
 		return;
 
 	Platform_GetMousePos( &in_lastvalidpos.x, &in_lastvalidpos.y );
+
 	in_mouse_savedpos = true;
 }
 
--- a/common/xash3d_types.h
+++ b/common/xash3d_types.h
@@ -115,7 +115,11 @@ typedef int qboolean;
 	#endif
 	#define NORETURN           __attribute__(( noreturn ))
 	#define NONNULL            __attribute__(( nonnull ))
-	#define FORMAT_CHECK( x )  __attribute__(( format( printf, x, x + 1 )))
+	#if defined( __MINGW32__ )
+		#define FORMAT_CHECK( x )  __attribute__(( format( gnu_printf, x, x + 1 )))
+	#else
+		#define FORMAT_CHECK( x )  __attribute__(( format( printf, x, x + 1 )))
+	#endif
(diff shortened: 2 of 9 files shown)


engine/client/input/input.c
```diff
@@ -143,7 +143,12 @@ void IN_MouseSavePos( void )
 	if( !in_mouseactive )
 		return;
 
-	Platform_GetMousePos( &in_lastvalidpos.x, &in_lastvalidpos.y );
+	int x, y;
+
+	Platform_GetMousePos( &x, &y );
```
a1batross: Why?
Platform_GetMousePos accepts `int *` and `in_lastvalidpos` members are also `int`.

a1batross (changes requested):
This would be the first time this engine branch even supports MinGW. I'm not really fond of MinGW hacks and incompatibilities on 32-bit x86, but maybe it's fine for 64-bit x86, at least it doesn't have over 9000 historical calling conventions. It also unfortunately still has it's own C++ ABI, which I don't think is that big of a deal, we only need to preserve C++ ABI on GoldSrc supported targets i.e. `{win32,linux,apple}-i386`.

a1batross:
Can you also reformat your commit messages the way they are in git history?

danglingptr0x0:
> This would be the first time this engine branch even supports MinGW. I'm not really fond of MinGW hacks and incompatibilities on 32-bit x86, but maybe it's fine for 64-bit x86, at least it doesn't have over 9000 historical calling conventions. It also unfortunately still has it's own C++ ABI, which I don't think is that big of a deal, we only need to preserve C++ ABI on GoldSrc supported targets i.e. `{win32,linux,apple}-i386`.

imo better than Microsoft's bloated, VS-bound compiler. afaik this is one of the only ways to compile for Windows from Linux

danglingptr0x0:
> Can you also reformat your commit messages the way they are in git history?

done

engine/client/input/input.c
danglingptr0x0: I'm blind. sorry

a1batross (approved):

a1batross:
Thanks!

Issue #1943, opened 2024-12-31
[MOD: Cry of Fear] I can't open menu with esc button, meanwhile main menu (c_game_menu1) literally told us to press esc button for options.


TheKingFireS:
self-explanatory.
In Goldsrc 4554, I can open menu with esc button while in main menu (c_game_menu1).

Info:
Patch version: 1.55 (1.2 and 1.0 untested)
Xash3D-FWGS: https://github.com/FWGS/xash3d-fwgs/commit/d6ebff81e210f6d8cd2233a4e2f0a20bb9b6abec
OS: Linux Mint 21.3 via Wine (functioned exactly like Windows 10)
GPU: GTX 1650 GDDR6
Custom renderer: Enabled
Map: c_game_menu1 (main menu)

MohiMad:
This bug is happening inside the [Key_Event](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L688) function where the engine seems to be asking the client if it wants to handle the ESC key twice:
First it asks in [line 713](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L713) which the client declines (i confirmed this by adding an if statement right under which would have never have ran had the client handled the ESC key). But ESC is handled exceptionally a [bit downward in the function](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L768-L788). In the case of CoF in the start menu the else-if statement `host.mouse_visible && cls.state != ca_cinematic` runs (start menu asks for a visible cursor and its not a cinematic, neither the default case or the showtexture if-statement ran. This is where it gets interesting: the engine asks the client if it wants to handle the key and returns immediately, assuming that the key is handled by the client. But the client returned 1 to the engine earlier when asked in [line 713](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L713). The engine is essentially asking the same question twice and ignoring the first response. Removing the [else-if](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L779-L783) statement "fixes" this issue. Hitting ESC in the main menu triggers `togglemenu`. The reason why this affects the main menu only is because its the only place where the client asks for a cursor and lets the engine handle the key (for other interfaces such as the inventory / computer interface / animal puzzle etc the client handles the key itself). 

[...]

Issue #528, opened 2021-07-22
[MOD: Natural Selection] Не работает кнопка выхода в меню (cancelselect)


Vladislav4KZ:
Команда cancelselect (по умолчанию задана на кнопку ESC) не заходит в меню, а когда пишешь команду escape или menu_main то меню открывается

MohiMad:
I believe this issue is directly related / a duplicate of https://github.com/FWGS/xash3d-fwgs/issues/1943

MohiMad:
Update: the suggested [else-if deletion](https://github.com/FWGS/xash3d-fwgs/issues/1943#issuecomment-5240744551) "fix" does not resolve this issue. 

If I understood the issue correctly: while in-game (not main menu), hitting ESC should show the options menu (Goldsrc 4554), meanwhile in FWGS hitting ESC does nothing. This is expected since mid-gameplay the client does not ask for visible cursor, and therefore the [else-if](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L779-L783) is irrelevant here. Hitting ESC in main menu does not open an options menu in Goldsrc 4554.

The issue could however be caused due to a similar duplicate call `clgame.dllFuncs.pfnKey_Event.

Vladislav4KZ:
Исправлено в: https://github.com/FWGS/xash3d-fwgs/commit/02de74ff02c2ea9f903b9b4fe5431eb96cc65a55

PR #2717, opened 2026-08-22
Fix ESCAPE menu issues

MohiMad:
This PR proposes a fix to two issues with Key_Event's escape button branch. Specifically [issue #1943 (CoF)](https://github.com/FWGS/xash3d-fwgs/issues/1943) and [issue #528 (NS)](https://github.com/FWGS/xash3d-fwgs/issues/528). It also fixes a issue that arised while fixing issue 1943. 

The explanations are scattered across the issue pages and commit descriptions but I'll write a summary here:

## Issue 1943 (CoF)
While in main menu, pressing ESC does not open options menu. This bug is happening inside the [Key_Event](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L688) function where the engine asks the client if it wants to handle the ESC key twice:
First it asks in [line 713](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L713) which the client declines (i confirmed this by adding an if statement right under which would have never have ran had the client handled the ESC key). But ESC is handled exceptionally a [bit downward in the function](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L768-L788). In the case of CoF in the start menu the else-if statement `host.mouse_visible && cls.state != ca_cinematic` runs (start menu asks for a visible cursor and its not a cinematic). This is where it gets interesting: the engine asks the client if it wants to handle the key and returns immediately, assuming that the key is handled by the client. But the client returned 1 to the engine earlier when asked in [line 713](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L713). The engine is essentially asking the same question twice and ignoring the first response. Removing the [else-if](https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input/in_keys.c#L779-L783) statement "fixes" this issue. Hitting ESC in the main menu triggers `togglemenu`. The reason why this affects the main menu only is because its the only place where the client asks for a cursor and lets the engine handle the key (for other interfaces such as the inventory / computer interface / animal puzzle etc the client handles the key itself). 

However, the else-if removal triggered another issue downstream: hovering over a vgui button and pressing ESC now opens that vgui button. Basically, a keyboard button press triggered a nonexistent mouse click. This is happening because `Key_ClearStates` calls `IN_MouseEvent` (which doesn't check if an event is an actual mouse click that has been released), unlike `Key_Event` which checks if a button has been pressed. This *could* be solved by adding a condition within `Key_ClearStates` before calling `IN_MouseEvent` but I thought it was appropriate here to refactor it into its own `IN_ClearMouseState` function. Thats a taste thing and a decision to be made by the maintainers, but the fix either way needs to be shipped with the else-if removal. 

How does this change the behaviour of other mods? This only affects mods that return 1 and shows a vgui cursor (CoF's case). Instead of ESC doing nothing now it opens menu. I can't think of a case where the client returns 1, shows a vgui cursor yet intentionally doesn't want ESC to open menu. 

## Issue 528 (NS)
This issue is included here because i at first thought it was related to 1943, turns out its different but operates on the same branch. As Vladislav4KZ described it: "`cancelselect` doesn't open menu but `escape` / `main_menu` does". This is because NS relies on the engine to call `escape` internally, which FWGS doesn't. Compare NS's [CHudAmmo UserCmd_Close](https://github.com/unknownworlds/NS/blob/master/main/source/cl_dll/ammo.cpp#L960-L970) to HL SDK's  [CHudAmmo UserCmd_Close](https://github.com/FWGS/hlsdk-portable/blob/master/cl_dll/ammo.cpp#L753-L763): NS removed the "vanilla" `escape` branch. FWGS currently relies on MODS to internally call `escape`, while GoldSrc doesn't.  I added an escape call after the r_showtextures case

How does this affect other mods? Well, it affects all mods. Instead of going through `ESC -> cancelselect -> escape -> menu` the engine skips cancelselect: `ESC -> escape -> menu` (cancelselect stil works, its just not the route taken). So any mod that intentionally doesn't want ESC to open menu now opens menu. Very few (if any) mods would want that kind of behaviour anyway. Users [can't unbind ESC](https://github.com/FWGS/xash3d-fwgs/blob/ed2239e6138501c03a80bae95f4988c717cb2a64/engine/client/input/in_keys.c#L352-L356) so no one would complain. If anything, the NS commit ensures that the engine's desired behaviour actually runs (it enforces it).

### DIFF
--- a/engine/client/input/in_keys.c
+++ b/engine/client/input/in_keys.c
@@ -776,11 +776,8 @@ void GAME_EXPORT Key_Event( int key, int down )
 				Cvar_DirectSet( &r_showtextures, "0" );
 				return;
 			}
-			else if( host.mouse_visible && cls.state != ca_cinematic )
-			{
-				clgame.dllFuncs.pfnKey_Event( down, key, keys[key].binding );
-				return; // handled in client.dll
-			}
+			// Note: we don't offer the client key handling here since reaching this line proves the client already declined handling it.
+			// this prevents accidentally declining a key the client asked the engine to handle. For more context, see https://github.com/FWGS/xash3d-fwgs/issues/1943
 			break;
 		default:
 			break;
--- a/engine/client/input/in_keys.c
+++ b/engine/client/input/in_keys.c
@@ -902,15 +902,16 @@ void GAME_EXPORT Key_ClearStates( void )
 
 	for( int i = 0; i < ARRAYSIZE( keys ); i++ )
 	{
-		if( i >= K_MOUSE1 && i <= K_MOUSE5 )
-			IN_MouseEvent( i - K_MOUSE1, false );
-		else
-			Key_Event( i, false );
+		if( i < K_MOUSE1 || i > K_MOUSE5 )
+			Key_Event( i, false ); // checks internally whether a key has been pressed or not
 
 		keys[i].down = 0;
 		keys[i].repeats = 0;
 		keys[i].gamedown = 0;
 	}
+
+	// ensure that only actual mouse buttons are released
+	IN_ClearMouseState();
 
 	if( clgame.hInstance )
 		clgame.dllFuncs.IN_ClearStates();
--- a/engine/client/input/in_keys.c
+++ b/engine/client/input/in_keys.c
@@ -778,7 +778,10 @@ void GAME_EXPORT Key_Event( int key, int down )
 			}
 			// Note: we don't offer the client key handling here since reaching this line proves the client already declined handling it.
 			// this prevents accidentally declining a key the client asked the engine to handle. For more context, see https://github.com/FWGS/xash3d-fwgs/issues/1943
-			break;
+			
+			// call escape within the engine, since the client may not handle it (Natural Selection doesn't. See https://github.com/FWGS/xash3d-fwgs/issues/528)
+			Cbuf_AddText( "escape\n" );
+			return;
 		default:
 			break;
 		}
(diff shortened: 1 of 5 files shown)


engine/client/input/in_keys.c
```diff
@@ -776,12 +776,12 @@ void GAME_EXPORT Key_Event( int key, int down )
...
-			{
-				clgame.dllFuncs.pfnKey_Event( down, key, keys[key].binding );
-				return; // handled in client.dll
-			}
-			break;
+			// Note: we don't offer the client key handling here since reaching this line proves the client already declined handling it.
+			// this prevents accidentally declining a key the client asked the engine to handle. For more context, see https://github.com/FWGS/xash3d-fwgs/issues/1943
+			
+			// call escape within the engine, since the client may not handle it (Natural Selection doesn't. See https://github.com/FWGS/xash3d-fwgs/issues/528)
+			Cbuf_AddText( "escape\n" );
```
a1batross: Ideally, we shouldn't even call `escape` command here.

GoldSrc never does, `K_ESCAPE` is always mapped to `cancelselect` and can't be rebound to something else.

engine/client/input/input.c
```diff
@@ -395,6 +394,24 @@ void IN_MouseEvent( int key, int down )
...
+/*
+===========
+IN_ClearMouseState
+
+mouse button releases are handled here to ensure that only actual mouse button clicks are released
+===========
+*/
+void IN_ClearMouseState( void )
+{
+	int i;
```
a1batross: Put `i` declaration into `for` loop: `for( int i = 0...`

engine/client/input/in_keys.c
a1batross: We should instead directly open menu here instead. I think that's what GoldSrc actually does.

a1batross (changes requested):

a1batross (approved):

PR #2726, opened 2026-08-26
engine: client: don't read stale pe->solid when deciding where a physent's model goes

oscarsolerfollana:
`CL_CopyEntityToPhysEnt` decides whether an entity's model goes into `pe->studiomodel` or `pe->model` by testing `pe->solid` — but `pe->solid` is only assigned ~15 lines later in the same function. The test therefore reads whatever value was left in that physent slot by a previous frame: `CL_SetSolidEntities` only memsets element 0 (the world), so slots 1+ keep stale contents across frames.

When a slot that previously held a `SOLID_BSP` entity (a door, for example) is reused for a player, the stale test routes the player's studio model into `pe->model`, where game code expects a brush model with valid hulls.

Mods whose movement/collision code walks `pe->model->hulls` directly then read garbage. Natural Selection (which is on the supported mod list) crashes with `c0000005` at a fixed offset inside its statically linked collision code, reliably within minutes whenever another player is in view. The crash comes and goes with physent list churn — more entities entering/leaving the PVS means faster slot reuse — which is why it looked random and movement-dependent. Diagnosed by disassembling the faulting code in NS's `client.dll` down to the exact structures it walks (`pmove->physents[i].model->hulls[0]`), then auditing what fills them.

Fix: test `state->solid`, matching what `SV_CopyEdictToPhysEnt` already does on the server side. Also guard the `STUDIO_TRACE_HITBOX` branch against `mod` being NULL (the `CL_SetSolidPlayers` path does not validate `CL_ModelHandle`), again matching the server side.

Verified in NS 3.2: before the fix, a listenserver with any second player (human or bot) in view crashes within minutes; with the fix, full matches run without issue. No behavior change for any other slot/entity combination — the stale value only alters the outcome in exactly the buggy case.

### DIFF
--- a/engine/client/dll_int/cl_pmove.c
+++ b/engine/client/dll_int/cl_pmove.c
@@ -333,12 +333,12 @@ static void CL_CopyEntityToPhysEnt( physent_t *pe, entity_state_t *state, qboole
 
 	if( state->solid == SOLID_BBOX )
 	{
-		if( FBitSet( mod->flags, STUDIO_TRACE_HITBOX ))
+		if( mod != NULL && FBitSet( mod->flags, STUDIO_TRACE_HITBOX ))
 			pe->studiomodel = mod;
 	}
 	else
 	{
-		if( pe->solid != SOLID_BSP && ( mod != NULL ) && ( mod->type == mod_studio ))
+		if( state->solid != SOLID_BSP && ( mod != NULL ) && ( mod->type == mod_studio ))
 			pe->studiomodel = mod;
 		else pe->model = mod;
 	}


engine/client/dll_int/cl_pmove.c
```diff
@@ -333,12 +333,15 @@ static void CL_CopyEntityToPhysEnt( physent_t *pe, entity_state_t *state, qboole
...
 	if( state->solid == SOLID_BBOX )
 	{
-		if( FBitSet( mod->flags, STUDIO_TRACE_HITBOX ))
+		if( mod != NULL && FBitSet( mod->flags, STUDIO_TRACE_HITBOX ))
 			pe->studiomodel = mod;
 	}
 	else
 	{
-		if( pe->solid != SOLID_BSP && ( mod != NULL ) && ( mod->type == mod_studio ))
+		// check state->solid here: pe->solid is not set yet at this point, so it
```
a1batross: I don't think it really needs an explanation here at all: we already read from `state->solid` two lines above. It's more a typo than a deliberate choice.

oscarsolerfollana:
One clarification on reproduction, after testing both ways: the crash is specific to **classic NS 3.2**. I also ran the community-maintained NS 3.3b9 (ENSL/NS) against an engine build with this fix deliberately reverted, and it does **not** crash — its updated collision code checks `ent->model && ent->model->type == mod_brush` before walking hulls (`CollisionChecker.cpp`), so it skips the mis-routed physent. The stale read on the engine side happens all the same; 3.3 just doesn't step on it. So anyone trying to reproduce should use vanilla NS 3.2 (the version on the supported mod list).

a1batross (approved):
As with your previous PR, I do 100% agree with your change but want to keep code from stale comments: here it's just a typo and was never meant to read uninitialized memory.

a1batross:
> One clarification on reproduction, after testing both ways: the crash is specific to **classic NS 3.2**. I also ran the community-maintained NS 3.3b9 (ENSL/NS) against an engine build with this fix deliberately reverted, and it does **not** crash — its updated collision code checks `ent->model && ent->model->type == mod_brush` before walking hulls (`CollisionChecker.cpp`), so it skips the mis-routed physent. The stale read on the engine side happens all the same; 3.3 just doesn't step on it. So anyone trying to reproduce should use vanilla NS 3.2 (the version on the supported mod list).

Your solution is still correct, I believe we encountered bugs related to this in other mods but never caught the actual reason.

oscarsolerfollana:
Fair enough — comment dropped, the code stays clean and the story lives in the commit message. Pushed.

a1batross (approved):

PR #2730, opened 2026-08-27
ref: gl: draw normal-rendermode sprites according to their texture format

oscarsolerfollana:
GoldSrc (inheriting from Quake) draws sprites in `kRenderNormal` according to the sprite's own `texFormat`: additive sprites blend additively, indexalpha sprites alpha-blend. Xash disables blending entirely, so they render opaque.

Visible in Natural Selection: the lerk's spore projectile is a client-side temp entity that never sets a rendermode — its indexalpha sprite renders as an opaque green blob instead of a translucent cloud (correct on GoldSrc). Verified fixed in-game.

Minimal repro with any indexalpha sprite, no mod needed:

```
ent_create env_sprite model sprites/spore2.spr spawnflags 1
```

Explicit rendermodes are unaffected — the change only touches the `kRenderNormal`/default path.

### DIFF
--- a/ref/gl/gl_sprite.c
+++ b/ref/gl/gl_sprite.c
@@ -421,7 +421,23 @@ void R_DrawSpriteModel( cl_entity_t *e )
 		break;
 	case kRenderNormal:
 	default:
-		pglDisable( GL_BLEND );
+		// GoldSrc parity: in the normal render mode sprites draw according to
+		// their own texture format (Quake heritage) — additive sprites blend
+		// additively and indexalpha sprites alpha-blend
+		switch( psprite->texFormat )
+		{
+		case SPR_ADDITIVE:
+			pglEnable( GL_BLEND );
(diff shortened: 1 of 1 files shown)


a1batross:
Thanks, will review it later

ref/gl/gl_sprite.c
```diff
@@ -421,7 +421,23 @@ void R_DrawSpriteModel( cl_entity_t *e )
 		break;
 	case kRenderNormal:
 	default:
-		pglDisable( GL_BLEND );
+		// GoldSrc parity: in the normal render mode sprites draw according to
```
a1batross, line 424: GoldSrc doesn't really check for texFormat when it decides the blending, it only matters during texture load.

a1batross: I think we just implement the same blending GoldSrc uses on sprites and a gl_spriteblend cvar.

a1batross (changes requested):

PR #2753, opened 2026-09-19
fix: preserve permanent decals on demo recording and prevent duplication during playback

Vladislav4KZ:
- Add R_ClearAllDecals(includePermanent) to efx_api_s
- SV_RestartDecals: skip FDECAL_PERMANENT when re-sending decals for demo
- Round reset (client): call R_ClearAllDecals(false) to keep permanent decals
- Demo recording: R_ClearAllDecals(false) keeps permanent decals visible
- Demo playback: no doubled permanent decals (sent via svc_bspdecal on map load)


common/r_efx.h
```diff
@@ -184,6 +184,7 @@ struct efx_api_s
 	short		(*R_LookupColor)( unsigned char r, unsigned char g, unsigned char b );
 	void		(*R_DecalRemoveAll)( int textureIndex ); // textureIndex points to the decal index in the array, not the actual texture index.
 	void		(*R_FireCustomDecal)( int textureIndex, int entity, int modelIndex, float *position, int flags, float scale );
+	void		(*R_ClearAllDecals)( qboolean includePermanent ); // remove all decals; includePermanent=true also removes permanent decals
```
a1batross: This is Valve's interface, we don't extend them.

engine/ref_api.h
```diff
@@ -589,7 +589,7 @@ typedef struct ref_interface_s
 	void (*R_DecalShoot)( int textureIndex, int entityIndex, int modelIndex, vec3_t pos, int flags, float scale );
 	void (*R_DecalRemoveAll)( int texture );
 	int (*R_CreateDecalList)( struct decallist_s *pList );
-	void (*R_ClearAllDecals)( void );
+	void (*R_ClearAllDecals)( qboolean includePermanent );
```
a1batross, line 593: If you change API in backwards incompatible way, bump the version and write a small changelog.

a1batross (changes requested):

engine/ref_api.h
Vladislav4KZ: Done.

a1batross (approved):

PR #2756, opened 2026-09-21
ref: soft: render skybox cubemaps

Comicao1:
## What

The software renderer ignored custom skyboxes: `R_SetupSky` dropped the texture array the engine passes in, and sky spans fell back to a flat fill. This PR makes ref_soft actually render the skybox.

## How

- `R_SetupSky` keeps the 6 side textures (and clears them on unload).
- Sky spans are filled per pixel: each sky pixel's world-space view ray is built from the camera basis (`base_vpn`/`base_vright`/`base_vup` + projection scales), the dominant axis picks the cubemap face, and `s`/`t` are projected with the same `vec_to_st` tables ref_gl uses in `DrawSkyPolygon` — including the `t` flip `MakeSkyVec` applies for GL's bottom-up texcoords, since our mip rows are stored top-down.

## Notes

- Sampling is nearest-neighbor from mip 0, consistent with how ref_soft samples everything else (no bilerp anywhere in this renderer).
- Verified against ref_gl's mapping numerically: the direction → (face, s, t) mapping was checked for equivalence over a dense direction grid, and visually in Half-Life / Afraid of Monsters DC on a custom OS target.
- If no skybox is loaded, behavior is unchanged (flat fill).


ref/soft/r_edge.c
```diff
@@ -813,6 +813,104 @@ static void D_CalcGradients( msurface_t *pface )
...
+/*
+==============
+D_SkyboxSample
+
+Sample the loaded skybox cubemap for a world-space view direction.
+Face selection and st mapping mirror ref_gl's DrawSkyPolygon tables.
+==============
+*/
+static const int r_soft_skyTexOrder[SKYBOX_MAX_SIDES] = { 0, 2, 1, 3, 4, 5 };
+static const int r_soft_vec_to_st[SKYBOX_MAX_SIDES][3] =
```
a1batross: No reason to prefix vec_to_st and skyTexOrder with the `r_soft`.

In fact, they could be moved into `ref_common` static library, as the constants seem to be the same.

ref/soft/r_edge.c
```diff
@@ -822,13 +920,22 @@ The grey background filler seen when there is a hole in the map
...
 // set up a gradient for the background surface that places it
 // effectively at infinity distance from the viewpoint
 	d_zistepu = 0;
 	d_zistepv = 0;
 	d_ziorigin = -0.9;
 
-	D_FlatFillSurface( s, (int)sw_clearcolor.value & 0xFFFF );
+	for( int i = 0; i < SKYBOX_MAX_SIDES; i++ )
+		if( !tr.skyboxTextures[i] )
+			have_skybox = false;
```
a1batross, line 901:
```
for( int i = 0; i < SKYBOX_MAX_SIDES; i++ )
{
	if( !tr.skyboxTextures[i] )
		have_skybox = false;
}
```

ref/soft/r_edge.c
```diff
@@ -813,6 +813,104 @@ static void D_CalcGradients( msurface_t *pface )
...
+{  2,  3, -1 },
+{  1,  3,  2 },
+{ -1,  3, -2 },
+{ -2, -1,  3 },
+{ -2,  1, -3 }
+};
+
+static pixel_t D_SkyboxSample( const vec3_t dir )
+{
+	vec3_t  av;
```
a1batross: It would be nice to have parts of this whole function also moved to `ref_common` library and used in both `ref_soft` and `ref_gl`.

Should be easy enough.

Also, write new code with C99 in mind, i.e. you can actually declare variables when they used first, if that makes it more readable.

a1batross (changes requested):

Comicao1:
Thanks for the review! Reworked:

- The skybox projection is now shared: new `ref/common/ref_skybox.c` holds `r_skyTexOrder`, `vec_to_st`, and the new helpers `R_SkyboxAxisFromDir()` (dominant-axis face pick) and `R_SkyboxProject()` (direction → s/t via the shared table). ref_gl's `DrawSkyPolygon` uses them now too — I split it into two helpers because ref_gl picks the axis once per polygon from the vertex average, while ref_soft picks it per pixel; both share the table and projection math. The GL path's behavior is unchanged (same axis-from-average, same per-vertex projection).
- ref_soft applies the `t` flip from `MakeSkyVec` after projecting, since software mip rows are top-down.
- New code uses C99-style declarations, and the `skyboxTextures` check loop got braces.
- No more `r_soft`-prefixed duplicated tables.

Compile-checked ref_soft, ref_common and ref_gl (gl_warp.c) with the engine's real build flags.

a1batross (approved):

PR #2757, opened 2026-09-21
ref: soft: skip transparent texels in polyset span drawers

Comicao1:
## What

The polyset span drawers used for studio models and sprites in translucent render modes (`R_PolysetDrawSpansBlended`, `R_PolysetDrawSpansAdditive`, `R_PolysetDrawSpansGlow`, `R_PolysetDrawSpans8_33`, `R_PolysetFillSpans8`) blended/wrote every sampled texel unconditionally. Color-keyed transparent texels (`TRANSPARENT_COLOR`) were painted as solid color, producing fringes/artifacts around sprites and model skins that use keyed transparency.

## Fix

Skip the blend/write when the sampled texel is `TRANSPARENT_COLOR`, matching what the surface/warp paths already do.

Tested in Half-Life and Afraid of Monsters DC on a custom OS target using ref_soft as the only renderer.


ref/soft/r_polyse.c
```diff
@@ -770,13 +770,17 @@ void R_PolysetDrawSpansBlended( spanpackage_t *pspanpackage )
...
 
 					pixel_t temp = *lptex; // vid.colormap[*lptex + ( llight & 0xFF00 )];
-					int     alpha = vid.alpha;
-					temp = BLEND_COLOR( temp, vid.color );
 
-					if( alpha == 7 )
-						*lpdest = temp;
-					else if( alpha )
-						*lpdest = BLEND_ALPHA( alpha, temp, *lpdest ); // vid.alphamap[temp+ *lpdest*256];
+					if( temp != TRANSPARENT_COLOR )
```
a1batross: Doesn't look `TRANSPARENT_COLOR` is possible to set when Blended draw'er is selected.

a1batross (changes requested):

Comicao1:
You're right — respun without the `Blended` hunk.

Checked the drawer selection in `GL_Bind` (r_image.c): `Blended` is only selected when `image->alpha_pixels == NULL`, and `TRANSPARENT_COLOR` is only written into `pixels[0]` when the texture has one-bit alpha (which implies `alpha_pixels`), so it can never occur there. Dropped that guard.

The guards remain where masked textures can actually reach the drawer:

- `R_PolysetFillSpans8`: `kRenderNormal` is selected *before* the `alpha_pixels` check, so a masked texture on a normal-rendered model lands here and its transparent texels get drawn solid — this is the main visible artifact.
- `R_PolysetDrawSpansAdditive`/`Glow`: `kRenderTransAdd`/`kRenderGlow` are also selected before the `alpha_pixels` branch, so masked+additive/glow hits the same case.
- `R_PolysetDrawSpans8_33`: translucent triangles path (`tr.blend`-based), masked triapi textures can carry `TRANSPARENT_COLOR` too.

a1batross:
@Comicao1 while I merged these two PRs I would appreciate if you _didn't_ threw AI generated replies and AI generated PR description here. Whatever AI tries to explain is just pure visual garbage.

The code itself is fine, the code review handles that. But I can't really do anything with messages you post here, so I kindly ask try to write them by yourself, the way you understand the fix you're proposing.

a1batross:
It gives at least _some_ validation to what you did, that you actually have red the code that was written for you by an LLM, that you actually tried to run it.

Comicao1:
> It gives at least _some_ validation to what you did, that you actually have red the code that was written for you by an LLM, that you actually tried to run it.

Now it's me, my bad, I've been kinda lazy lately, so I've told my AI agents to do my work for me. I noticed those issues recently on my OS while working on it and thought they were small additions so it was fine. 

Yes, I ran it on linux/my os with cpu rendering, I did also compare to normal OGL half life and they were 1:1.

Thanks!

a1batross:
@Comicao1 sure, everything is OK, just wanted to know. :)

